PR #4, opened 2016-04-06
Update popper.js

ishanray:


src/popper.js
```diff
@@ -142,10 +142,10 @@
 
         // if the popper variable is a configuration object, parse it to generate an HTMLElement
         // generate a default popper if is not defined
-        var isNotDefined = popper === undefined || popper === null;
+        var isNotDefined = typeof popper === undefined || popper === null;
```
FezVrasta: typeof will return a string if I'm not wrong.

smashercosmo: Why not just ```popper == null```?

FezVrasta: I don't like such kind of generic comparisons

ishanray: `typeof popper === 'undefined'`

In case popper is not defined then that check will throw an error.

FezVrasta:
May you add a test to cover the code you changed?

FezVrasta:
Thank you, awesome!

If you may be so kind to fix the conflicts I can then proceed to merge this PR into master.

FezVrasta:
Great thanks.

Issue #6, opened 2016-04-07
Usage with React


schickling:
Hi @FezVrasta. Thanks for this great package!

I haven't yet looked at it too much but I'd be very interested to learn how this could (and should?) be used together with ReactJS. I think this might also be very interesting for others and therefore worthy to be added to the Readme?

Looking forward to hear your thoughts!

FezVrasta:
I'm very interested in adding full React.js support (along with Ember.js).

I'm working to move the code which applies the style to the popper into a dedicated modifier to make possible then to write a replacement modifier which will allow you to just export the coordinates to use them wherever you want.

schickling:
Amazing!

eek:
React support would be epic. If you need any beta testing, count me in! :)

mgcrea:
@FezVrasta ideally extracting the core coordinate calculus to another module would be great, eg. `popper-engine`, that would enable easy integration with either JS or even CSS frameworks.

FezVrasta:
refer to this PR please https://github.com/FezVrasta/popper.js/pull/11

FezVrasta:
#11 has been merged into master and released as v0.3.0

Now we plan to add a deeper integration with React.js using the API provided by v0.3.0

Any feedback or suggestion here please :)

FezVrasta:
I have written a simple implementation of a simple `PopperLoader` React component.

http://codepen.io/FezVrasta/pen/eZVJQZ

I'd like to get feedbacks from your to understand if this could be enough or if I should write something more. Thanks!

FezVrasta:
I've updated the component to be more modular. It now takes as arguments the reference component, the popper component and the optional configuration.

http://codepen.io/FezVrasta/pen/eZVJQZ

ebakan:
Here's a component I've been using that accepts the reference as its first child and the popper as its second child (a la [`react-tether`](https://github.com/souporserious/react-tether))

https://gist.github.com/ebakan/f3d94874a2e8bfbb06e3f9510c78ffd1

FezVrasta:
thanks @ebakan I've added your component to the README file.

I think now we are good with React.js support.

63b9840de3adaf0d766b92749f8ad5bd40c14036

PR #11, opened 2016-04-07
React.js support and improvements

FezVrasta:
- moved the DOM manipulating logic inside the `applyStyle` modifier
- the callback is now called using `.onCreate` method
- added `.onUpdate` callback
- improved README examples

ping @schickling @eek and @mgcrea for feedbacks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v0.3.0-dev
+
+- 5521b0d: added support for integration with React.js and Ember.js
+
 ## v0.2.6
 
 - 6732353: removed shared state;
--- a/README.md
+++ b/README.md
@@ -8,18 +8,18 @@
 Popper.js is a library used to create **poppers** in web applications.
 
 ## Wut? Poppers?
-A popper is an element on the screen which "pops out" from the natural flow of your application.  
+A popper is an element on the screen which "pops out" from the natural flow of your application.
 Common examples of poppers are tooltips and popovers.
 
 ## So, yet another tooltip library?
-Well, basically, **no**.  
-Popper.js is built from the ground up to being modular and fully ~~hackable~~ **customizable**.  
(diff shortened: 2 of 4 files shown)


schickling:
That was quite fast @FezVrasta! :+1: 

I think it would be very helpful if you'd feature a complete React component in the Readme as an example, so that there is no guess work needed.

FezVrasta:
I've not a lof of experience with React, can someone prepare one?

vampolo:
Separating the logic from the actual rendering is a first step into making this project more modular but i wouldn't call it React support.
What we mean *React support* might have different opinions, one that seems pretty straightforward and simpler to the user of this library would be to have a React Component that  can be used as

```
class MyElementWithPopper extends React.Component {
  render() {
    return <div>
              <div ref="reference">My text</div>
              <div ref="popper"> Content of the popper</div>
              </div>;
  }
}
```

There might be vary variations, like creating

```
<Popper reference=MyReferenceComponent popper=MyPopperComponent data=data>
```

That of course depends from how far the integration wants to go... What some libraries do like `redux` they have a base `redux` module and a `react-redux` integration...

FezVrasta:
I think I'm then going to merge this PR as is and then I'll try to prepare something more specific to support React out of the box.

shockwork:
@FezVrasta hey buddy, long time no see! I suggest forking it into a react-popper module, want me to do that?

FezVrasta:
Hey @shockwork :)
Happy to see you here!

I'd like to provide a basic usage example directly in this repository to keep the informations all in the same place (I'd like to avoid people to get confused or wander around different repositories).

I think that putting the stuff inside a `react/` folder could be a good start for now. I don't expect it to need webpack and the full react-style environment.

If you have time and it's not a problem for you then I'd love to get a contribution by you.

What do you think?

FezVrasta:
Please continue here: #6

PR #15, opened 2016-04-08
Changes to the assign polyfill

fodonnel:
Changes to the assign polyfill to convert the target and the nextSource to object.
I can determine the original source of this version of the polyfill to compare it to.

### DIFF
--- a/src/popper.js
+++ b/src/popper.js
@@ -1177,13 +1177,13 @@
                     throw new TypeError('Cannot convert first argument to object');
                 }
 
-                var to = target;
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thanks for the PR, I remember to have removed the two object conversions because I had tried to use `Object.assign` to extend an object with a `ClientRect` generated by `getBoundingClientRect`.

But since now we use the `getBoundingClientRect` function which takes care of converting the `ClientRect` in `Object` I think we can merge this PR safely.

PR #17, opened 2016-04-09
Performance optimization

FezVrasta:
This PR aims to introduce performance optimization to avoid useless computations between frames.
The idea is that we don't need to update the popper position more frequently then 60fps.

Dropping the updates between frames (the one our eye could not see) we reduce the computations during scrolling and resize.

One test is failing...
If someone has time to look at it and understand where's the problem it would be great.


creativeprogramming:
I think the problem is here:

```javascript  
// to avoid useless computations we throttle the popper position refresh to 60fps
 +        root.requestAnimationFrame(function() {
 +            var now = root.performance.now();
 +            if(now - this.state.lastFrame <= 16) {
 +                // this update fired to early! drop it
 +                return;
 +            }
```
Even I agree it's good as concept, I don't think you should try to throttle the requestAnimationFrame, throttling it's already automatically handled by the browser, limiting it again is teorically more performant, but may become complex.

What about if the browser decides to fire the last requestAnimationFrames very nerarly in time? Maybe the last animation frame will be never cause a repaint so you will have some final positining of some pixel less than expected (I don't know if your tests fails due to this, I hate tests so please don't ask me to read them, i just read the code for potential issues in my opinion, i cannot test it)

So please try to use a simple requestAnimation frame without limiting it again, it's anyway a good improvement: requestAnimation frame will be called only when browser knows that an expensive repaint will not have any negative side effect on performance, let's trust it!

creativeprogramming:
I confirm it can be this, and not just due to browser dom model, but you  too are using float as pixel value in the calculations, for instance, in `    Popper.prototype._getOffsets = function(popper, reference, placement) `

` popperOffsets.top = referenceOffsets.top + referenceOffsets.height / 2 - popperRect.height / 2;`

this can return a float value, not always integer! 

So your tests for boundaries can fail with a good chance it the last frame calculations are skipped due to the browser offering us  requestAnimationFrame invokations a a > 60fpm rate!

creativeprogramming:
Anyway, i hate tests, they may slow down development a lot, i think this PR is very good, as there is no real side effect!

creativeprogramming:
Anyway i think even a BETTER performance optimization can be achieved with a more simple solution, keep the .update() code as it is in master, get rid of all those issues and **do throttling only at 'scroll' and 'resize' eventListeners level**

Are them the only cases when the update() repaints are called like them are an animation right?

I'll open a separate issue for this discussion

FezVrasta:
About floating numbers, I already round them before they get outputted to the DOM, so no problems in this sense.

Right now on Travis everything works, on local I get:

```
Chrome 52.0.2743 (Mac OS X 10.11.6) Popper.js inits a popper near a reference element, both inside a fixed element on bottom of viewport, inside a scrolled body FAILED
	Expected [ 730, 739 ] to contain 727.
	    at Object.createCalback (/Users/federicozivolo/Progetti/popper-js/tests/test-popper.js:262:33)
	    at Popper.<anonymous> (/Users/federicozivolo/Progetti/popper-js/src/popper.js:258:32)
Chrome 52.0.2743 (Mac OS X 10.11.6): Executed 23 of 23 (1 FAILED) (0.34 secs / 0.33 secs)
```

Issue #27, opened 2016-04-13
Use with detached (fabricated) DOM node fails


rosskevin:
[Codepen test case](http://codepen.io/rosskevin/pen/zqWwmP?editors=1010)

**Scenario:**
I'm fabricating content then want to pop it up.  It appears since this content is not attached to the document that popper fails with:

```
VM2242 popper.js:1017 Uncaught TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'
....
Popper._getBoundaries	@	VM2242 popper.js:481
Popper.update	@	VM2242 popper.js:216
Popper	@	VM2242 popper.js:175
```

1. Is this a bug?
2. If not, what is the recommended way to pop dynamic content?

FezVrasta:
Popper.js needs to know the context of your element to properly position it, I don't think it's a bug.

The easiest way would be to spawn your element in the DOM and then instantly call Popper.js to position it.
Let me know if it can work for you.

rosskevin:
Works, I'm a little concerned it might flash, but I can probably work around that if it does (i.e. off page).

Here's a [working Codepen sample](http://codepen.io/rosskevin/pen/KzomLN?editors=1010)

rosskevin:
I'm thinking about PR'ing a change to the constructor.  Now that I'm using it a bit more, I'd like to get the functionality of the `#parse` function, but with a detached node.  I think it makes sense to allow the node, since the `#parse` creates the nice container with classes and appends.

I'll go ahead, let me know if you think this is a bad idea.  I'm thinking similar to `allowHtml`.  Perhaps `allowHtml` should be `allow` with `html | node`?

FezVrasta:
I don't think it could work for your purpose. If the detached node is larger than the generated container, once you append the node, the container will change its size and the position will probably be wrong.

rosskevin:
This works:
http://codepen.io/rosskevin/pen/PNRjOz?editors=0110

You want me to test with a large element?

rosskevin:
[Large node test case](http://codepen.io/rosskevin/pen/ONvgzR?editors=0110)

rosskevin:
PR available, I think it's a good change.  `#parse` is a handy method and I want all it's goodness applied to my generic content.

FezVrasta:
Closing in favor of the PR

Issue #25, opened 2016-04-13
Hello World - popper text overwrites reference until resize


rosskevin:
I feel foolish like I'm missing something essential.

[Codepen test case](http://codepen.io/rosskevin/pen/XdERKN?editors=1010)

![generic_popper_overwrites_text](https://cloud.githubusercontent.com/assets/136564/14505848/993d41b0-0180-11e6-90b7-218db209a6b5.png)

```html
<span class="myReference">Hello</span>
```

```javascript
var reference = document.querySelector('.myReference');
new Popper(reference, {content: 'World'}, {placement: 'right'})
```

rosskevin:
Is there a recommended baseline css to use with this for placement?  I'm [threw in the popper.css](http://codepen.io/rosskevin/pen/KzomZV) from the popper.js.org site and it's closer.

rosskevin:
It looks like minimally, if you don't have the following, you'll see this behavior:

```css
.popper {
  position: absolute;
}
```

FezVrasta:
Seems like a problem with the first update, if you then scroll or resize the window, it works properly.

Webarkitekt:
Still no solution about this issue ?

FezVrasta:
I need a free weekend to work on it, still had no time :/

A PR is welcome tho

FezVrasta:
Fixed the problem.

The `position` of the popper element wasn't set before the computations performed by the first `.update()`. Now I set the position before everything else, so we are sure the update works properly.

PR #28, opened 2016-04-13
Allow use of detached node, as well as text or html for `content`

rosskevin:
@see #27 - Use with detached (fabricated) DOM node fails

I changed option `allowHtml` to `allow` with options `text` (default) | `html` | `node` - supports adding dynamic/detached nodes with the #parse method and all it's goodness.

See #27 for the codepens.


rosskevin:
BTW - I didn't commit build or re-run docs just so it cleanly showed changes.  I can run these in the future if you would like.

src/popper.js
```diff
@@ -87,8 +87,8 @@
      * @param {Array} [popper.classNames=['popper']] Array of classes to apply to the generated popper.
      * @param {Array} [popper.attributes] Array of attributes to apply, specify `attr:value` to assign a value to it.
      * @param {HTMLElement|String} [popper.parent=window.document.body] The parent element, given as HTMLElement or as query string.
-     * @param {String} [popper.content=''] The content of the popper, it can be text or HTML, in case of HTML, enable `allowHtml`.
-     * @param {Boolean} [popper.allowHtml=false] If set to true, the `content` will be parsed as HTML.
+     * @param {String} [popper.content=''] The content of the popper, it can be text, html, or node; if it is not text, set `allow` to `html` or `node`.
+     * @param {String} [popper.allow=text] If `html`, the `content` will be parsed as HTML. If `node`, it will be appended as-is.
```
FezVrasta: This should be `popper.allow='text'`

FezVrasta:
I would rename `allow` to `contentType`

FezVrasta:
About the build and docs tasks, it's okay to not commit them. I will run them when I'll release a new version.

FezVrasta:
By the way, I don't feel comfortable in merging this PR until #25 is fixed.

I want to be sure that the initial placement of "injected nodes" works properly.

rosskevin:
Update to `contentType`, I think that makes more sense.

FezVrasta:
@rosskevin now that #25 is fixed I'd like to continue with this PR.

May you remove the build result from this PR and fix the conflicts? Thanks.

rosskevin:
Ok, PR is all set. 

It contains the master's build artifacts so my test case won't work until it is pointed at a new build.

PR #35, opened 2016-04-21
fix object check in ie

judge:
We found an error under IE11 and below, it seems that ```popper.constructor.name``` gives undefined in IE, it should be 'Object'.
Created a more robust solution to check the config.


FezVrasta:
Thanks, I need just the `src/` edits, not the whole build.

May you remove the build files from the PR?

judge:
Done ✌️

PR #37, opened 2016-04-27
make onCreate and onUpdate chainable

FezVrasta:
I've added a test to make sure the feature work...
If someone can understand where the problem is...

It correctly enter inside the `prototype.onUpdate` function, but when it sets `this.state.updateCallback` something goes wrong.
In fact,  when `update` is called, `updateCallback` is `undefined`

### DIFF
--- a/src/popper.js
+++ b/src/popper.js
@@ -240,6 +240,7 @@
     Popper.prototype.onCreate = function(callback) {
         // the createCallbacks return as first argument the popper instance
         callback(this);
+        return this;
     };
 
     /**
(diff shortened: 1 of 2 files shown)


sqal:
Actually `this` should be also returned at the on of constructor function https://github.com/FezVrasta/popper.js/blob/master/src/popper.js#L181 becouse currently i can't do check like this

```
if (!this.popper) {
  this.popper = new Popper()
}
```

FezVrasta:
@sqal I'm pretty sure you can store the instance as you described. I test it in the automatic tests.

https://github.com/FezVrasta/popper.js/blob/master/tests/test-popper.js#L36-L46

sqal:
Ohh you're right! I made some mistake inmy app! :P

Issue #46, opened 2016-05-06
Result of getOuterSizes is not consistent


rafaelverger:
### Steps to reproduce the problem

1.  Go to https://popper.js.org/
2.  Scroll page to section **POPPER ON YOUR SIDE!**
3.  Wait popper elements on the left side toggle
4.  When a new popper appears, scroll the page a bit

*the change is more notable when it is the left popper that is visible*

### What is the expected behavior?

Popper element stays on the same position

### What went wrong?

Popper element reposition itself due to changes in getComputedStyle result

### Any other comments?

Check this gif:
![image](https://s3.amazonaws.com/f.cl.ly/items/1R3y0b2d2Z3D2S1w3s3C/Screen%20Recording%202016-05-06%20at%2002.41%20PM.gif?v=88c2ca12)

rafaelverger:
I believe that the error is related to the use of `root.getComputedStyle` method with an invisible element;

marciovicente:
I got the same error!

PR #47, opened 2016-05-10
Fix calculation of popper elements' outerSizes

rafaelverger:
This pull requests fixes popper.js code in order to correctly calculate outerSize of popper element and its arrows, event if they are set as `display: none`. This will fix the issue #46 .

This codepen is using the new popper.js script and doesnt have the re-positioning bug:
http://codepen.io/rafaelverger/pen/WwPNEd

### DIFF
--- a/src/popper.js
+++ b/src/popper.js
@@ -158,7 +158,7 @@
         // refactoring modifiers' list
         this._options.modifiers = this._options.modifiers.map(function(modifier){
             // remove ignored modifiers
-            if (this._options.modifiersIgnored.indexOf(modifier) >= 0) return;
+            if (this._options.modifiersIgnored.indexOf(modifier) !== -1) return;
 
             // set the x-placement attribute before everything else because it could be used to add margins to the popper
             // margins needs to be calculated to get the correct popper offsets
@@ -167,7 +167,7 @@
             }
 
             // return predefined modifier identified by string or keep the custom one
-	    return this.modifiers[modifier] || modifier;
+            return this.modifiers[modifier] || modifier;
         }.bind(this));
 
         // make sure to apply the popper position before any computation
@@ -931,13 +931,13 @@
         var _display = element.style.display, _visibility = element.style.visibility;
         element.style.display = 'block'; element.style.visibility = 'hidden';
         var calcWidthToForceRepaint = element.offsetWidth;
-        
+
         // original method
         var styles = root.getComputedStyle(element);
         var x = parseFloat(styles.marginTop) + parseFloat(styles.marginBottom);
         var y = parseFloat(styles.marginLeft) + parseFloat(styles.marginRight);
         var result = { width: element.offsetWidth + y, height: element.offsetHeight + x };
-        
+
         // reset element styles
         element.style.display = _display; element.style.visibility = _visibility;
         return result;
--- a/src/popper.js
+++ b/src/popper.js
@@ -155,17 +155,9 @@
         // with {} we create a new object with the options inside it
         this._options = Object.assign({}, DEFAULTS, options);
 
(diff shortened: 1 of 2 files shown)


src/popper.js
```diff
@@ -155,25 +155,27 @@
         // with {} we create a new object with the options inside it
         this._options = Object.assign({}, DEFAULTS, options);
 
+        // remove ignored modifiers
+        this._options.modifiers = this._options.modifiers.filter(function(modifier){
+            return this._options.modifiersIgnored.indexOf(modifier) === -1;
+        }.bind(this));
```
FezVrasta: why have you moved this part outside of the map function? Now we have two loops instead of one.

rafaelverger: Because the old way makes the below assertion always evaluate to false and `x-placement` attribute was never set;

rafaelverger: It is possible to unify all the loops and the assertion into a single loop if it is ok to have a list of modifiers with null values..
```js
this._options.modifiers = this._options.modifiers.map(function(modifier){
  if (this._options.modifiersIgnored.indexOf(modifier) >= 0) return;
  if (modifier === 'applyStyle') this._popper.setAttribute('x-placement', this._options.placement);
  return this.modifiers[modifier] || modifier;
});
```

FezVrasta: Yes it's okay. We already check if the current element is a function here:
https://github.com/FezVrasta/popper.js/blob/master/src/popper.js#L557

We still need such check to be sure that the passed modifier is a function, so we don't care.

rafaelverger: But if we run the `filter` after replacing modifiers strings by functions, it will not remove the replaced modifiers because `modifiersIgnored` has only strings..

We also cannot set `x-placement` attribute checking only if `modifiers` has `applyStyle` because it can be in the list `modifiersIgnored`

src/popper.js
```diff
@@ -155,24 +155,20 @@
...
-        }.bind(this));
 
-        // set the x-placement attribute before everything else because it could be used to add margins to the popper
-        // margins needs to be calculated to get the correct popper offsets
-        if (this._options.modifiers.indexOf('applyStyle') !== -1) {
-            this._popper.setAttribute('x-placement', this._options.placement);
-        }
+            // return predefined modifier identified by string or keep the custom one
+	    return this.modifiers[modifier] || modifier;
+        }.bind(this));
```
rafaelverger: @FezVrasta check this out.. just one iteration over modifiers and doing `filtering`, `replacement` and set attribute `x-placement`

src/popper.js
```diff
@@ -155,24 +155,20 @@
...
-                if (this._options.modifiersIgnored.indexOf(modifier) !== -1) {
-                    return;
-                }
-                return this.modifiers[modifier];
-            } else {
-                return modifier;
+        // refactoring modifiers' list
+        this._options.modifiers = this._options.modifiers.map(function(modifier){
+            // remove ignored modifiers
+            if (this._options.modifiersIgnored.indexOf(modifier) >= 0) return;
```
FezVrasta: Please use `!==`, it's the one used in the rest of the app.

src/popper.js
FezVrasta: seems good, thanks. may you fix the indentation pls?

rafaelverger: Sure!

rafaelverger:
@FezVrasta , do you have any other comments?

FezVrasta:
Nope thanks, I'll test the PR in the next days and I'll then merge it.

I'm a bit busy in these days, going to relocate to a new city (and country)

rafaelverger:
No problem, dude :) I hope everything goes well (with the tests and your move) !

Issue #52, opened 2016-06-06
hide popper when it scrolls outside of its boundaries


andresgutgon:
### What is the expected behavior?
I would expect popper to disappear when element is not visible

## Demo
![overflow](https://cloud.githubusercontent.com/assets/49499/15817609/244521d8-2bd9-11e6-9196-0efb5f813393.gif)

To reproduce this you just need to change height on `.example2__scroll-box` from `200%` to `10000px`
![image](https://cloud.githubusercontent.com/assets/49499/15817628/3c737688-2bd9-11e6-8f22-cef232c69692.png)

## Solution
I'll try to do something with a custom modifier. But I think this should be fixed in the core.

FezVrasta:
Yes probably we should add a "hide when reference is not visible" modifier (maybe with a shorter name ahah).

If you create one please send a PR, it would be a nice addition!

andresgutgon:
What is the best why to hide popper? I mean. With a modifier you just past `data` object modified. What should I pass to hide the popper?

FezVrasta:
I think the `applyStyle` modifier should be edited a bit to allow custom CSS properties.

andresgutgon:
Hi, I'm not sure I have time to do the PR now. But this is what I did:

I defined a modifier called `hideOnReferenceNotVisibleOnBoundaries` and if boundaries is defined check their visibility. Be aware that this code is jQuery dependent

``` javascript
      /**
       * Check if reference is visible inside boundaries
       *
       * @param {Object} data
       * @param {Object} boundariesElement
       * @return {Boolean}
       */
      function isVisible(data, boundariesElement) {
        const { boundaries, offsets } = data;
        const { top, bottom } = offsets.reference;
        const { height: boundariesHeight} = boundaries;
        const boundariesOffsetTop = $(boundariesElement).offset().top;
        const boundariesBottom = boundariesOffsetTop + boundariesHeight;

        return ((top <= boundariesBottom) && (bottom >= boundariesOffsetTop));
      }

      /**
       * Modifier to hide popper if is outside boundaries
       * If boundaries is not defined we skip the check
       *
       * @param {Object} data
       * @return {Object}
       */
      const hideOnReferenceNotVisibleOnBoundaries = (data) => {
        const boundariesElement = data.instance._options.boundariesElement;

        if (boundariesElement === 'viewport') return data;

        const visible = isVisible(data, boundariesElement);

        // Hide popper
[...]

PR #53, opened 2016-06-14
added ability to set custom styles

FezVrasta:
This PR allows 3rd party modifiers to hook inside the `applyStyles` modifier.

Thanks to this edit, your custom modifier can now apply custom styling to the popper element defining the CSS properties inside `data.styles`.

Example:

```js
function hidePopperModifier(data) {
  data.styles.display = 'none';
  return data;
}
```

This should help with #52


andresgutgon:
👏

PR #57, opened 2016-06-25
Added support for a reference el that is both fixed and transformed.

### LINKED ISSUE #56: Incorrect placement inside of fixed container with transforms
I have an application with a drawer with `position: fixed` and is pushed in and out of the interface using transforms. I've found that placing a Popper within that drawer causes the position of that Popper to be way outside of the viewport. It seems that this is caused by the transform on the parent drawer because the transform origin is no longer relative to `document.body` but the drawer element.

Here's a [CodePen](http://codepen.io/hgascoigne/pen/YWNeBw) illustrating the issue. Uncomment the `transform: translateX(0)` line in the CSS to see what I mean. I'm going to try and work up a solution on my end, but since I don't know your code very well I wanted to put it up here so you might find a better way than me to fix it.

hgascoigne:
This fixes #56.

Similar to how the parent's position is checked, I added a check for any transformations on the parent. If transforms are present, Popper changes the way it calculates offsets. This is necessary because when a parent has a transform applied, the origin of the transform switches to that element instead of the viewport. See the [W3C Spec](http://www.w3.org/TR/css3-2d-transforms/#transform-rendering):

> In the HTML namespace, any value other than none for the transform results in the creation of both a stacking context and a containing block. The object acts as a containing block for fixed positioned descendants.

Since I added the check for `isParentTransformed`, any existing Popper should (in theory) work the same way. I tried to mimic the existing code flow/style as much as possible, but if you have any suggestions I am all ears.

Thanks!

### DIFF
--- a/src/popper.js
+++ b/src/popper.js
@@ -180,6 +180,9 @@
         this.state.position = this._getPosition(this._popper, this._reference);
         setStyle(this._popper, { position: this.state.position});
 
+        // determine how we should set the origin of offsets
+        this.state.isParentTransformed = this._getIsParentTransformed(this._popper);
+
         // fire the first update to position the popper in the right place
         this.update();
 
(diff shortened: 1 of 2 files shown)


FezVrasta:
Thanks! What if I'm performing a 2d transformation? Seems like you just check for any kind of trasformation but only the 3d ones should affect the position I think.

FezVrasta:
Sorry if I've missed something but I'm from mobile these days

hgascoigne:
No worries! I simply check that transform is not `none` because any transformation will change the origin position of the transformation. From the snippet from the spec above:

> any value other than none for the transform results in the creation of both a stacking context and a containing block

PR #64, opened 2016-07-25
Performance optimization: rate-limiting by throttling the onscroll and onresize events

creativeprogramming:
Added also a 'maybe useful' new attachable event: onAfterUpdate 

see it in action here:
http://codepen.io/creativeprogramming/pen/KrRXLQ

image

*note that on those codepens i added some raw and expensive code to debug the fps on the screen, so it may slow down rate a bit more

you can compare with  the behavior of the current master same demo here:

http://codepen.io/creativeprogramming/pen/OXkOgx
image

### DIFF
--- a/src/popper.js
+++ b/src/popper.js
@@ -75,6 +75,10 @@
         modifiers: [ 'shift', 'offset', 'preventOverflow', 'keepTogether', 'arrow', 'flip', 'applyStyle'],
 
         modifiersIgnored: [],
+        
+        //limit the max repaint and computation framerate to achieve more performance and less resource usage on the user browser and even save some trees ;)
+        maxFps: 60,
+        afterUpdateTriggerDelayMs: 200
     };
 
     /**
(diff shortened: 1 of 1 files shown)


creativeprogramming:
[refs #17 refs #67]

src/popper.js
```diff
@@ -266,6 +366,20 @@
...
+    /**
+     * If a function is passed, it will be executed after each update of popper with as first argument the set of coordinates and informations
+     * used to style popper and its arrow.
+     * NOTE: it doesn't get fired on the first call of the `Popper.update()` method inside the `Popper` constructor! 
+     * NOTE: it will be triggered with <options.afterUpdateTriggerDelayMs> milliseconds of delay after any update() 
+     * @method
+     * @memberof Popper
+     * @param {Function} callback
+     */
+    Popper.prototype.onAfterUpdate = function(callback) {
```
FezVrasta: This doesn't seem something useful, you can just write your own logic inside the `onUpdare` hook.

I'm trying to keep the size of the library small.

src/popper.js
```diff
@@ -191,6 +195,94 @@
...
+     * so we can port here just some tiny bits it as we need just some utils functions: a 'dep' that's as tiny
+     * to be considered a 'no-dep'
+     * 
+     * @license https://github.com/jashkenas/underscore/blob/master/LICENSE
+     * @external 
+     * @see {@link http://underscorejs.org/}
+     *
+     * 
+    */
+    Popper.prototype._ = {
```
FezVrasta: I would consider some lighter alternative to this function, it seems pretty complex for nothing.

src/popper.js
creativeprogramming: it's to avoid too invokations of that logic, what if i need to execute something just when positioning is terminated and not at 60fps?

FezVrasta: I would tweak the `onUpdate` to get fired only when the position is actually updated. Otherwise it becomes useless.

creativeprogramming:
Anyway, i did this PR just for one reason: i wanted to thank you for this awesome work, already very good performant, i don't have any issue with it, i just saw your 'HELP WANTED' and I gave you some time for my 2 cents

image

No need for merging my PR, take it just for inspiration

FezVrasta:
Thanks for your time, I think I'll fork it and continue working on it then.

src/popper.js
creativeprogramming: yes i wanted to do that, actually i personally don't understand the sense of calling it at any animation frame, but i leaved as it is as i don't know if someone needs this

creativeprogramming: anyway onAfterUpdate is used for instance to print the last red line in the codepen demo i did for this:

http://codepen.io/creativeprogramming/pen/KrRXLQ
image

i needed it for just for this ;)

creativeprogramming:
wait, merge if you fork, please don't take if you don't like but if you take credits is expected

creativeprogramming:
yes, you can also merge and then rework, or fork to cut my pr off, no issue for me, anyway first solution (merge+rework) will be more correct if you'll end in taking from mine the very same logic ;) 

I'm joking... really it's up to you, credits will not change my life, i enjoyed doing it

FezVrasta:
The commits would be preserved, no problem :)

FezVrasta:
Closing in favor of the existing #17, I've almost fixed it (only 1 test failing for 3pixels) :tada:

PR #77, opened 2016-08-23
Using parentNode.removeChild when element.remove is not defined

rafaelverger:
This fix `popper.destroy` method on browsers that doesn't support [Element.remove method](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove#Browser_compatibility)

### DIFF
--- a/src/popper.js
+++ b/src/popper.js
@@ -210,11 +210,7 @@
 
         // remove the popper if user explicity asked for the deletion on destroy
         if (this._options.removeOnDestroy) {
-            if (this._popper.remove) {
-                this._popper.remove();
-            } else {
-                this._popper.parentNode.removeChild(this._popper);
-            }
+            this._popper.parentNode.removeChild(this._popper);
         }
         return this;
     };
--- a/src/popper.js
+++ b/src/popper.js
@@ -210,7 +210,11 @@
 
         // remove the popper if user explicity asked for the deletion on destroy
         if (this._options.removeOnDestroy) {
-            this._popper.remove();
+            if (this._popper.remove) {
+                this._popper.remove();
+            } else {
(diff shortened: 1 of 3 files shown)


src/popper.js
```diff
@@ -210,7 +210,11 @@
 
         // remove the popper if user explicity asked for the deletion on destroy
         if (this._options.removeOnDestroy) {
-            this._popper.remove();
+            if (this._popper.remove) {
```
FezVrasta: Why this check? Couldn't we simply call `parentNode.removeChild(this._popper)`? We do this on the `v1-dev`branch

rafaelverger: Yeah, could be only a call of parentNode.removeChild, but I thought u had chosen Element.remove for some reason and that's why I've kept it;

FezVrasta: No I just forgot about the missing support on IE, if you could fix this thing we can merge this and release a patch version

FezVrasta: For reference, this is what we do on `v1` https://github.com/FezVrasta/popper.js/blob/v1-dev/src/popper.js#L328

rafaelverger:
@FezVrasta it's done :)

tests/test-popper.js
```diff
@@ -335,7 +335,7 @@ describe('Popper.js', function() {
             done();
         });
     });
-
+
```
FezVrasta: Argh, can you remove these? 😔

FezVrasta:
Also, it would be nice if you could remove the notice from the README about Internet Explorer.

rafaelverger:
@FezVrasta done again :)

FezVrasta:
merged

Issue #78, opened 2016-08-24
Improve detection strategy of flip modifier


FezVrasta:
> @nadiam84 commented:
> @FezVrasta Popper behaves a bit weird inside a scrollable container (alfa-version).
> 
> https://codepen.io/nadiam84/pen/grywdy
> 
> When it's scrolled vertically, it waits when the parent will be completely hidden, and only than start to move itself. When it's scrolled horizontally, it doesn't go away at all.
>
>  ![d58dee92-69e2-11e6-8ca2-a28c14fae970](https://cloud.githubusercontent.com/assets/5382443/17923861/6d6becfc-69e6-11e6-8427-240f34d3060a.gif)

> 
> Is there a way to make it permanently attached to its refElement? Isn't 'keepTogether' options suppose to do this?

------

So, nothing new so far, seems like this behavior is not supported.

https://codepen.io/FezVrasta/pen/pbXozA

We use `preventOverflow` modifier to keep the popper inside the boundaries.
`flip` modifier flips the popper when it detects that it's overlapping its reference element.

We should probably change `flip` to use a different detecting strategy (aka, check the boundaries overflow like `preventOverflow` does, but flip the popper instead of move its offsets to stay within the boundaries.)
We should then apply the `flip` modifier **before** the `preventOverflow` modifier.

I hope it makes sense. I'm quite busy in these days so I don't know when I'll be able to work on it. The good thing is that whoever wants to work on this can just edit the `flip` modifier copying how `preventOverflow` works without having to fiddle with the rest of the code.

nadiam84:
Created a PR for this, @FezVrasta could you please take a look?
Tests locally are all green.

PR #82, opened 2016-09-04
Improve detection strategy of flip modifier

nadiam84:
This PR is an attempt to solve the issue #78. There are two commits here. 

The first one is solving "moving after target" problem. I.e. popper will behave like it doesn't have 'preventOverflow' modifier, but still will flip itself from the parent. It looks like this:

https://goo.gl/photos/H5QafmbZ1xjVuHeD6

Enabling this behaviour would look like this:
```
modifiers: {
    preventOverflow: {
        moveWithTarget: true
    }
}
```

The second commit is improving flipping behavior so that popper can also flip its position's variation ('start' to 'end' and vice versa). It looks like this:

https://goo.gl/photos/MtneBBm9nNd6Nynz6

```
modifiers: {
    flip: {
        flipVariations: true
    }
}
```

### DIFF
--- a/src/modifiers/flip.js
+++ b/src/modifiers/flip.js
@@ -15,7 +15,6 @@ import runModifiers from '../utils/runModifiers';
  * @argument {Object} options - Modifiers configuration and options
  * @returns {Object} The data object, properly modified
  */
-
 export default function flip(data, options) {
     // check if preventOverflow is in the list of modifiers before the flip modifier.
     // otherwise flip would not work as expected.
@@ -64,10 +63,10 @@ export default function flip(data, options) {
             !a && Math.floor(data.offsets.reference[placement]) < Math.floor(popperOffsets[placementOpposite]);
 
         const flippedVariation = options.flipVariations && (
-            b && (variation === 'start') && popperOffsets.left < data.boundaries.left ||
-            b && (variation === 'end') && popperOffsets.right > data.boundaries.right ||
-            !b && (variation === 'start') && popperOffsets.top < data.boundaries.top ||
-            !b && (variation === 'end') && popperOffsets.bottom > data.boundaries.bottom);
+            b && (variation === 'start') && Math.floor(popperOffsets.left) < Math.floor(data.boundaries.left) ||
+            b && (variation === 'end') && Math.floor(popperOffsets.right) > Math.floor(data.boundaries.right) ||
+            !b && (variation === 'start') && Math.floor(popperOffsets.top) < Math.floor(data.boundaries.top) ||
+            !b && (variation === 'end') && Math.floor(popperOffsets.bottom) > Math.floor(data.boundaries.bottom));
 
         if (
             flippedPosition || flippedVariation
--- a/src/modifiers/flip.js
+++ b/src/modifiers/flip.js
@@ -1,4 +1,5 @@
 import getOppositePlacement from '../utils/getOppositePlacement';
+import getOppositeVariation from '../utils/getOppositeVariation';
 import getPopperClientRect from '../utils/getPopperClientRect';
 import getOffsets from '../utils/getOffsets';
 import isModifierRequired from '../utils/isModifierRequired';
@@ -14,6 +15,7 @@ import runModifiers from '../utils/runModifiers';
  * @argument {Object} options - Modifiers configuration and options
  * @returns {Object} The data object, properly modified
  */
+
 export default function flip(data, options) {
     // check if preventOverflow is in the list of modifiers before the flip modifier.
     // otherwise flip would not work as expected.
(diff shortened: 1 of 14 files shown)


FezVrasta:
It looks cool, thanks!

The build has some problem with the authentication... No idea what changed, I'll take a look.

I have a flight for SF in an hour so I'll not be able to review this PR anytime soon, I'll work on it once I'm arrived.

Could you add some functional test to test the new behaviors you added?

src/modifiers/flip.js
```diff
@@ -14,6 +15,7 @@ import runModifiers from '../utils/runModifiers';
  * @argument {Object} options - Modifiers configuration and options
  * @returns {Object} The data object, properly modified
  */
+
```
FezVrasta: No space between JSdocs comments and the function please 🙃

src/modifiers/preventOverflow.js
```diff
@@ -1,5 +1,5 @@
 import getPopperClientRect from '../utils/getPopperClientRect';
-
+import getOppositePlacement from '../utils/getOppositePlacement';
```
FezVrasta: add a newline after this pls? Sorry if I ask for such silly things but if you can fix this I don't have to clone it locally and merge manually, thanks! 😬

src/utils/getOppositeVariation.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Get the opposite placement variation of the given one/
+ * @method
+ * @memberof Popper.Utils
+ * @argument {String} placement variation
+ * @returns {String} flipped placement variation
+ */
+export default function getOppositeVariation(variation) {
+    const hash = {start: 'end', end: 'start'};
+    return variation.replace(/start|end/g, (matched) => hash[matched]);
```
FezVrasta: This could be simplified, `getOppositePlacement` used this approach because it takes as input `position-variation`, this function instead takes only `variation`, so we could increase its performance with a simpler code:

```
if (variation === 'end') {
  return 'start';
} else if (variation === 'start') {
  return 'end';
}
return variation;
```

src/modifiers/flip.js
```diff
@@ -54,18 +57,32 @@ export default function flip(data, options) {
...
 
         // using Math.floor because the reference offsets may contain decimals we are not going to consider here
+        const flippedPosition = a && Math.floor(data.offsets.reference[placement]) > Math.floor(popperOffsets[placementOpposite]) ||
+            !a && Math.floor(data.offsets.reference[placement]) < Math.floor(popperOffsets[placementOpposite]);
+
+        const flippedVariation = options.flipVariations && (
+            b && (variation === 'start') && popperOffsets.left < data.boundaries.left ||
+            b && (variation === 'end') && popperOffsets.right > data.boundaries.right ||
+            !b && (variation === 'start') && popperOffsets.top < data.boundaries.top ||
+            !b && (variation === 'end') && popperOffsets.bottom > data.boundaries.bottom);
```
FezVrasta: Are you sure we don't need `Math.floor` here?

nadiam84:
Sorry about the formatting, too used to rely on linting for this kind of things :) I'll fix it tomorrow. Have a safe flight!

nadiam84:
@FezVrasta updated according to your comments, sorry for the delay

FezVrasta:
Hey thanks! I'm still trying to figure out how to make the tests pass also on PRs, I'm going to try some suggestions from Sauce Labs in the next days. I'm sorry but I'm not very responsive these days due to the trip to SF, once I'll be back at home I'll be responsive as usual!

FezVrasta:
@nadiam84 I sent a PR to this PR (LOL) to make the tests pass here.

I'd say that once you merge it and some tests for this new behavior are added we can merge this into `v1-dev`!

In the next days I'm going to finish all the remaining work on v1 and I'll release the first beta

FezVrasta:
I was playing with the `flipWithoutPreventOverflow.html` visual test and it doesn't seem to work...

flip

As you see, it follows the ref only on vertical scroll, on horizontal it stays between boundaries.

nadiam84:
@FezVrasta did you build the js files before this? I haven't commited any build results as per your contribution guidlines :)
I'll write some tests today/tomorrow, completely forgot about them. Although the saucelabs tests are still not working.

FezVrasta:
I think I compiled them... But now you make me doubt about it lol.

I'll check again shortly

FezVrasta:
Ah! You were right lol. Yes it seems to work 🙃

Need some tests to test the new functionalities, it's okay that the old tests still work but we must be sure that the new features are tested as well.

nadiam84:
@FezVrasta updated the PR with tests. Locally they were green :D Checked in the latest Chrome, Firefox and Safari.

nadiam84:
@FezVrasta gentle nudge

FezVrasta:
Yes sorry! Forgot about it. The tests are failing because of a problem with Sauce Labs... Going to merge this anyway.

PR #87, opened 2016-09-18
auto placement feature

FezVrasta:
This PR should implement #1 

an `autoPlacement` modifier have been added, it simply converts the string `auto` in the `placement` option with the direction where there's more space available.

I thought it would be enough to make the rest work accordingly but as you can see using `tests/visual/test.html` the feature is broken.

The problem seems to rely on the fact that we don't have the computed `placement` value when we run `getOffsets` inside the `update` method of Popper.js.

Some refactor is needed in order to be able to create and run an `autoPlacementOnLoad` modifier that will be executed on initialization...
Another alternative would be to get rid of the need of the `placement` value inside `utils/getOffsets.js` (if possible)

Any help is welcome


FezVrasta:
This could be useful for the guys at https://github.com/twbs/bootstrap/issues/19796

FezVrasta:
Ok the feature seems to work now. I'll add some tests once I have some time and I'll merge it into `v1-dev` 💪😎

FezVrasta:
There's still that stupid test failing... the flip modifier kicks in because the popper offsets are not properly detected. Maybe we need to force a re-render before we can play with it... I'm not sure.

FezVrasta:
The tests introduced by #82 fails here, more investigation required.

FezVrasta:
Closing this PR, anyone interested feel free to open a new one to re-implement this stuff into the latest `v1-dev`. I'll take care of it once I have some free time otherwise.

PR #96, opened 2016-11-02
getBoundaries does not consider offsetParent left when calculating for an element

andrewnicols:
### CodePen demo

Basic example: https://codepen.io/anon/pen/bwPzXa

1.  Create a popper:
   1. on a page whose body has a left margin
   1. with a `boundariesElement` set to a _grandparent_ of the popper (not a direct parent) and which is not the body
   1. attached to a fixed positioned target

The popper should appear:
1. within the boundariesElement; and
2. directly under the target

### What went wrong?

The Popper + Arrow appear off to one side
On the example above, the margin on the body is not taken into consideration when looking at the boundary.

screen shot 2016-10-31 at 19 47 38

Manually adjusting the transform back to the left also show that the arrow position is incorrectly calculated.

screen shot 2016-10-31 at 19 48 16

### Any other comments?

Where the target is a child of the body, the behaviour is slightly different (though possibly the same):
https://codepen.io/anon/pen/jMjdgK

screen shot 2016-10-31 at 19 56 07

### Analysis

It appears as though the `getBoundaries` function is not considering the possibility of its outer rectangle being shifted in all scenarios.

In my local installation (moodle on master), I found that adding the offset parent's left (from getOffsetRect) to the boundaries.left and boundaries.right was enough to resolve the issue in the second codepen. I haven't had a chance to put this together into a codepen yet this evening.


andrewnicols:
I accidentally screwed up the pull request and made a pull request comparing master to this branch.

PR #92, opened 2016-10-18
fixes #75

FezVrasta:
The new test is still failing on IE10, 11 and Firefox. I'm working on it.

PR #97, opened 2016-11-02
Calculate boundaries relative to position (Fixes #96)

### LINKED ISSUE #96: getBoundaries does not consider offsetParent left when calculating for an element
### CodePen demo

Basic example: https://codepen.io/anon/pen/bwPzXa

1.  Create a popper:
   1. on a page whose body has a left margin
   1. with a `boundariesElement` set to a _grandparent_ of the popper (not a direct parent) and which is not the body
   1. attached to a fixed positioned target

The popper should appear:
1. within the boundariesElement; and
2. directly under the target

### What went wrong?

The Popper + Arrow appear off to one side
On the example above, the margin on the body is not taken into consideration when looking at the boundary.

screen shot 2016-10-31 at 19 47 38

Manually adjusting the transform back to the left also show that the arrow position is incorrectly calculated.

screen shot 2016-10-31 at 19 48 16

### Any other comments?

Where the target is a child of the body, the behaviour is slightly different (though possibly the same):
https://codepen.io/anon/pen/jMjdgK

screen shot 2016-10-31 at 19 56 07

### Analysis

It appears as though the `getBoundaries` function is not considering the possibility of its outer rectangle being shifted in all scenarios.

In my local installation (moodle on master), I found that adding the offset parent's left (from getOffsetRect) to the boundaries.left and boundaries.right was enough to resolve the issue in the second codepen. I haven't had a chance to put this together into a codepen yet this evening.

andrewnicols:


FezVrasta:
Looking at the tests seems like this change breaks almost anything...

andrewnicols:
Hi Fez,

I've fixed one of the tests and fixed my code for one other. I still see 19 failing tests, but these are also failing for me on the v1-dev branch. They're also failing on Travis: https://travis-ci.org/FezVrasta/popper.js/builds

The rest of the issues are because of reverting to using the viewport when the offsetParent _is_ the documentElement (https://github.com/andrewnicols/popper.js/commit/getBoundaries#diff-4a37a363f58aaf691b73b80bbcee90b8R58). I _believe_ that this is the correct behaviour, and when I view the tests in a browser, they appear to work as I would expect.

Any thoughts on this?

FezVrasta:
Yes you are right, we need to fix the build on `v1-dev` before we can merge any other PR, there is a PR to fix the tests (https://github.com/FezVrasta/popper.js/pull/92) but I have not time to finish it at the moment :(

src/utils/getBoundaries.js
```diff
@@ -51,11 +51,24 @@ export default function getBoundaries(popper, data, padding, boundariesElement)
...
+                boundaries.width = getScrollParent(popper).offsetWidth - getOffsetParent(popper).offsetLeft;
+                boundaries.right = boundaries.width;
+            }
         }
     }
     boundaries.left += padding;
     boundaries.right -= padding;
-    boundaries.top = boundaries.top + padding;
-    boundaries.bottom = boundaries.bottom - padding;
+    boundaries.top = boundaries.top + (padding * 2);
```
andrewnicols: Ah - that's probably me misunderstanding and getting it confused with the use of height. I'll remove that part.
I'll remove that shortly.

FezVrasta:
@andrewnicols may you try to open a PR targeting #92? I've the feeling that your code may fix the test that isn't passing on my branch.

andrewnicols:
Regarding the other change to use the viewport if the offsetParent is the documentElement, does this sound correct to you? It seems correct to me to make this change...

Without this change:
screen shot 2016-11-04 at 19 37 52

After this change:
screen shot 2016-11-04 at 19 38 20

I _think_ that this is correct, and it certainly works best in our examples. I think it also causes some unit test failures, but I think that the unit tests _could_ be wrong. I've put some of them into test files and they appear correctly in my opinion.

We (Moodle) have a release planned in 3 weeks and I'd really like to get this issue in our release, so if you're able to give an indication as to whether its correct aside from the unit test issues and the padding change I need to make, that would be awesome.

andrewnicols:
Re:
> @andrewnicols may you try to open a PR targeting #92? I've the feeling that your code may fix the test that isn't passing on my branch.

What do you mean? Are there new tests in there which you want me to run this change against?

FezVrasta:
@andrewnicols the PR #92 fixes all the tests except for one that has something wrong because of the way the viewport is calculated.

If you could base your edits on #92 you could fix the latest test and we could release the first stable beta.

FezVrasta:
@andrewnicols are you interested to modify your PR as I described?

andrewnicols:
I've been trying to look at doing that for the past few days but I'm still seeing 12 failures - all legitimate. The boundaries seem to be correct and I think it's still a flip issue.

When I was working on this I found:
* src/modifiers/flip.js - `bottom` is in both `a`, and `b` whilst `left` is in neither. I _think_ it belongs in `b`
* fixed positioned poppers with defined boundaries need to consider _both_ the viewport boundaries, and the specified boundary
* Viewport boundaries need to `Math.max` the left and top in some situations

I'm still trying to find time to look at the flip issues.
I've pushed my latest branch (getBoundariesPlus92)

andrewnicols:
https://github.com/andrewnicols/popper.js/commits/getBoundariesPlus92

FezVrasta:
Thanks for your work!

With the new features introduced by #92 it should be much easier to fix the problem we are facing, since now we have two different settings for the boundaries used by `flip`  and the ones used by `preventOverflow`.

1. That part of code could be rewritten to be clearer probably... They have not the same purpose, `a` is used to distinguish `right` and `bottom` from `left` and `top`.
    `b` is used to distinguish `vertical` from `horizontal`.
2.You are absolutely right, this is something missing right now that should be addressed.
3. I'm not sure what you mean with this, sorry.

Let me know if you need help with anything not clear with the code please.

FezVrasta:
@andrewnicols I'm writing to let you know that I'm working on this issue as well since I found some time to allocate to this project again.

If you want to share your progress we may find a solution faster.

FezVrasta:
Closing this in favor of `v1-dev-fix-everything` (I really like the branch name 😅)

In that branch I fixed almost all the tests, the only broken part is the flip modifier but I count to fix it in the next few days.

PR #105, opened 2016-11-22
Fix dom touching in destroy callback

lukebatchelor:
Hi, this fixes an issue where Popper would modify inline styles of an element during it's destroy lifecycle even when the applyStyle modifier is disabled.

A common scenario of this would be in React for instance, where the consumer needs to have full control of the DOM.


src/popper.js
```diff
@@ -320,11 +320,16 @@ export default class Popper {
      */
     destroy() {
         this.state.isDestroyed = true;
-        this.popper.removeAttribute('x-placement');
-        this.popper.style.left = '';
-        this.popper.style.position = '';
-        this.popper.style.top = '';
-        this.popper.style[getSupportedPropertyName('transform')] = '';
+        if (this.options.modifiers &&
```
FezVrasta: we have a "isModifierRequired" to make this check

lukebatchelor: I've had a look at IsModifierRequired, I don't think it does what I'm looking for. 

That would just check that a modifier exists, that doesnt tell you if a modifier is enabled or not though. 
The function itself seems kinda strange too?

```js
export default function isModifierRequired(modifiers, requesting, requested) {
    return !!modifiers.filter(modifier => {
        if (modifier.name === requested) {
            return true;
        } else if (modifier.name === requesting) {
            return false;
        }
        return false;
    }).length;
}
```

What purpose is `requesting` serving here? This function could be rewritten as

```js
export default function isModifierRequired(modifiers, requested) {
    return modifiers.some(modifier => modifier.name === requested);
}
```

and it would do the same thing right?

bradleyayers: You've done the correct solution here @lukebatchelor, good job.

lukebatchelor: I'm pretty sure what i wrote will have the same results as yours. Returning false if the modifier name is `requesting` and then returning false regardless is redundant. Sonic. The function in filter will only return true if the name is `requested` the only way for the function to return true is for that filter to exist in the list. 

If you wanna check that one canes before the other you'd do something like 

```js
cost requestedIdx = modifiers.indexOf(modifier => modifier.name === requested);
cost requestingIdx = modifiers.indexOf(modifier => modifier.name === requesting);
return requestedIdx >= 0 && requestedIdx < requestingIdx;
```

lukebatchelor: Writing on mobile so can't edit that, should be `findIndex` not `indexOf`

FezVrasta: Since we already have a function to perform this check (even if bugged), I prefer to have such function fixed and used here.

Repeating the same checks in multiple parts of the app will make it difficult to maintain the code.

For example, now we just have to change `isModifierRequired`, instead if I had done as you are suggesting, I would have needed to fix multiple occurrences of the same check.

lukebatchelor:
> Since we already have a function to perform this check (even if bugged), I prefer to have such function fixed and used here.

I'm still not sure this is correct? `isModifierRequired` is supposed to check if one modifier was in the modifiers list before the other and my code is checking if the applyStyle is enabled? 

I've fixed `isModifierRequired` and added `isModifierEnabled` if that helps?

lukebatchelor:
Hmm.. Seems to be failing on Travis for some reason. I just  pulled your latest changes into this and ran them locally, seems to work fine. Any known issues for these tests?

FezVrasta:
Sorry for the misunderstanding.

`isModifierRequired` had the purpose to check that a modifier was enabled and present before another given one.

The name derives by the fact that in v0 all the required modifiers were enabled by default and there was not the `enabled` flag.

You should simply replace all the instances of `isModifierRequired` with `isModifierEnabled` and get rid of the old function.

Thanks!

FezVrasta:
Fixed in bcca06adb8126d3b0dffc18d5c821483728d0a8b

I had to rebase lot of stuff and I tried to simplify the approach. Please let me know if it looks good.

PR #109, opened 2016-12-01
V1 dev fix everything

FezVrasta:
This PR aims to fix ALL the problems we faced with Popper.js v1 and make all the tests pass.

Right now ~9 tests are failing (one fails inconsistently and I suspect a problem with the test suite itself).

The other 8 tests are the `flip` modifier tests that are failing consistently in the whole test suite.


FezVrasta:
Now we have 3 tests failing plus the flaky one. 😁

FezVrasta:
only 1 test is failing now, and it's failing only on IE10.

I'm going to fix it tomorrow and then I'll probably release the last alpha 👍

FezVrasta:
So, I have not a Windows machine to run IE10 at the moment.

I have skipped that test on IE10 for now just to be able to merge these fixes and proceed with the works.

If someone has access to IE10 please help me fixing that test.

Issue #106, opened 2016-11-23
V1 docs are quite out of date.


lukebatchelor:
I'll preface this with, thank you so much for this library! Super awesome and SO much better than the alternatives.

I know you're already aware of this and are super busy at the moment, but just raising this for awareness.

We've been relying on the v1 dev branch for a long time now hoping for the v1 release as thats what had all the fixes we required. It's just super painful trying to dev at the moment when none of the docs match up and we have to go to the src to check how things work.

Things like: (it looks like) the boundariesElement API has changed, the offset API is commented in `modifiers/offset.js` but not in the docs anywhere, the flipVariation api is not documented anywhere, etc. 

It's just been a lot of paper cuts for us trying to use this.

Thanks again!

FezVrasta:
Hi, there is a command to generate an updated documentation. It should be `grunt docs` but please double check in the Gruntfile for confirmation.

If you want you can send a PR with the updated documentation. It would be very helpful.

Thanks!

lukebatchelor:
Oh! It's possible i wasn't generating them properly then!

I was cd'ing into docs and running `jekyll serve`

Definitely looking at pushing some docs for the things we've found. Cheers!

lukebatchelor:
So, having a look at this, the docs for the offset modifier /were/ in the docs, apologies. There were two copies, one as the options to  the new popper Instance and one as the Modifiers object. The modifiers one is correct.

We could easily copy and paste the offsets docs from the offset modifier but that doesnt seems scalable, keeping two sets of docs up to date per modifier. 

But perhaps there wont be ~too~ many changes and this is acceptable.

Just raising this here in case you have any ideas. Maybe just copying to main bits across (the API shape, type and default value) and linking the the relevant Modifer docs?

FezVrasta:
I have updated some docs, generated new markdown files that are now stored in `/doc` folder.
Added the docs links in the README and updated the README as well with new informations.
I also updated CONTRIBUTING.md following your indications from PR #110.

Is it enough to close this issue? If not, please reply here and I'll reopen it.

PR #110, opened 2016-12-03
Initial spike on docs updating

lukebatchelor:
So, I'm looking at helping out with #106 where I can, just looking to make sure I'm doing it right here.

This PR adds some npm commands that wrap the grunt commands, removing the need to install grunt globally (which is generally bad practice due to differences in grunt versions).  Running grunt directly from the CLI will use the users global grunt, using it from an npm script will use the one from the package.json.

I'm wondering what you think the best way for a dev to check the docs would be? If I recall correctly I was using `jekyll serve` from within the docs directory and that was working fine (on a different machine that had jekyll installed globally). 

Could we install jekyll as a devDep and add an npm script for that perhaps? I havent done it yet because I thought I'd check what you prefer first.

Cheers


CONTRIBUTING.md
```diff
@@ -47,14 +44,24 @@ npm test
...
+
+## Docs
+
+Documentation is generated by [JSDoc](http://usejsdoc.org/) annotations within the code. To build the docs simply run
+
+```bash
+npm run docs
+```
+
+You should then be able to see the generated markdown file in `doc/includes/documentation.md`.
```
lukebatchelor: I'd like a section here of how to serve the docs, any suggestion of the best way that doesnt involve globally installing anything?

src/modifiers/offset.js
```diff
@@ -6,6 +6,8 @@ import getPopperClientRect from '../utils/getPopperClientRect';
  * The offsets will shift the popper on the side of its reference element.
  * @method
  * @memberof Modifiers
+    * @param {Object} options.modifiers.offset - Offset modifier configuration
```
lukebatchelor: Actually, I'm just going to double check this is still the case, I made this change quite a while ago

lukebatchelor:
I'm not sure how I got so confused with the docs or where I added that "fix" from, it had been sitting in a git stash for a while now whilst I work on other things. My problem was simply that the docs hadnt been generated.

I'd still like to discuss the rest of this PR though so we can figure out the best way to work on docs changes. Should the generated docs be commited for example?

lukebatchelor:
Another question on this, how are you currently deploying the docs?

FezVrasta:
With v0 I was simply copying manually the generated json.

With v1 we'll serve the gh-page directly from the `/doc` folder thanks to a new feature of GitHub introduced few months ago.

CONTRIBUTING.md
FezVrasta: What's the problem with the markdown file? You can simply open it with any markdown viewer and it will look exactly like the docs on the website.

CONTRIBUTING.md
```diff
@@ -11,13 +11,10 @@ Feature requests are welcome!
 
 ## Setup
 
-Make sure to have `grunt-cli` and `bower` installed globally.
+Simply `nom install` and you are ready to go.
```
FezVrasta: Typo `nom/npm`

CONTRIBUTING.md
```diff
@@ -47,14 +44,24 @@ npm test
 **Note:** we use Chrome even for the tests on Travis CI, despite this, the CSS offsets of the page elements seem to be different when the tests are ran on Travis CI or on your local machine.
```
FezVrasta: Since we are already here, this section could be removed since we no longer perform different checks on TravisCI

FezVrasta:
@lukebatchelor any updates on this?

lukebatchelor:
Oh yes, sorry. I tried adding grunt-jekyll a while back but that seems painfully out of date. I'm still trying to figure out where the actual generated static pages end up that a dec could inspect. Is there a jekyll command you use?

I can fix the typos in this soon.

FezVrasta:
The workflow is the following:

```
grunt doc
cd /docs
jekyll serve
```

FezVrasta:
Read:
https://github.com/FezVrasta/popper.js/issues/106

Maybe it would be easier for you to create a new PR with only the changed to `npm` to get rid of the global `grunt` installation?

FezVrasta:
I just pushed some changes that get rid of Grunt completely, I'm going to close this PR.

PR #121, opened 2016-12-31
Add enable/disable method

sqal:
Added what we talked about in https://github.com/FezVrasta/popper.js/issues/117

- [x] new boolean option `enabled` where user can decide whether popper is enabled by default or not
- [x] added enable/disable methods
- [x] tests for the changes
- [ ] missing changes in the docs

### DIFF
--- a/src/popper/index.js
+++ b/src/popper/index.js
@@ -188,7 +188,6 @@ export default class Popper {
         this.state = {
             isDestroyed: false,
             isCreated: false,
-            isEnabled: this.options.eventsEnabled,
         };
 
         // get reference and popper elements (allow jQuery wrappers)
@@ -242,10 +241,13 @@ export default class Popper {
         // fire the first update to position the popper in the right place
         this.update();
 
-        if (this.state.isEnabled) {
+        const eventsEnabled = this.options.eventsEnabled;
+        if (eventsEnabled) {
             // setup event listeners, they will take care of update the position in specific situations
-            this.enable(true);
+            this.enable();
         }
+
+        this.state.isEnabled = eventsEnabled;
     }
 
     //
@@ -341,8 +343,8 @@ export default class Popper {
      * @method
      * @memberof Popper
      */
-    enable(initial) {
-        if (!this.state.isEnabled || initial) {
+    enable() {
+        if (!this.state.isEnabled) {
             setupEventListeners(this.reference, this.options, this.state, this.scheduleUpdate);
             this.state.isEnabled = true;
         }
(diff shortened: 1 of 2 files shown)


src/popper/index.js
```diff
@@ -329,6 +336,31 @@ export default class Popper {
     }
 
     /**
+     * Enables popper
```
FezVrasta: These comments will be converted into documentation, may you add more info to let the people understand what this method does?

src/popper/index.js
```diff
@@ -236,8 +242,10 @@ export default class Popper {
         // fire the first update to position the popper in the right place
         this.update();
 
-        // setup event listeners, they will take care of update the position in specific situations
-        setupEventListeners(this.reference, this.options, this.state, this.scheduleUpdate);
+        if (this.state.isEnabled) {
```
FezVrasta: Here it would make sense to find a way to use the `enable` method instead of duplicate the code.

src/popper/index.js
```diff
@@ -120,6 +123,8 @@ const DEFAULTS = {
  *      Placement of the popper accepted values: `top(-start, -end), right(-start, -end), bottom(-start, -right),
  *      left(-start, -end)`
  *
+ * @param {Boolean} options.enabled=true
+ *      Whether popper is initially enabled or not
```
FezVrasta: Please specify this refers only to the event listeners and not to the whole library.

It would probably make sense to rename the option to `eventsEnabled`?

src/popper/index.js
```diff
@@ -24,6 +24,9 @@ const DEFAULTS = {
     // placement of the popper
     placement: 'bottom',
 
+    // is initially enabled or not
```
FezVrasta: what? be specific please.

src/popper/index.js
```diff
@@ -236,8 +241,13 @@ export default class Popper {
         // fire the first update to position the popper in the right place
         this.update();
 
-        // setup event listeners, they will take care of update the position in specific situations
-        setupEventListeners(this.reference, this.options, this.state, this.scheduleUpdate);
+        const eventsEnabled = this.options.eventsEnabled;
+        if (eventsEnabled) {
+            // setup event listeners, they will take care of update the position in specific situations
+            this.enable();
```
FezVrasta: Last thing, wouldn't make sense to rename these to `enableEventListeners` and `disableEventListeners` to make it clear what they do? I know they are longer but I rather have a long method name than an unclear one.

Issue #128, opened 2017-01-05
Add more unit tests along the existing functional tests


FezVrasta:
Right now most of the stuff is tested through functional tests that query the DOM to check that the proper behavior is achieved.

We should add unit tests on the utils and the modifiers that don't rely on DOM, this would make the code much more reliable.

FezVrasta:
I added a coverage reported to make it easier to decide which modules need more tests:
https://codeclimate.com/github/FezVrasta/popper.js/coverage

PR #131, opened 2017-01-07
added sinon and chai as dev dependencies; added some unit tests

nadiam84:
This PR not exactly solves #128, but it's a start.

Added chai dependency because with it tests look a bit cleaner.
Also added sinon which is also very useful library for spying and mocking everything. I haven't exactly used it in the PR, it's an investment for the future.


FezVrasta:
Cool stuff, thanks!

Any reason why you added `sinon` and `chai` instead of using the built-in Jasmine utils?

nadiam84:
chai just looks nicer and more natural, almost lIke a real language. And sinon is much more powerful. I guess it's just a matter of preferences. Happy to rewrite those tests with built-in tools or change other tests to use them, for consistency. Your choice :)

FezVrasta:
I don't like Jasmine and I'm looking for alternatives, using `chai` and `sinon` is perfectly fine to me because they make it easier the transition to a different framework. I just wanted to know if there was any particular reason. Thanks!

PR #132, opened 2017-01-07
Fix JSDoc type for 'options.modifiers.preventOverflow.boundariesElement'.

DanielRosenwasser:
Was browsing through the docs, and this looked like an error. Hope I'm not mistaken. 😄


src/popper/index.js
```diff
@@ -89,7 +89,7 @@ const DEFAULTS = {
  * @param {Array} [options.modifiers.preventOverflow.priority=['left', 'right', 'top', 'bottom']]
  *      Priority used when Popper.js tries to avoid overflows from the boundaries, they will be checked in order,
  *      this means that the last one will never overflow
- * @param {Number} options.modifiers.preventOverflow.boundariesElement='scrollParent'
+ * @param {String|Element} options.modifiers.preventOverflow.boundariesElement='scrollParent'
```
FezVrasta: Thanks, definitely a mistake. I use the `HTMLElement` type in these cases. Are you able to update the PR using it instead of `Element`?

Thanks.

DanielRosenwasser:
Done - there was another `Element` lower in the docs so I fixed that up. Hopefully that was appropriate.

FezVrasta:
Absolutely, thanks you for the PR!

Issue #136, opened 2017-01-08
Popper without any parents


nadiam84:
When popper doesn't have any positioned or scrollable parents, its behavior on scroll is a bit broken. Examples:

Popper is okay, moves with ref element as expected:
https://codepen.io/nadiam84/pen/PWqVKo

Popper stays on the same place when scrolled, although it shouldn't:
https://codepen.io/nadiam84/pen/xgGMrz

The only difference in those two codepens is that the parent has 'position: relative'.

PR #137, opened 2017-01-08
fix popper's behavior when it doesn't have any parents (#136)

nadiam84:
Solves this: #136

### DIFF
--- a/src/popper/utils/getOffsetRectRelativeToCustomParent.js
+++ b/src/popper/utils/getOffsetRectRelativeToCustomParent.js
@@ -37,11 +37,10 @@ export default function getOffsetRectRelativeToCustomParent(element, parent, fix
         const offsetParent = getOffsetParent(element);
         const offsetParentIsHTML = offsetParent.nodeName === 'HTML';
         const scrollParentIsBody = scrollParent.nodeName === 'BODY';
-        /*
-         When a popper doesn't have any positioned or scrollable parents, the check getOffsetParent(element).contains(scrollParent)
-         will be 'false positive'. This is happening because getOffsetParent(element) returns 'html' node,
-         and scrollParent is the 'body' node. Hence the additional check.
-         */
+
+        // When a popper doesn't have any positioned or scrollable parents, `offsetParent.contains(scrollParent)`
+        // will return a "false positive". This is happening because `getOffsetParent` returns `html` node,
+        // and `scrollParent` is the `body` node. Hence the additional check.
         const isParentScrolledOrPositioned = !(offsetParentIsHTML && scrollParentIsBody);
         if (isParentScrolledOrPositioned && offsetParent.contains(scrollParent)) {
             const scrollTop = getScroll(parent, 'top');
--- a/src/popper/utils/getOffsetRectRelativeToCustomParent.js
+++ b/src/popper/utils/getOffsetRectRelativeToCustomParent.js
@@ -33,23 +33,17 @@ export default function getOffsetRectRelativeToCustomParent(element, parent, fix
         rect.bottom -= scrollTop;
         rect.left -= scrollLeft;
         rect.right -= scrollLeft;
-    } else {
-        const offsetParent = getOffsetParent(element);
-        const offsetParentIsHTML = offsetParent.nodeName === 'HTML';
-        const scrollParentIsBody = scrollParent.nodeName === 'BODY';
-
-        // When a popper doesn't have any positioned or scrollable parents, `offsetParent.contains(scrollParent)`
-        // will return a "false positive". This is happening because `getOffsetParent` returns `html` node,
-        // and `scrollParent` is the `body` node. Hence the additional check.
-        const isParentScrolledOrPositioned = !(offsetParentIsHTML && scrollParentIsBody);
-        if (isParentScrolledOrPositioned && offsetParent.contains(scrollParent)) {
-            const scrollTop = getScroll(parent, 'top');
-            const scrollLeft = getScroll(parent, 'left');
-            rect.top += scrollTop;
-            rect.bottom += scrollTop;
-            rect.left += scrollLeft;
-            rect.right += scrollLeft;
-        }
+    }
+    // When a popper doesn't have any positioned or scrollable parents, `offsetParent.contains(scrollParent)`
+    // will return a "false positive". This is happening because `getOffsetParent` returns `html` node,
+    // and `scrollParent` is the `body` node. Hence the additional check.
+    else if (getOffsetParent(element).contains(scrollParent) && scrollParent.nodeName !== 'BODY') {
+        const scrollTop = getScroll(parent, 'top');
+        const scrollLeft = getScroll(parent, 'left');
+        rect.top += scrollTop;
+        rect.bottom += scrollTop;
+        rect.left += scrollLeft;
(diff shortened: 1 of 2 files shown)


src/popper/utils/getOffsetRectRelativeToCustomParent.js
```diff
@@ -33,7 +33,12 @@ export default function getOffsetRectRelativeToCustomParent(element, parent, fix
         rect.bottom -= scrollTop;
         rect.left -= scrollLeft;
         rect.right -= scrollLeft;
-    } else if (getOffsetParent(element).contains(scrollParent)) {
+        /*
+         When popper doesn't have any positioned or scrollable parents, the check getOffsetParent(element).contains(scrollParent)
+         will be 'false positive'. Because getOffsetParent(element) returns html node, and scrollParent would be the body node.
+         Hence the additional check for nodeNames.
+        */
+    } else if (!(getOffsetParent(element).nodeName === 'HTML' && scrollParent.nodeName === 'BODY') && getOffsetParent(element).contains(scrollParent)) {
```
FezVrasta: `getOffsetParent` should be cached here instead of get called twice.

FezVrasta:
May you add a test for this case?

src/popper/utils/getOffsetRectRelativeToCustomParent.js
```diff
@@ -38,13 +38,16 @@ export default function getOffsetRectRelativeToCustomParent(element, parent, fix
...
-    } else if (!(getOffsetParent(element).nodeName === 'HTML' && scrollParent.nodeName === 'BODY') && getOffsetParent(element).contains(scrollParent)) {
-        const scrollTop = getScroll(parent, 'top');
-        const scrollLeft = getScroll(parent, 'left');
-        rect.top += scrollTop;
-        rect.bottom += scrollTop;
-        rect.left += scrollLeft;
-        rect.right += scrollLeft;
+    } else {
+        const offsetParent = getOffsetParent(element);
+        if (!(offsetParent.nodeName === 'HTML' && scrollParent.nodeName === 'BODY') && offsetParent.contains(scrollParent)) {
```
FezVrasta: Sorry if I'm being picky, I hope you don't mind.

`!(offsetParent.nodeName === 'HTML' && scrollParent.nodeName === 'BODY')` is quite confusing to read, I was thinking that we could split it into multiple variables to make it easier to understand what's happening here.

Below you find my attempt, I'm not sure if it's correct though...

```
const offsetParent = getOffsetParent(element);
const offsetParentIsHTML = offsetParent.nodeName === 'HTML';
const scrollParentIsBody = scrollParent.nodeName === 'BODY';
const isBodyChildOfHTML = offsetParentIsHTML && scrollParentIsBody;
if (!isBodyChildOfHTML && offsetParent.contains(scrollParent)) {
``

**Comment by nadiam84** [comment]
About test: I was planning to work on unit tests for the rest of the functions on the next weekend (can't do that sooner), so it will be covered in there. Writing a functional tests is a nightmare, maybe need to do something about them as well...

This fix doesn't break anything, all the existing tests are green locally.

File: src/popper/utils/getOffsetRectRelativeToCustomParent.js (hunk 2 above)
Comment (nadiam84) [inline src/popper/utils/getOffsetRectRelativeToCustomParent.js]: Don't mind of course. `isBodyChildOfHTML` doesn't make a lot of sense though, since body is always a child of html. But I see your point, I'll make it more readable.

**Comment by FezVrasta** [comment]
I was thinking about a simple functional test like the others we already have, unit tests make sense to me to check stuff that doesn't interact with DOM (IMHO) so I wouldn't bother trying to write an unit test for this case.

**Comment by FezVrasta** [comment]
By the way, SauceLabs is in maintenance until Monday so we have time before we are able to merge this PR.

**Comment by nadiam84** [comment]
I would strongly disagree with the 'simple, like others we have' :D There has to be a better way to do this...

Done, renamed and tested.

File: src/popper/utils/getOffsetRectRelativeToCustomParent.js [hunk 3]
```diff
@@ -33,13 +33,23 @@ export default function getOffsetRectRelativeToCustomParent(element, parent, fix
...
-        rect.right += scrollLeft;
+    } else {
+        const offsetParent = getOffsetParent(element);
+        const offsetParentIsHTML = offsetParent.nodeName === 'HTML';
+        const scrollParentIsBody = scrollParent.nodeName === 'BODY';
+
+        // When a popper doesn't have any positioned or scrollable parents, `offsetParent.contains(scrollParent)`
+        // will return a "false positive". This is happening because `getOffsetParent` returns `html` node,
+        // and `scrollParent` is the `body` node. Hence the additional check.
+        const isParentScrolledOrPositioned = !(offsetParentIsHTML && scrollParentIsBody);
```
FezVrasta: I don't get why this is called `isParentScrolledOrPositioned`? We don't check if the parent is scrolled nor if it's positioned...

FezVrasta:
@nadiam84 any idea to make the tests simpler is more than welcome 🙂

src/popper/utils/getOffsetRectRelativeToCustomParent.js
nadiam84: Because:
`getOffsetParent(element)` - returns a positioned parent or HTML if there is none.
`scrollParent = getScrollParent(parent)` - returns a scrollable parent or BODY if there is none.
Which means that if we want to know if our popper isn't inside some container that is positioned or scrollable, we need offsetParent to be HTML and scrollParent to be BODY.

`isScrolledParentBodyAndOffsetParentHtml` doesn't seem like a good idea.

FezVrasta: I see, but `isParentScrolled` means that we are checking that the `parent` element has a `scrollTop` or `scrollLeft` value different from `0`.

nadiam84: Yeah, you're right... Can you change it to something that you think is better?

src/popper/utils/getOffsetRectRelativeToCustomParent.js
```diff
@@ -33,7 +33,11 @@ export default function getOffsetRectRelativeToCustomParent(element, parent, fix
         rect.bottom -= scrollTop;
         rect.left -= scrollLeft;
         rect.right -= scrollLeft;
-    } else if (getOffsetParent(element).contains(scrollParent)) {
+    }
+    // When a popper doesn't have any positioned or scrollable parents, `offsetParent.contains(scrollParent)`
+    // will return a "false positive". This is happening because `getOffsetParent` returns `html` node,
+    // and `scrollParent` is the `body` node. Hence the additional check.
+    else if (getOffsetParent(element).contains(scrollParent) && scrollParent.nodeName !== 'BODY') {
```
FezVrasta: @nadiam84 does it make sense? I avoided to check that `offsetParent` is `html` because we already know that `scrollParent` is child of `offsetParent`.

There is the case where `body` is the `offsetParent` AND the `scrollParent`, do you know if this condition must evaluate to `true` or to` false` in this check?

nadiam84: `body doesn't contains itself` Actually it does ;) But all's good I think, it's the same situation. And tests are green :)

PR #146, opened 2017-01-13
fixed #143

### LINKED ISSUE #143: css overflow on the popper = bad placement
### CodePen demo

http://codepen.io/anon/pen/egZyZQ

1.  add the css property `overflow: auto` (or any other `overflow` value) on your popper div
2.  set the `placement` option to any value (ex: "left")
3.  the popper is always at the bottom left, whatever the value of placement. It gets even worse on "fixed" poppers because it will stay at the top of the window

Using placement should place the popper, even if it has an overflow set

### What went wrong?

The popper isn't positionned because of the "overflow" css property :'(

### Any other comments?

By the way, the link to the popper.js lib in your codepen example (given in the issue template) doesn't work: http://rawgit.com/FezVrasta/popper.js/master/src/popper.js --> 404
I used unpkg.com instead: https://unpkg.com/popper.js@1.0.2

FezVrasta:
fixes #143, please @ocombe may you test this as well and confirm that everything is working fine?

### DIFF
--- a/src/popper/utils/getBoundaries.js
+++ b/src/popper/utils/getBoundaries.js
@@ -1,5 +1,6 @@
 import getOffsetParent from './getOffsetParent';
 import getScrollParent from './getScrollParent';
+import getParentNode from './getParentNode';
 import getOffsetRect from './getOffsetRect';
 import getPosition from './getPosition';
 import getOffsetRectRelativeToCustomParent from './getOffsetRectRelativeToCustomParent';
@@ -45,7 +46,7 @@ export default function getBoundaries(popper, padding, boundariesElement) {
     else {
         let boundariesNode;
         if (boundariesElement === 'scrollParent') {
-            boundariesNode = getScrollParent(popper);
+            boundariesNode = getScrollParent(getParentNode(popper));
         } else if (boundariesElement === 'window') {
             boundariesNode = window.document.body;
         } else {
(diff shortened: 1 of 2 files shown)


ocombe:
Hello, yes it fixes my bug perfectly, thanks !

ocombe:
This fixed the problem with elements whose parents are in position fixed, but not for absolutely position.
Though I don't think it's the same bug, because I still have the problem even if I don't set the overflow property. I'll keep investigating until I find the cause, I'll you know (might be a problem on my part).

FezVrasta:
Thank you, feel free to open an issue if you can reproduce the problem on codepen.

ocombe:
So far, I found out that when using a placement of "bottom-end", "shift" returns the correct x offset (-138px for "end"), but then somewhere along the way it isn't applied (left is at "0", as if it was "bottom-start" instead of -138px for example)

ocombe:
Ok it's my fault, it's the prevent overflow method that uses the value "scrollParent" for the boundariesElement, when I want it to use the window instead :)

No problem then ! Waiting for the new release with this fix :)

FezVrasta:
Umh, do you mean that since the default value is `scrollParent`, it detects the popper as its own scrollParent?

ocombe:
No, it uses the direct parent, which in my case is a div that is much smaller than the popper (the div has the width of the button that shows the "dropdown" element).
Since it's smaller, it detects that the popper overflows on all sides, and resets the position to "left: 0".

FezVrasta:
I see, thank for the info!

PR #154, opened 2017-01-14
define popper alias only during tests (fix #144)

### LINKED ISSUE #144: Error when requiring with Webpack
Hello,

I installed both Popper.js and Tooltip.js, here is an extract from my package.json

```
"dependencies": {
    // ...
    "popper.js": "^1.0.2",
    "tooltip.js": "^1.0.2",
    // ...
}
```

I require Tooltip.js like this in my JS file:

```
const Tooltip = require('tooltip.js');
```

(same problem if I also instanciate Popper.js before)

I should receive no error and be able to use Tooltip.

### What went wrong?

When I start webpack, I get the following error in my console:

```
15:27:33 client.1 | Version: webpack 1.14.0
15:27:33 client.1 | Time: 14249ms
15:27:33 client.1 |              Asset     Size  Chunks             Chunk Names
15:27:33 client.1 |  webpack-bundle.js  3.55 MB       0  [emitted]  main
15:27:33 client.1 | webpack-styles.css  0 bytes       0  [emitted]  main
15:27:33 client.1 |    [0] multi main 76 bytes {0} [built]
15:27:33 client.1 |     + 328 hidden modules
15:27:33 client.1 |
15:27:33 client.1 | ERROR in ./~/tooltip.js/dist/tooltip.es5.js
15:27:33 client.1 | Module not found: Error: Cannot resolve 'file' or 'directory' ../popper/index.js in /vagrant/client/node_modules/tooltip.js/dist
15:27:33 client.1 |  @ ./~/tooltip.js/dist/tooltip.es5.js 26:90-119
```

Thanks in advance!

FezVrasta:
This should fix problems with webpack. ping @Spone

To test this out you have to clone the repository, run `yarn build:tooltip` (or `npm run build:tooltip`) and copy the `dist/tooltip*` to `node_modules/tooltip.js/dist`.

Please let me know if this fixes the problem once for all.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,10 +1,7 @@
 {
     "presets": ["stage-2"],
     "plugins": [
-        "external-helpers",
-        ["module-alias", [
-            { "src": "./src/popper/index.js", "expose": "popper.js" }
-        ]]
+        "external-helpers"
     ],
     "env": {
         "coverage": {
--- a/scripts/karma.conf.js
+++ b/scripts/karma.conf.js
@@ -69,6 +69,11 @@ module.exports = function(config) {
                     ['es2015', { modules: false }],
                     'stage-2',
                 ],
+                plugins: [
+                    ['module-alias', [
+                        { src: './src/popper/index.js', expose: 'popper.js' },
+                    ]],
+                ],
             })],
         },
         files: [


FezVrasta:
This should fix problems with webpack. ping @Spone

To test this out you have to clone the repository, run `yarn build` (or `npm run build`) and copy the `dist/popper*` to `node_modules/popper.js/dist` and `dist/tooltip*` to `node_modules/tooltip.js/dist`.

Please let me know if this fixes the problem once for all.

FezVrasta:
it should work because now the header of tooltip.js is:

```js
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory(require('popper.js')) :
  typeof define === 'function' && define.amd ? define(['popper.js'], factory) :
  (global.Tooltip = factory(global.Popper));
}(this, (function (Popper) { 'use strict';
```

Previously here we had `../popper/index.js`

Spone:
👍 Yes it works! I no longer have the error! Thanks

FezVrasta:
Just released tooltip.js@1.0.5 with this fix.

PR #157, opened 2017-01-15
refactor half of the core tests

nadiam84:
This is more of a suggestion on how functional tests could be refactored. The initial idea was to use `jsx` instead of a custom wrapper, but I wasn't able to make it work with your building tools :( If someone can enable it, then those tests can be easily refactored to use it.

I'll finish with the rest of the tests on the next week, if you approve of the changes.


nadiam84:
Not sure about broken IE10 tests though... The error is `ReferenceError: 'Symbol' is undefined`, but there is no Symbol anywhere in the code. It's only in the minified js file, so maybe some uglifying/transpiling weirdness.

FezVrasta:
The `Symbol` error is thrown by a missing polyfill, the Popper.js build doesn't include the Babel polyfills.

FezVrasta:
I'm not sure I'd like to add VDOM in the equation, maybe a better alternative would be to use Preact that doesn't rely on a virtual DOM algorithm?

leeoniya:
Preact does use VDOM, as do all the fastest libs. No need to be scared of it :)

https://rawgit.com/krausest/js-framework-benchmark/master/webdriver-ts/table.html

FezVrasta:
I remember it differ the real DOM 🤔

My concern is that I don't want to step into troubles introduced by async operations or similar stuff. But I may be totally wrong,

FezVrasta:
Ah, here were I read about it:

> Inferno is much faster than Preact in rendering, updating and removing elements from the DOM. Inferno diffs against virtual DOM, rather than the real DOM (except for when loading from server-side rendered content) which means it can make drastic improvements. Unfortunately, diffing against the real DOM has a 30-40% overhead cost in operations.

https://github.com/infernojs/inferno#differences-from-preact

leeoniya:
you're right, it does look like it diffs a temporary vdom against the real dom [1]. but there is no speed benefit to this, as the Inferno docs point out.

> My concern is that I don't want to step into troubles introduced by async operations or similar stuff. But I may be totally wrong,

this isn't the result of them using vdom, it's just the way many libs choose to behave. there are some that allow for synchronous behavior, if you need this.

[1] https://github.com/developit/preact/blob/master/src/vdom/diff.js

nadiam84:
Returning to the matter of missing Symbol :) I don't use it, and some of the tests that are failing are the old tests, I haven't touched them yet (other than changing `var` to `const`).

@FezVrasta should I continue with refactoring as it is in the PR? Or should I wait until Preact/vdom is enabled for tests?

FezVrasta:
The Symbol error is probably thrown by React that internally uses something. To note that `Symbol` isn't directly used, if you use `for...in` you get an error with `Symbol` as well.

FezVrasta:
I tried to enable Preact + JSX but I couldn't get it to work 🤕

nadiam84:
I'm super busy at the moment, so I won't be able to continue with this PR for at least two weeks :( If anyone wants to pick it up it would be awesome.

FezVrasta:
No problem, the library is very stable at the moment, this is just an enhancement but nothing critical, take your time. 🙂

PR #165, opened 2017-01-20
Docs: Fixed default value for options.template

jvorcak:
This fixes a markup in the docs (default value for options.template) https://popper.js.org/tooltip-documentation.html


FezVrasta:
These files are automatically generated by [jsdoc-to-markdown](https://www.npmjs.org/package/jsdoc-to-markdown), we can't edit them directly.

PR #166, opened 2017-01-25
[fix]: Importing into universal apps

HHogg:
Currently the [util/debounce.js](https://github.com/FezVrasta/popper.js/blob/master/src/popper/utils/debounce.js) file is accessing `navigator` and `window` objects at the time of import. This causes universal applications that involve any server side rendering/static generation to fail on importing Popper.js. 

I understand this is a client side library, and I'm not entirely sure where the responsibility should be placed to gracefully handle this, but conditional imports are nasty, and I don't think attaching mock objects to `global` should be the way to go. 

_If_ you're ok with this fix, as for testing the _"universalness"_ in the future, I don't know much about Karma (I'm a mocha man) but a very quick search and I can't see a way to emulate a non-browser env?

### DIFF
--- a/src/popper/utils/debounce.js
+++ b/src/popper/utils/debounce.js
@@ -1,9 +1,10 @@
 import isNative from './isNative';
 
+const isBrowser = typeof window !== 'undefined';
 const longerTimeoutBrowsers = ['Edge', 'Trident', 'Firefox'];
 let timeoutDuration = 0;
 for (let i = 0; i < longerTimeoutBrowsers.length; i += 1) {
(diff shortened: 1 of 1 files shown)


FezVrasta:
LGTM, thanks!

We haven't any way to test the library outside of browser scope, but it's on the roadmap.

HHogg:
Awesome thanks @FezVrasta  👍 I'll have a look more into it to something that easily fits in with the current project setup. 

Is there a planned release, where this fix will be included?

FezVrasta:
I'm going to release it this weekend probably.

PR #175, opened 2017-02-22
Make navigation immediately visible on desktop

silvenon:
As someone who frequently visits the documentation, I would like to reduce the path to navigation by one click. What do you think?

screen shot 2017-02-22 at 22 36 19

screen shot 2017-02-22 at 22 36 31

On smaller devices it looks the same as before.

### DIFF
--- a/docs/css/main.scss
+++ b/docs/css/main.scss
@@ -1341,35 +1341,51 @@ button,
 }
 
 #menu {
-	@include vendor('transform', 'translateX(20em)');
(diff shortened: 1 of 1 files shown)


FezVrasta:
what about small desktops?

silvenon:
On smaller desktops it collapses to the off-canvas menu. Do you think that I should create a solution for smaller desktops as well?

FezVrasta:
Looks great thanks :)

PR #178, opened 2017-02-23
[fix] Calculate position to documentRoot instead of offsetParent (fixes #177)

### LINKED ISSUE #177: Popper position calculation fails if offsetParent is inside an offsetParent
We tried already to narrow the bug down and opened #174 however we didn't manage to reproduce the real underlying problem.

### CodePen demo

https://codepen.io/claudiobmgrtnr/pen/QpWNVL

popper-offsetparentbug

1.  Create a `wrapper` and a `outer-wrapper` with `position:relative`
2.  Add a `ref` and `pop` with `position:relative`
3.  Pop the popper: 

```js
new Popper(ref, pop, {
  modifiers: {
    preventOverflow: {
      boundariesElement: 'viewport',
    }
  }
})
```

Popper keeps tooltip element into the viewport

### What went wrong?

Tooltip is outside the viewport

### Possible solution

After looking into popper.js we found the root for this issue here:

https://github.com/FezVrasta/popper.js/blob/f5a309e2cf56064ffca3a2c64796aed67e83f367/src/popper/utils/getOffsetRect.js#L23-L26

```js
   left: element.offsetLeft, // Relative to the offsetParent not to the viewport
   top: element.offsetTop, // Relative to the offsetParent not to the viewport
```

We created an additional function for the cases where the position relative to the viewport was needed and could fix the error.

Do you want us to create a merge request for this?

We are not 100% sure but this might also fix #172

claudiobmgrtnr:
For details please see #177

### DIFF
--- a/src/popper/utils/getOffsetRectRelativeToViewport.js
+++ b/src/popper/utils/getOffsetRectRelativeToViewport.js
@@ -1,0 +1,23 @@
+import getOffsetRect from './getOffsetRect';
+import getOffsetParent from './getOffsetParent';
+
+export default function getOffsetRectRelativeToViewport(element) {
+    // Offset relative to offsetParent
+    const relativeOffset = getOffsetRect(element);
+
+    if (element.nodeName !== 'HTML') {
+        const offsetParent = getOffsetParent(element);
+        const parentOffset = getOffsetRectRelativeToViewport(offsetParent);
+        const offset = {
+            width: relativeOffset.offsetWidth,
+            height: relativeOffset.offsetHeight,
+            left: relativeOffset.left + parentOffset.left,
+            top: relativeOffset.top + parentOffset.top,
+            right: relativeOffset.right - parentOffset.right,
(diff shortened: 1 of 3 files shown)


src/popper/utils/getOffsetRectRelativeToViewport.js
```diff
@@ -0,0 +1,23 @@
+import getOffsetRect from './getOffsetRect';
+import getOffsetParent from './getOffsetParent';
+
+export default function getOffsetRectRelativeToViewport(element) {
+  // Offset relative to offsetParent
```
FezVrasta: silly question, may you fix the indentation here?

claudiobmgrtnr: Sure thing. Should be OK now.

jantimon:
@FezVrasta do you already know when this change will be released?

FezVrasta:
Released as 1.0.7

jantimon:
Wow thanks 🎉

Issue #151, opened 2017-01-13
Popper doesn't listen the `scroll` event of all the scrollable parents.


FezVrasta:
Right now if the popper is child of a scrollable parent AND of a scrollable body (or another scrollable custom parent), only the first scrollable parent will have the event listeners attached to it.

It should attach the event listeners to all the scrollable parents.

This is the interested module:
https://github.com/FezVrasta/popper.js/blob/master/src/popper/utils/setupEventListeners.js

PR #182, opened 2017-03-11
[fix]: Fixed scroll event listeners (#151)

FezVrasta:
Added a test to make sure we don't have regressions in future and bumped version.

### DIFF
--- a/tests/functional/core.js
+++ b/tests/functional/core.js
@@ -1073,4 +1073,33 @@ describe('[core]', () => {
             },
         });
     });
+
+    it('checks that all the scrollable parents have an event listener attached', (done) => {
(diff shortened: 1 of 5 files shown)


tests/functional/core.js
```diff
@@ -1073,4 +1075,35 @@ describe('[core]', () => {
             },
         });
     });
+
+    it('checks that all the scrollable parents have an event listener attached', () => {
```
FezVrasta: @nadiam84 I experimented with this method to define the templates of the tests. To me it looks like a good compromise between JSX and pure DOM API.

What do you think?

nadiam84: Yeah, I agree. And this looks much better than what I had in my forgotten PR :)

Issue #172, opened 2017-02-20
Popper content inside translateX parent causes misalignment


bengummer:
### CodePen demo

http://codepen.io/bgummeratlassian/pen/egwEjZ

### Steps to reproduce the problem

(see codepen demo)

1. Place the trigger and content as siblings, in the same parent
2. Position the parent at the bottom of the screen, with translateX style
3. Create a new Popper instance for the trigger and content, with `placement: top`

### What is the expected behavior?

Popper content should be shown above the button, which is at the bottom of the page

### What went wrong?

With translateX on the parent, the popper content is shown underneath the button even though we have `placement: top`

### Any other comments?

In the codepen example, disabling the `translateX` style (then refreshing the page and clicking the button) works as expected.

nblackburn:
Getting this issue too

FezVrasta:
@nblackburn please may you open a new issue and post a minimal codepen example without any extra dependencies? Thank you!

nblackburn:
@FezVrasta The dependencies were added to make it realistic to the scenario i was faced with so you could see what i was seeing.

FezVrasta:
yup I get it, but please may you open a new issue with a simplified codepen? 🙂

nblackburn:
Do we have a ETA on the fix for this?

FezVrasta:
It would be easier for me to work on this if someone could send a PR that adds a failing test for this specific bug in our test suite.

By the way, nope, I haven't an ETA, I don't work full-time on this project and I try to allocate some time on it during my free time.

PRs are super welcome.

FezVrasta:
I'm merging this into #202 since they are strictly related.

ventralnet:
I had a similar problem to this if a parent node had overflow: auto;

PR #192, opened 2017-03-27
added a test to show a bug with translateX

nadiam84:
This is a test that shows the #172 issue


tests/functional/core.js
```diff
@@ -1001,4 +1001,26 @@ describe('[core]', () => {
             },
         });
     });
+
+    it('inits a popper near a reference element, both inside an element with CSS translateX', (done) => {
```
FezVrasta: May you `xit` this?

nadiam84:
@FezVrasta Did a bit of a debugging. Turns out this bug happens when:
1. parent is 'fixed' AND
2. parent is transformed AND
3. preventOverflow modifier is enabled :)
Here is the codepen: http://codepen.io/nadiam84/pen/gmZwbG

Not sure how to fix it yet, maybe you can figure it out quicker.

I can update the test here as well if its easier for you.

nblackburn:
@nadiam84 I knew it was to do with the parent being fixed, glad we have confirmation of this now.

PR #196, opened 2017-04-02
[fix] Properly compute positions when HTML has margin (fix #191)

### LINKED ISSUE #191: Incorrect positioning when margin is applied to HTML
### CodePen demo

https://codepen.io/miljan/pen/VpBMPv

1.  Open the pen and check the positions

#pop to be positioned below #ref 

### What went wrong?

The #pop gets positioned as if there was no margin applied to HTML, to where the #ref would be in that case.

### Any other comments?

Many offcanvas animations apply margin to html which brakes all popper positions.

FezVrasta:
Fixes #191


FezVrasta:
@miljan-aleksic please may you confirm that this fixes the bug on your side?

miljan-aleksic:
@FezVrasta, fix confirmed. Thanks :)

PR #197, opened 2017-04-06
[feature] Added ReferenceClass constructor

### LINKED ISSUE #193: Add documentation to use reference object as replacement to DOM node
It would be nice if a function could be passed instead of an html element for `reference` at init.  The function would return a bounding box that could be used for positioning.

The function is necessary --- rather than just a plain object --- to allow for cases where the bounding box would change over time.  This would allow updates with `popper.update()`.

Having this feature would make support for [libs like Cytoscape](https://github.com/nrnb/GoogleSummerOfCode/issues/84) much better.

Thanks for your consideration.

FezVrasta:
close #193

Please do let me know if the documentation is clear enough.

The usage is:

```
import Popper, { ReferenceClass } from 'popper.js';

function getRect() {
  return {
    top: 0,
    right: 100,
    bottom: 100,
    left: 0,
    width: 100,
    height: 100,
  };
}

const reference = new ReferenceClass(getRect);

new Popper(reference, document.querySelector('#popper'));
```

I still have to add some tests to cover this.


docs/_includes/popper-documentation.md
```diff
@@ -1,6 +1,12 @@
 ## Classes
 
 <dl>
+<dt><a href="#Reference">Reference</a></dt>
+<dd><p>This class can provided to Popper.js as <code>reference</code> argument instead of
```
bedeoverend: Might be clearer if this said: "An instance of this class can be provided"? Rather than "This class can be provided". I'm being nitpicky I know, sorry, but I feel it might things a little clearer.

Same with further down a little.

bedeoverend:
Docs look good 👍 easy to understand what's going on IMO. Just had one comment, but that was just me being nitpicky - LGTM

FezVrasta:
Closing this, we want just new docs.

Issue #198, opened 2017-04-07
Various questions regarding the position property.


QuentinLB:
Hi,

I am currently working on a quite complex UI with custom dropdown lists placed with popper.js and I have few issues and questions regarding the fixed vs. absolute positionning.

Here is my layout :

![image](https://cloud.githubusercontent.com/assets/6490070/24795745/b79475c8-1b8a-11e7-9879-a26e07a22081.png)

My two main issues :

1. When I setup a popper in the scrolling area (pink), it is lagging quite a lot during scrolls. The positionning is good but there is this one second lag that make the UX feel really sluggish. 
If I remove the `position: fixed;` on the page container (brown) the lag disappear but it messes up my layout. 
2. In nested absolutely positionned elements popper.js seems to get lost and put the popper in a wrong place.

I tried to replicate the conditions in a codepen https://codepen.io/anon/pen/ryEBmb but it is not lagging there.

Since I cannot provide much more information, I was wondering if the community could help me with some known issues or diagnosis.

On another note I was also wondering, why does the fixed position take priority on absolute positionning. What I mean is that if the popper has a common parent (relative) with the target it should use absolute. I am confident you did this for good reasons but by tweaking the position manually I achieve greater consistency especially in my second issue.

I am grateful for this is well rounded tool and thanks for looking into this. 👍

FezVrasta:
Hi, regarding the fixed vs absolute inside a fixed parent, the choice was made because usually (at least, in my case), fixed elements aren't scrollable (they are headers, sidebars etc) and then having a `fixed` position on the popper as well reduces the need for computations when the popper isn't child of the `fixed` element. I agree that this could be tweaked to make it behave differently to suite your use case.

About point 2, without a reproducible pen I can't help you to understand where's the problem unfortunately.

QuentinLB:
Thank you for your answer.

For the second point I made another pen but as for the previous one it is still not behaving as my app.
http://codepen.io/anon/pen/KWjMMj
In my app the popper is misplaced (overlapping)

I keep looking and I'll let you know if I find anything that could improve popper.

FezVrasta:
Unfortunately if you can't reproduce the problem on the pen I can't do anything 😥

QuentinLB:
I posted these just in case you recognized an anti-pattern or a known issue. Thank you though for taking a look at it.

FezVrasta:
[...]

PR #201, opened 2017-04-10
[fix] Rewritten the algorithm used to get the reference position

### LINKED ISSUE #171: placement: 'top' - does not flip the popper when it touches the upper viewport boundary
1.  Visit https://popper.js.org
2.  In the first example (#example10positionSelector) choose 'top'
3.  Scroll down

Popper should be placed under the trigger, as soon as the upper viewport boundary is hitting the popper's top border.

### What went wrong?
Popper is not flipped to bottom.

### Any other comments?
Choosing placement: "bottom" works as expected

### LINKED ISSUE #188: Positioning incorrect when popper inside relatively or absolutely positioned parent
### CodePen demo

http://codepen.io/anon/pen/gmRRwz

1. Instantiate a new popper, with default settings, inside of a relatively or absolutely positioned  parent.
2. Scroll or resize the page so that popper hits the bottom of the viewport, thus causing it to flip to the other side.

Initial placement should be correct.

screen shot 2017-03-13 at 8 18 46 am

The popper should flip to the opposite side of the target.

screen shot 2017-03-13 at 8 14 14 am

### What went wrong?

The initial placement is incorrect.

screen shot 2017-03-13 at 8 17 47 am

The popper flips, but the placement is incorrect.

screen shot 2017-03-13 at 8 14 28 am

### Any other comments?

Not at this time.

FezVrasta:
This work is needed to:

1. fix #188
1. fix #171
1. Probably #198
1. (bonus) reduce size
1. (bonus) reduce complexity

Some/lot of tests are failing, any help to fix them is super welcome (really!). Just fork this branch, apply your changes, and send a PR against this branch.

Thank you.


src/popper/utils/findCommonOffsetParent.js
```diff
@@ -0,0 +1,21 @@
+import isOffsetContainer from './isOffsetContainer';
+
+export default function findCommonOffsetParent(element1, element2) {
```
FezVrasta: One of the Popper problems was that we didn't consider that reference and popper could live in two different offset contexts.

Now we find the common offset parent and use it as reference to compute all the positions.

src/popper/utils/getOffsetRectRelativeToArbitraryNode.js
```diff
@@ -0,0 +1,28 @@
+import getStyleComputedProperty from './getStyleComputedProperty';
+
+export default function getOffsetRectRelativeToArbitraryNode(children, parent) {
```
FezVrasta: This is basically an improved version of `getOffsetRectRelativeToCustomParent`

FezVrasta:
There are only 3 tests failing in the `core` suite and few failing in `Tooltip.js`.

The ones in `core` are about Shadow DOM and they are related to the new way I use to find the common offset parent, I think it should be quite easy to workaround tho.

I had no time to look at the Tooltip.js tests, but I don't expect them to be broken in bad ways...

tests/functional/flips.js
```diff
@@ -166,13 +166,11 @@ describe('[flipping]', () => {
 
     xit('flips to opposite side when rendered inside a positioned parent', (done) => {
```
nadiam84: Should this be un-xit also? :)

src/popper/modifiers/preventOverflow.js
```diff
@@ -24,8 +24,9 @@ import getBoundaries from '../utils/getBoundaries';
  */
 export default function preventOverflow(data, options) {
     const boundariesElement = options.boundariesElement || getOffsetParent(data.instance.popper);
-    const boundaries = getBoundaries(data.instance.popper, options.padding, boundariesElement);
+    const boundaries = getBoundaries(data.instance.popper, data.instance.reference, options.padding, boundariesElement);
     options.boundaries = boundaries;
+    console.log('boundaries', {...boundaries});
```
nadiam84: Should it be here? :)

FezVrasta: It's still under development, I'll clean everything later

tests/functional/flips.js
FezVrasta: I'm going to work on these once I get the existing tests passing

FezVrasta:
Fixed the tooltips tests, now we have only the 3 shadow DOM tests failing (wuut!)

FezVrasta:
So, 1 test is failing on Firefox and Chrome on CI (no idea why, locally they work...) and some are failing on IE and Edge. Will look at them tomorrow. Feel free to pick up the branch and fix them if you feel in mood. (I'll stop asking it at some point...)

FezVrasta:
The failing test on Chrome and Firefox has been fixed (even if not really cleanly actually...), now I'm working on the latest `x`'ed test.

FezVrasta:
And #188 is fixed as well.

img

Now, it's your turn Internet Explorer.

FezVrasta:
So, IE11 and Edge now pass. But IE10 is hard to die.

I can't work on it anymore for today, again, if someone (maybe someone that uses this library in its multi billionaire company, Microsoft, anyone?) could spend some time and fix it for me. 🙂

FezVrasta:
Seems like this PR introduced a regression, note that it doesn't take margins of `body` into account:
https://codepen.io/FezVrasta/pen/qmEObw?editors=1100

FezVrasta:
I thought I was screwed because I used the "party hard" gif in a previous message, but luckily now that **all our test suite pass** I found this one:

img

PR #227, opened 2017-04-23
[fix] Allow negative %/vh/vw offsets

FezVrasta:
Previously there was a bug that prevented to use a negative offset value, e.g. `-50%`.

### DIFF
--- a/tests/functional/offset.js
+++ b/tests/functional/offset.js
@@ -20,7 +20,7 @@ describe('[offset]', () => {
           offset: offset,
         },
       },
-      onCreate: () => {
+      onCreate: (data) => {
         const refLeft = reference.getBoundingClientRect().left;
         const refWidth = reference.offsetWidth;
         const popperLeft = popper.getBoundingClientRect().left;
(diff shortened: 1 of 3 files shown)


tests/functional/offset.js
```diff
@@ -124,6 +156,43 @@ describe('[offset]', () => {
...
+          offset: offset,
+        },
+        flip: { enabled: false },
+      },
+      onCreate: (data) => {
+        const refLeft = reference.getBoundingClientRect().left;
+        const refBottom = reference.getBoundingClientRect().bottom;
+        const refWidth = reference.offsetWidth;
+        const refHeight = reference.offsetHeight;
+        const popperLeft = popper.getBoundingClientRect().left;
```
shmax: You could have done a single call to `getBoundingClientRect`, no?

FezVrasta: That's just a test and I don't care about performance in the code I write in the tests, but I agree it would have been cleaner.

shmax: Ah, so it is. Yeah, I don't care about performance in tests, either.

PR #229, opened 2017-04-24
[feature] Allow math operations in offset modifier

### LINKED ISSUE #228: Allow basic math operations in offset modifier
The offset modifier should allow to define an offset with this fingerprint:

```
'10px + 15vh - 10 20% + 4vw - 4px'
```

I'm already working on it.

FezVrasta:
fix #228, pinging @shmax for review.

New documentation:

> The `offset` modifier can shift your popper on both its axis.
> 
 > It accepts the following units:
>  - `px` or unitless, interpreted as pixels
>  - `%` or `%r`, percentage relative to the length of the reference element
>  - `%p`, percentage relative to the length of the popper element
>  - `vw`, CSS viewport width unit
> - `vh`, CSS viewport height unit
> 
>  For length is intended the main axis relative to the placement of the popper.
>  This means that if the placement is `top` or `bottom`, the length will be the
>  `width`. In case of `left` or `right`, it will be the height.
> 
>  You can provide a single value (as `Number` or `String`), or a pair of values
>  as `String` divided by a comma or one (or more) white spaces.
> The latter is a deprecated method because it leads to confusion and will be
removed in v2.
>  Additionally, it accepts additions and subtractions between dfferent unit.
>  Note that multiplications and divisions aren't spported.
> 
>  Valid examples are:
>  ```
>  10
>  '10%'
>  '10, 10'
>  '10%, 10'
>  '10 + 10%'
>  '10 - 5vh + 3%'
>  '-10px + 5vh, 5px - 6%'
>  ```

### DIFF
--- a/tests/unit/parseOffset.js
+++ b/tests/unit/parseOffset.js
@@ -1,0 +1,46 @@
+import { parseOffset } from 'src/popper/modifiers/offset';
+
+const popperOffsets = {
+  top: 0,
+  left: 0,
+  right: 100,
+  bottom: 100,
+  width: 100,
+  height: 100,
+};
+
+const referenceOffsets = popperOffsets;
+
+describe('parseOffset', () => {
+  it('parses `10px 10px` and throws deprecation warning', () => {
+    console.warn = jasmine.createSpy('warn');
+    expect(
+      parseOffset('10px 10px', referenceOffsets, popperOffsets, 'bottom')
+    ).toEqual([10, 10]);
+    expect(console.warn).toHaveBeenCalledWith(
+      'Offsets separated by white space(s) are deprecated, use a comma (,) instead.'
+    );
+  });
+
+  const cases = [
+    ['10,10', [10, 10]],
+    ['10px, -10px', [10, -10]],
+    ['10px, 10px', [10, 10]],
+    ['10px + 5%, 10px - 5%', [10 + 5, 10 - 5]],
+    ['-10px + 5%, -10px - 5%', [-10 + 5, -10 - 5]],
+    ['5%p, 5%r', [5, 5]],
+    ['100%+50px, 100px-10+10%', [100 + 50, 100 - 10 + 10]],
+    ['100% + 50% + 10%, 100%', [100 + 50 + 10, 100]],
+    ['100%p - 10%r, 100%', [100 - 10, 100]],
+    ['100%p - -10%r, 100%', [100 - -10, 100]],
+    ['+10px - -10% + -10%r', [10 - -10 + -10, 0]],
+  ];
+
+  cases.forEach(([input, output]) => {
+    it(`parses '${input}'`, () => {
+      expect(
+        parseOffset(input, referenceOffsets, popperOffsets, 'bottom')
+      ).toEqual(output);
+    });
+  });
+});
--- a/docs/_includes/popper-documentation.md
+++ b/docs/_includes/popper-documentation.md
@@ -740,7 +740,7 @@ It can be one either `-end` or `-start`.
 The `offset` modifier can shift your popper on both its axis.
 
 It accepts the following units:
-- unitless, interpreted as pixels
+- `px` or unitless, interpreted as pixels
 - `%` or `%r`, percentage relative to the length of the reference element
 - `%p`, percentage relative to the length of the popper element
 - `vw`, CSS viewport width unit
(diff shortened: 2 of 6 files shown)


shmax:
I'll have to look at this in more detail after work, but my first impression is that the syntax for the compound units is a bit loose and your tests for this feature could be a bit more comprehensive. 

For example, you have this in your tests: `5 - 25% 10px + 25%`. Visually, it's a bit difficult to parse; there are 5 spaces in it, but the third space is meant to be a divider between the axes. It's not too difficult to imagine it getting even more elaborate:

`5 - 25% + 5 -10px - 10 + 25%`

If you're absolutely sure that your regex can handle it then okay, but I'm wondering if it might be helpful to make the syntax a little tighter. Some examples:

`5-25%+5 -10px+25% // allow only one space` 

`5 - 25% + 5, -10px + 25%` // require a comma

As for your tests, I would probably think about adding some unit tests (not functional) and really beating on your parsing function. Feed it all kinds of weird scenarios.

And finally a comment on the tests you added in this PR: they seem to have a little too much logic going on. Your `expectedPopperLeft` value, for instance, is constructed from a fairly elaborate mathematical expression, which means it could have bugs. You should strive to remove any trace of logic from your tests, make them more comprehensive, and easier to read.

Need to get to work, but this is a great start!

FezVrasta:
About the regex, I'm quite confident it can handle the cases you mentioned (I tried it). I would like to make it simpler using an array or an object with `x` and `y` (or a comma separated as you said). But it would break the current API and I don't want to release a new major version for this change.

About adding unit tests you are right, it would make it easier to test the function, I'll add them.

FezVrasta:
@shmax I added some unit tests and I also deprecated the space separated offset in favor of a comma separated string.

If you use the space as separator it will throw a warning but will still work.

tests/unit/parseOffset.js
```diff
@@ -0,0 +1,70 @@
...
+
+  it('parses `10px + 5%, 10px - 5%`', () => {
+    expect(
+      parseOffset(
+        '10px + 5%, 10px - 5%',
+        referenceOffsets,
+        popperOffsets,
+        'bottom'
+      )
+    ).toEqual([10 + 5, 10 - 5]);
```
shmax: You lost me on this one; `10 + 5` is only the correct answer if `5%` of some value is `5`, right?

shmax: Ah, I see it is, okay

shmax: Also, I know you're deprecating it, but I don't see much in the way of tests for comma-less syntax...

FezVrasta: I don't wanted to add too much tests for a deprecated syntax, feel free to send an additional PR to add more tests if you prefer.

shmax: Also, what about weird cases like:

```
100%p - 10%r, 100% // mixing of p and r
100% + 50% + 10%, 100% // stringing more than one add
100%+50px, 100px-10+10% // tight syntax
```

FezVrasta: Yes that's something we should add, thank you for the comment.

FezVrasta: The tests you suggested caught a new bug 🙃

Now I think we are well covered.

shmax:
Yee-haw, nice! LGTM

mqklin:
@FezVrasta how can I offset relative to cross-axis?

mqklin:
@FezVrasta I mean this: 
`For length is intended the main axis relative to the placement of the popper.
This means that if the placement is top or bottom, the length will be the width. In case of left or right, it will be the height.`
How can I use `cross axis`, not `main axis` for this purpose?

FezVrasta:
I may be wrong but I don't think this is implemented 🤔

mqklin:
Sad! Anyway, I found a workaround - just use get `getBoundingClientRect` from `reference` element and put `top: ${rect.height / 2}px;` to tooltip's css.

FezVrasta:
I think we could directly get rid of all this stuff and make `offset` accept a function that provides the `rect` of `popper` and `reference` as arguments... That would be easier and more powerful.

```js
{
  offset: (reference, popper) => [reference.width / 2, popper.height * reference.height]
}
```

mqklin:
Brilliant idea!

PR #232, opened 2017-04-24
[fix] Make `auto` placement use `flip.boundariesElement`

### LINKED ISSUE #231: Unexpected position of popper (placement: 'auto').
If free space above or under reference element is not enough, it is expected that popper will be placed to the right or left of the element, but this does not happen.

Example: https://codepen.io/imkremen/pen/YVGxeB

FezVrasta:
fixes #231

@imkremen may you check this out? It's enough to use this URL:
https://rawgit.com/FezVrasta/8c9b742ced8e78e45fbbc1166f366430/raw/60902150babc6dcca2541e571a4c8b48ac219d63/popper-1.8.4.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "popper.js",
-  "version": "1.8.3",
+  "version": "1.8.4",
   "description": "A kickass library to manage your poppers",
   "main": "dist/umd/popper.js",
   "module": "dist/esm/popper.js",
--- a/src/popper/index.js
+++ b/src/popper/index.js
@@ -223,11 +223,12 @@ export default class Popper {
       this.options.placement,
       data.offsets.reference,
       this.popper,
-      this.reference
+      this.reference,
+      this.options.modifiers.flip.boundariesElement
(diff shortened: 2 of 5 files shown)


imkremen:
@FezVrasta As I see now, it's not only about "auto" placement and x-axis, it's about any value of placement and wrong flipping.

Big width and "auto": https://codepen.io/imkremen/pen/wdzXgg
Top: https://codepen.io/imkremen/pen/JNRvge
Bottom: https://codepen.io/imkremen/pen/OmREMN
Left: https://codepen.io/imkremen/pen/ybaEJp
Right: https://codepen.io/imkremen/pen/YVGvGK

FezVrasta:
I can see a bug only in the first of the pens, the others are behaving correctly.

The `flip` modifier by default will flip the placement only once, it means that if it flips it from `top` to `bottom`, but there's not enough space on `bottom`, it will not try to flip it to `top` again.

May you create a Pull Request that introduces a failing test for the first example please?

Issue #224, opened 2017-04-21
Position issue when scroller is inside another scroller


Cryde:
### CodePen demo

https://codepen.io/FezVrasta/pen/vmKWXp

### Steps to reproduce the problem

1.  simply hover the button => Tooltip appear correctly
2.  scroll on the outer scroller (body one)
3.  Tooltip isn't properly displayed (not at the right position)

### What is the expected behavior?

The tooltip should be displayed at the right position (the same when the page is render, when no scroll is made)

### What went wrong?

The position calculation is made based on the body scroller and also the tooltip is inserted into the body instead of the inner scroller.
So when I scroll outside, the position become wrong.

FezVrasta:
Can you please convert the codepen to use only Popper.js and not any 3rd party library?

Cryde:
Here is the example : https://codepen.io/anon/pen/dWXZXR

FezVrasta:
The interested file is [`getOffsetRectRelativeToArbitraryNode`](https://github.com/FezVrasta/popper.js/blob/master/src/popper/utils/getOffsetRectRelativeToArbitraryNode.js). We are addressing the scroll offsets of only the first scrollable parent instead of including all the scrollable parents contained in the common offset parent.

marvinhuebner:
Is there a possible fix available in the near future? 

I'm using https://github.com/atomiks/tippyjs in a Project and it is really awesome and popper.js does also a good Job. But we have some trouble with Elements which are inside a scrollable Container.

Really looking forward.

FezVrasta:
You could send a PR to introduce a failing test case in our functional tests to help me fix the bug

vladferix:
+1

FezVrasta:
@vladferix unfortunately `+1` will not make a CI test appear magically 😞

FezVrasta:
Adding a different codepen from #245 for completeness.

https://codepen.io/FezVrasta/pen/XRpQXJ

FezVrasta:
Added CI test dfb6129160e28cbb4515af7e51be6353c59e7784

aristidesfl:
Is there any known version of popper.js not affected by this bug?

FezVrasta:
I don't think so

aristidesfl:
Why doesn't scrolling example in the landing page exhibit the problem? https://popper.js.org/

FezVrasta:
Probably because I manually define the `boundariesElement`

aristidesfl:
[...]

PR #235, opened 2017-04-25
Auto-placement fix

imkremen:


FezVrasta:
Thank you for the test but I need automatic/CI tests, they live in `functional` folder.

May you move it there? You can look at the existing ones (in `core.js` mainly) to see how to write one.

imkremen:
Ok, I'll try.

scripts/bundle.js
```diff
@@ -6,7 +6,7 @@ const { build } = argv;
 
 const rollup = 'rollup -c ./scripts/rollup.config.js';
 const minify = 'node ./scripts/minify.js';
-const gzipped = '$(npm bin)/gzipped';
+const gzipped = 'gzipped';
```
FezVrasta: I think this is going to work only if `gzipped` is globally installed.

imkremen: No, it works good

src/popper/utils/computeAutoPlacement.js
```diff
@@ -14,24 +14,61 @@ export default function computeAutoPlacement(
...
   };
+  
+  const sortedRects = [];
+  for (const rect in rects) {
+    sortedRects.push([rect, rects[rect]]);
+  }
+
+  sortedRects.sort(function(a, b) {
+    return b[1].getArea() - a[1].getArea();
+  });
```
FezVrasta: Cool! This could be written as:

```
const sortedRects = Object.keys(rects)
  .map(rect => [rect, rects[rect])
  .sort(([a], [b]) => b.getArea() - a.getArea());
```

src/popper/utils/computeAutoPlacement.js
```diff
@@ -14,24 +14,61 @@ export default function computeAutoPlacement(
...
+    },
+    bottom: {
+      width: boundaries.width,
+      height: boundaries.bottom - refRect.bottom,
+      getArea: function() { return this.width * this.height; },
+    },
+    left: {
+      width: refRect.left - boundaries.left,
+      height: boundaries.height,
+      getArea: function() { return this.width * this.height; },
```
FezVrasta: This seems redundant, you could define it as an external function:

```
function getArea({ width, height }) { return width * height; }
```

And use it below, instead of define it on each element of the list.

src/popper/utils/computeAutoPlacement.js
```diff
@@ -14,24 +14,61 @@ export default function computeAutoPlacement(
...
+
+  sortedRects.sort(function(a, b) {
+    return b[1].getArea() - a[1].getArea();
+  });
+
+  const resultRects = sortedRects.filter(function(rect) {
+    if (rect[1].width >= popper.clientWidth && rect[1].height >= popper.clientHeight) {
+      return rect;
+    }
+  });
```
FezVrasta: Maybe written as:

```
const resultRects = sortedRects.filter(([{ width, height }]) =>
  return width >= popper.clientWidth && height >= popper.clientHeight;
);
```

aristidesfl:
Any estimate for this? Asking just in to decide on something.

FezVrasta:
There isn't any reply from the OP.

If you want you can fork his branch and continue the work.

aristidesfl:
```
We still need a functional test to avoid regressions.
```
You are missing this?

FezVrasta:
Yes, and a my other comments need to be addressed.

aristidesfl:
Would it be viable for you to address the changes on your comments?

FezVrasta:
Yes I can do it.

aristidesfl:
Updated the original codepen with a version of popper built with this branch, but the problem still exists. 
https://codepen.io/aristidesfl/pen/gWvXmG

FezVrasta:
This PR doesn't aim to fix #224

PR #236, opened 2017-04-25
[refactor] use iteration over recursion

kybishop:
Iteration is slightly faster since javascript doesn't have tail-call optimization until ES2015.
 This should still [provide a small speed boost](https://softwareengineering.stackexchange.com/questions/179863/performance-recursion-vs-iteration-in-javascript#comment746263_231673) until pre ES2015 modules are no longer required :)

### DIFF
--- a/src/popper/utils/getRoot.js
+++ b/src/popper/utils/getRoot.js
@@ -6,8 +6,8 @@
  * @returns {Element} root node
  */
 export default function getRoot(node) {
-  if (node.parentNode !== null) {
-    return getRoot(node.parentNode);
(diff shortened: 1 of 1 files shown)


FezVrasta:
Did you need this because you noticed slowness?

kybishop:
No slowness, just saw it when I was going through the codebase. Feel free to close if you want to avoid the churn

FezVrasta:
I use function recursion in the whole code base because I prefer a functional approach, sorry but I think I'll keep it as is.

If in the future we see any performance bottlenecks because of this I'll be happy to reconsider this.

PR #237, opened 2017-04-25
support poppers that are children of their reference

### LINKED ISSUE #233: Add support for poppers child of their (positioned) reference element
### CodePen demo

http://codepen.io/kybishop/pen/zwKmyX

1.  Add `position: relative` to any of the popper's parent elements
2. Note how the popper's offsets are greater than expected

The popper adjusts its position relative to the relatively positioned parent.

### What went wrong?

The popper adjusts itself as if it were positioned on the document body

kybishop:
Fixes #233 

If the test coverage isn't quite there, happy to add additional tests if pointed in the right direction. I'm not quite sure where the tests for this change are/should be

### DIFF
--- a/src/popper/utils/findCommonOffsetParent.js
+++ b/src/popper/utils/findCommonOffsetParent.js
@@ -31,7 +31,8 @@ export default function findCommonOffsetParent(element1, element2) {
 
   // Both nodes are inside #document
   if (
-    element1 !== commonAncestorContainer && element2 !== commonAncestorContainer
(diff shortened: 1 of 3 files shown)


FezVrasta:
Thank you. I think we need to add a test to cover this case to prevent regressions. You can use as reference the last tests of the core.js file

kybishop:
@FezVrasta Added two tests which fail without the fix and squashed commits. Happy to add other test cases if you think we need them!

PR #240, opened 2017-04-26
fix: lazily check for IE10

### LINKED ISSUE #238: 1.8.4: Regression: ReferenceError: navigator is not defined
In 1.8.4 the IE10 detection was changed from lazy (within the `getBoundingClientRect` method) to eager (in the global scope):

https://github.com/FezVrasta/popper.js/commit/d185897814c2b44e55a53d59d5cd9653018fe4fe#diff-948f8780a0b0051f85e85abd0b5337c4R7

The problem with that is, that parts of our app runs in a node context - and due to popper doing the check eagerly (see https://github.com/FezVrasta/popper.js/commit/d185897814c2b44e55a53d59d5cd9653018fe4fe#diff-948f8780a0b0051f85e85abd0b5337c4R7) this now fails in the node context with:

```
/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:31292
  return navigator.appVersion.indexOf('MSIE 10') !== -1;
         ^
 
ReferenceError: navigator is not defined
    at runIsIE10 (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:31292:10)
    at Object../node_modules/popper.js/dist/esm/popper.js (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:31295:16)
    at __webpack_require__ (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:20:30)
    at Object../src/ui/pin/pin.tsx (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:68362:14)
    at __webpack_require__ (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:20:30)
    at Object../src/ui/flyout/flyout.tsx (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:65476:13)
    at __webpack_require__ (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:20:30)
    at Object.<anonymous> (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:65533:16)
    at Object../src/ui/flyout/stories/flyout.stories.tsx (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:65546:30)
    at __webpack_require__ (/var/lib/buildkite-agent/builds/canva-org/canva-web-build/web/target/build/stories/stories.js:20:30)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

can we move the IE10 detection back into the method where it is actually used, so it will only be executed if a `Popper` instance is actually constructed? Opened #240.

joscha:
fixes #238

### DIFF
--- a/src/popper/utils/getBoundingClientRect.js
+++ b/src/popper/utils/getBoundingClientRect.js
@@ -3,8 +3,7 @@ import getBordersSize from './getBordersSize';
 import getWindowSizes from './getWindowSizes';
 import getScroll from './getScroll';
 import getClientRect from './getClientRect';
-import runIsIE10 from './isIE10';
-const isIE10 = runIsIE10();
+import isIE10 from './isIE10';
 
 /**
  * Get bounding client rect of given element
@@ -19,7 +18,7 @@ export default function getBoundingClientRect(element) {
   // IE10 10 FIX: Please, don't ask, the element isn't
   // considered in DOM in some circumstances...
   // This isn't reproducible in IE10 compatibility mode of IE11
-  if (isIE10) {
+  if (isIE10()) {
     try {
       rect = element.getBoundingClientRect();
       const scrollTop = getScroll(element, 'top');
--- a/src/popper/utils/getOffsetRectRelativeToArbitraryNode.js
+++ b/src/popper/utils/getOffsetRectRelativeToArbitraryNode.js
@@ -4,9 +4,9 @@ import getScrollParent from './getScrollParent';
 import getBoundingClientRect from './getBoundingClientRect';
 import runIsIE10 from './isIE10';
 import getClientRect from './getClientRect';
-const isIE10 = runIsIE10();
 
 export default function getOffsetRectRelativeToArbitraryNode(children, parent) {
+  const isIE10 = runIsIE10();
   const isHTML = parent.nodeName === 'HTML';
   const childrenRect = getBoundingClientRect(children);
   const parentRect = getBoundingClientRect(parent);
--- a/src/popper/utils/isIE10.js
+++ b/src/popper/utils/isIE10.js
@@ -4,6 +4,11 @@
  * @memberof Popper.Utils
  * @returns {Boolean} isIE10
  */
+let isIE10 = undefined;
+
 export default function() {
-  return navigator.appVersion.indexOf('MSIE 10') !== -1;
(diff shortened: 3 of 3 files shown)


FezVrasta:
I moved these check outside the functions to avoid to call them on each call (if we are on IE, we are on IE, it will never change).

Could you change the checks to not fail on Node.js instead?

joscha:
> Could you change the checks to not fail on Node.js instead?

Generally I think avoiding side-effects by imports should be the way to go, but sure, I can guard that call.

FezVrasta:
I understand but at the moment I prefer to improve performance rather than have a super clear code for Node.js.

joscha:
> I understand but at the moment I prefer to improve performance rather than have a super clear code for Node.js.

sure, change is in f2cea9d

FezVrasta:
Would you like to add a test for Node in a separated PR maybe?

joscha:
> Would you like to add a test for Node in a separated PR maybe?

can do, which test runner would you like me to use? mocha? Because currently everything seems to be run in karma, which has all those globals.

FezVrasta:
Maybe Jest? So we move all our unit tests there add well.

joscha:
Jest has JSDOM by default, which fills all the attributes that break in a pure node env I think?

FezVrasta:
Nope yup have to enable it manually

joscha:
> Nope yup have to enable it manually

👍 jest it is then :) Do you want this as a prereq for this PR being merged or can we do that independently?

FezVrasta:
It's okay in a new pr

FezVrasta:
I was thinking that alternatively we could cache the value the first time the function is ran and reuse it at subsequent calls. This would probably be clear? What do you think?

joscha:
thanks @FezVrasta - can we get this and #238 in a patch release?

FezVrasta:
1.8.5

PR #242, opened 2017-04-26
fix: expectation of `window` being defined at import time

### LINKED ISSUE #241: 1.8.4: Regression: window.PopperUtils
https://github.com/FezVrasta/popper.js/commit/f6c7e757e13727d260814ddede8baec2ccc5a7ad#diff-94e560edeef11a82815d4eeaff44e53cR344

broke any simple tool that involves loading popper, due to:

```
Popper.Utils = window.PopperUtils;
                     ^

TypeError: Cannot read property 'PopperUtils' of undefined
    at Object../node_modules/popper.js/dist/esm/popper.js (/Users/joscha/dev/canva/web/target/build/stories/stories.js:32963:22)
    at __webpack_require__ (/Users/joscha/dev/canva/web/target/build/stories/stories.js:20:30)
    at Object../src/ui/pin/pin.tsx (/Users/joscha/dev/canva/web/target/build/stories/stories.js:68216:14)
    at __webpack_require__ (/Users/joscha/dev/canva/web/target/build/stories/stories.js:20:30)
    at Object../src/ui/flyout/flyout.tsx (/Users/joscha/dev/canva/web/target/build/stories/stories.js:65476:13)
    at __webpack_require__ (/Users/joscha/dev/canva/web/target/build/stories/stories.js:20:30)
    at Object.<anonymous> (/Users/joscha/dev/canva/web/target/build/stories/stories.js:65533:16)
    at Object../src/ui/flyout/stories/flyout.stories.tsx (/Users/joscha/dev/canva/web/target/build/stories/stories.js:65546:30)
    at __webpack_require__ (/Users/joscha/dev/canva/web/target/build/stories/stories.js:20:30)
    at webpackContext (/Users/joscha/dev/canva/web/target/build/stories/stories.js:64267:9)
```

whereas before the import was resolved by node, now `window` is needed at import time.

This however only works for environments where `window` is available, which is not true for a lot of test runners and other node tools.

joscha:
This removes the expectation of popper that `window` is defined at import time (which is incorrect for any non-browser environment).

fixes #241

### DIFF
--- a/src/popper/index.js
+++ b/src/popper/index.js
@@ -190,11 +190,11 @@ export default class Popper {
   //
 
   /**
-     * Updates the position of the popper, computing the new offsets and applying the new style
-     * Prefer `scheduleUpdate` over `update` because of performance reasons
-     * @method
-     * @memberof Popper
-     */
+   * Updates the position of the popper, computing the new offsets and applying the new style
+   * Prefer `scheduleUpdate` over `update` because of performance reasons
+   * @method
+   * @memberof Popper
+   */
   update() {
     // if popper is destroyed, don't perform any further update
     if (this.state.isDestroyed) {
@@ -252,17 +252,17 @@ export default class Popper {
   }
 
   /**
-     * Schedule an update, it will run on the next UI update available
-     * @method scheduleUpdate
-     * @memberof Popper
-     */
+   * Schedule an update, it will run on the next UI update available
+   * @method scheduleUpdate
+   * @memberof Popper
+   */
   scheduleUpdate = () => requestAnimationFrame(this.update);
 
   /**
-     * Destroy the popper
-     * @method
-     * @memberof Popper
-     */
+   * Destroy the popper
+   * @method
+   * @memberof Popper
+   */
   destroy() {
     this.state.isDestroyed = true;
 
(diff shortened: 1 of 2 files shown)


joscha:
> Would not be better to fallback to global to make it work on node?

I suppose it would definitely prevent the `... is undefined` error. The best way of dealing with this is by having the user manually inject the `PopperUtils` symbol to wherever it is needed. If we can't do that, at least it shouldn't expect `window` to befined. Happy to change the PR - how do you want me to fill `global` for the browser? Does popper do that anywhere else?

FezVrasta:
I was just thinking of 

window ? window.PopperUtils = ... : global.PopperUtils = ...

joscha:
changed it in 101063c

joscha:
> window ? window.PopperUtils = ... : global.PopperUtils = ...

ah, just saw this after the change I made, I can update.

FezVrasta:
Nope it's good, i just woke up lol

joscha:
cool! Let me know if there is anything else I can do - there is also a related ticket #238 with a regression which was introduced in the same version

FezVrasta:
I was thinking that alternatively we could cache the value the first time the function is ran and reuse it at subsequent calls. This would probably be clear? What do you think?

src/popper/index.js
```diff
@@ -322,7 +322,7 @@ export default class Popper {
      * include `popper-utils.js` before `popper.js`.
      * @memberof Popper
      */
-  static Utils = window.PopperUtils;
+  static Utils = (window || global).PopperUtils;
```
FezVrasta: Umh, I tried this in the node console and seems like it doesn't work?

If we do `(this.window === this ? window || global)` it doesn't throw errors instead.

FezVrasta:
I hit a blocker with https://github.com/rollup/rollup/issues/1385, for now I'm checking for `window` being defined.

In the next major of Popper, the Popper.Utils object will be removed completely so I think it's okay to keep it as is (if nobody answers to my issue and provides a solution).

PR #255, opened 2017-05-10
Fix broken table caused by HTML example

thomas479:
The broken HTML example in the documentation of tooltip.js caused the table to get three extra columns.


FezVrasta:
Thank you for the PR, these files are auto generated by JSdoc, it means we can't directly edit them, but we have to fix it from the comments of the code that are used by JSdoc to create the markup.

FezVrasta:
Just an update to let you know the problem is now fixed

PR #258, opened 2017-05-11
Fix `dispose` not removing events

Akryum:
The events were removed from the tooltip element itself instead of the target `reference` element.

### DIFF
--- a/src/tooltip/index.js
+++ b/src/tooltip/index.js
@@ -241,7 +241,7 @@ export default class Tooltip {
 
       // remove event listeners
       this._events.forEach(({ func, event }) => {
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thank you, do you have time to also fix the unit test we already have? If it didn't catch this problem it means it's broken.

Akryum:
Not anytime soon, sorry :smile:

FezVrasta:
Ok I'll try to allocate some time for it. Thank you again.

Until the unit test is broken I'm holding on this PR. I'll merge it ASAP.

FezVrasta:
moved to #269

Issue #256, opened 2017-05-11
Methods not listed in prototype


joscha:
### Steps to reproduce the problem

1.  Call `.destroy` on a popper instance

### What is the expected behavior?

Destroys the popper instance

### What went wrong?

I am getting:

```
TypeError: myPopper.destroy is not a function
```

### Any other comments?

This works in `1.9.1`.

joscha:
seems to happen due to sinon not being able to recognize the existence of`.destroy` any longer:

```
const p = sinon.createStubInstance(Popper);
p.destroy(); // works in 1.9.1
```

joscha:
Workaround:

```
const popper = sinon.createStubInstance(Popper);
// TODO: remove workaround for: https://github.com/FezVrasta/popper.js/issues/256
popper.destroy = sinon.spy();
```

FezVrasta:
I didn't change anything about that method in 1.9.2, can you provide a codepen or a github repo to reproduce the problem?

joscha:
```
const p = sinon.createStubInstance(Popper);
p.destroy(); // works in 1.9.1
```
is all you need to do - I suppose the refactor in https://github.com/FezVrasta/popper.js/commit/3523662701df36b060272e842eda389e97eb1ddc somehow causes the `destroy` method to not appear on the proto any more, hence sinon won't be able to stub it.

FezVrasta:
It seems to work just fine here:
https://codepen.io/FezVrasta/pen/xdWqjL

I'm not sure how Sinon `createStubInstance` work... Inside the code, I simply define `this.destroy = destroy` in the `constructor`.

Previously it was defined in the list of methods. The code change was:

```diff
-destroy() {
-  // stuff
-}
+destroy = destroy // destroy is a function with stuff inside
```

I wonder if it could be enough to do:

```
destroy() {
  return destroy();
}
```

joscha:
\o/ - let me know when there is a new version @FezVrasta

FezVrasta:
1.9.3

PR #259, opened 2017-05-11
[fix] #256

FezVrasta:
Please @joscha verify that this fixes your problem with Sinon.


joscha:
@FezVrasta yes, this fixes #256

PR #261, opened 2017-05-12
Added cached math methods to prevent namespace pollution

flibX0r:
Was annoyed that **babel-plugin-minify-builtins** was adding _Mathxxx functions to the global namespace.

Gives an extremely minor performance increase, cleaner minified output


FezVrasta:
I think a proper fix would be to keep the matxxx stuff inside the umd bundle...

flibX0r:
Ideally yes, but **babel-plugin-minify-builtins** is written very simplistically and dumps those globals instead of determining the highest common scope

FezVrasta:
I opened https://github.com/babel/babili/issues/534

PR #266, opened 2017-05-12
[fix] Don't attempt to use scrollSize to detect window sizes

### LINKED ISSUE #251: Popper does not update its position to fit entirely in viewport/document if it is overflowing
### CodePen demo

https://codepen.io/FezVrasta/pen/QvQQgv

1. Simply resize the browser window and you'll notice part of the Popper begins to overflow the document edge at the boundaries.

That it is fully visible in the viewport with no overflow on a single update (i.e., it fits within the content of the document excluding itself).

### What went wrong?

It overflows.

FezVrasta:
Please @atomiks may you test this version and confirm it fixes #251?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "popper.js",
-  "version": "1.9.3",
+  "version": "1.9.4",
   "description": "A kickass library to manage your poppers",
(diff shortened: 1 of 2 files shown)


atomiks:
Yes, that fixes it!

PR #279, opened 2017-05-19
[fix] Fix rounding of arrow offset

FezVrasta:
should fix #227

ping @atomiks

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "popper.js",
-  "version": "1.9.4",
+  "version": "1.9.5",
   "description": "A kickass library to manage your poppers",
   "main": "dist/umd/popper.js",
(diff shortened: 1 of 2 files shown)


atomiks:
Should be `floor()` not `round()`?

FezVrasta:
What if is `x.9`? Isn't it better to round?

atomiks:
The problem is that x.5 rounds up to (x+1).0 which is more incorrectly placed than x.0. Also, it is never anything other than x.0 and x.5 if it's divided by 2? I just resized the window so the arrow moved away from the center and it was always .0 or .5.

atomiks:
So going with `floor()`? `round()` wouldn't fix the issue if it's a half pixel and rounds up rather than down (as I mentioned in the issue)

PR #280, opened 2017-05-19
Fix arrow not moving in tooltips from Tooltip.js

aristidesfl:


FezVrasta:
I took a look at the code, I define the arrow selector here:
https://github.com/FezVrasta/popper.js/blob/master/src/tooltip/index.js#L206

it should override the `x-arrow` selector. Are you able to investigate to understand why this setting isn't used?

aristidesfl:
I see, not at the moment unfortunately. Why did you decide to change the selector instead of adding `x-arrow` and maintain the standard which popper.js uses?

FezVrasta:
Mostly to keep the template simpler.

aristidesfl:
So the fact that `x-arrow` is working means that 2 selectors are being used? Or was the custom selector supposed to override the `x-arrow` selector?

Independently of that, I'm inclined to think that the increase in consistency between tooltip.js and popper are enough to justify adding `x-arrow` to the template in this case.

FezVrasta:
Tooltip.js was created to show to other developers how to properly integrate Popper.js in their own libraries, so I wanted to use the API to make the code as clean as possible.

Use a custom selector instead of the default `x-arrow` is one of the API I want to showcase.

We may fix the problem with this PR, but if this thing isn't working properly, we can't simply stop using it and fake nothing is broken.

The proper solution is to find out what's wrong with the API.

aristidesfl:
Question unanswered: 
> So the fact that x-arrow is working means that 2 selectors are being used? Or was the custom selector supposed to override the x-arrow selector?

FezVrasta:
The custom selector should overerride the default one (`x-arrow`).

PR #283, opened 2017-05-22
Fix 224

vladferix:
Looks like it fixes the following issue (#224)

### DIFF
--- a/src/popper/utils/getOffsetRectRelativeToArbitraryNode.js
+++ b/src/popper/utils/getOffsetRectRelativeToArbitraryNode.js
@@ -44,7 +44,7 @@ export default function getOffsetRectRelativeToArbitraryNode(children, parent) {
   }
 
   if (
-    parent.contains(scrollParent) &&
+    parent === scrollParent &&
     (isIE10 || scrollParent.nodeName !== 'BODY')
   ) {
     offsets = includeScroll(offsets, parent);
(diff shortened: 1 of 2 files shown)


vladferix:
Sorry, for some strange reason tests were passed locally, will check what is missing

FezVrasta:
Looks like there's something wrong with IE10

vladferix:
@FezVrasta yes, and method includes() instead of direct comparison was originally used to fix some related ie10 issues.

FezVrasta:
Are you going to work on this?

vladferix:
Yes, but I am not sure that I can fix this. I am trying to approach to fix, but still don't fully understand all cases and logic behind this

FezVrasta:
Oh really? Was is just that? Fucking method names...

Thank you so much!

vladferix:
Still don't fully understand why, but using strict comparison (which I believe is more correct) if parent is scrollable with all browsers, and keeping contains() with IE10, makes all tests to pass :)

FezVrasta:
The use of `contains` is needed to take into account multiple scrollable elements between the `commonOffsetParent` and the reference element.

I'm going to review these changes to see if we can get rid of some of that logic.

vladferix:
No issues with method names. Using includes() instead of contains() was just my mistake (always messing these two methods). The issue was detection whether we need to include scroll offset of main parent or not, and very strange logic for this detection with IE10, which I still not fully understand.

vladferix:
I assume that all multiple scrolls are calculated implicitly in the 

`let offsets = getClientRect({
    top: childrenRect.top - parentRect.top,
    left: childrenRect.left - parentRect.left,
    width: childrenRect.width,
    height: childrenRect.height,
  });`

but the issue is when parent itself is scrollable, not its childs...

But anyway, I don't know if I am right or wrong, waiting for your review.

kybishop:
I can confirm that this fixes popper position for me when using them inside `overflow: scroll` elements! 🎊

PR #289, opened 2017-05-24
[fix] Fixed #288

### LINKED ISSUE #288: Pop auto placement when pop is created dynamicaly
Hi @FezVrasta ,

Currently I try to implement `auto` position in Bootstrap
but it throw an error.

I created a CodePen see : http://codepen.io/Johann-S/pen/ZKwRvo

1.  Just look the console logs

Pop element should be auto placed

FezVrasta:
Fixes #288

cc @Johann-S for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "popper.js",
-  "version": "1.9.8",
+  "version": "1.9.9",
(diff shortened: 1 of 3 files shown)


Johann-S:
I can't check that now, but later
Thank you for your reactivity 👍

PR #291, opened 2017-05-26
Fix tooltip trigger events

### LINKED ISSUE #290: Tooltip.js focus trigger only flashes tooltip
### CodePen demo

https://codepen.io/anon/pen/wdZxaz

1.  Setup a Tooltip.js tooltip on a text input field with a `focus` trigger
2.  Click inside the input to focus

The tooltip displays as long as focus is in the field.

### What went wrong?
[](url)
The tooltip only displays very quickly then disappears. This happens in latest Chrome and Firefox. It does not happen if the tooltip is triggered by tabbing in the field.

alexturpin:
Fixes #290. Trying to figure out how to write a proper test for this.

### DIFF
--- a/src/tooltip/index.js
+++ b/src/tooltip/index.js
@@ -286,12 +286,15 @@ export default class Tooltip {
         case 'hover':
           directEvents.push('mouseenter');
           oppositeEvents.push('mouseleave');
+          break;
         case 'focus':
           directEvents.push('focus');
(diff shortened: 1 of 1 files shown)


alexturpin:
I'm afraid I couldn't figure out how to add a proper test for this in the time that I had to give to this. I'd like to welcome anyone else who has an idea to contribute one.

PR #294, opened 2017-05-29
Add NuGet spec to publish Popper.js on NuGet

Johann-S:
Add `popper.js.nuspec` to register Popper.js on NuGet.

@FezVrasta when you want to publish your new release of Popper.js you have to do : 
`nuget pack popper.js.nuspec` it will create a file : `popper.js.<release version>.nupkg`

And to finish the publishing see : https://docs.microsoft.com/en-us/nuget/create-packages/publish-a-package

Note : 
 - you have to create a `dist` folder with all your dist files (`dist/umd`, `dist/esm`) because its allow NuGet to share your dist files
 - To get the latest release of NuGet : https://dist.nuget.org/win-x86-commandline/latest/nuget.exe

PR #296, opened 2017-05-30
[meta] Added nugetPublish script

FezVrasta:
This PR supersedes #294

The `postpublish` step will generate a `package.nuspec` file, I'm copying below the result:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
    <metadata>
        <id>popper.js</id>
        <version>1.9.9</version>
        <description>A kickass library to manage your poppers</description>
        <authors>FezVrasta</authors>
        <projectUrl>https://popper.js.org</projectUrl>
    </metadata>
    <files>
        <file src="*" target="content\Scripts" exclude="**/dist/**/tooltip*;**/dist/**/tooltip*/**;**/docs;**/docs/**;**/scripts;**/scripts/**;**/src;**/src/**;**/tests;**/tests/**;**/popperjs.png;**/popperjs.png/**"/>
    </files>
</package>
```

The problem is that the script fails with:

```
Error: Nuget: cannot build /Users/FezVrasta/Projects/popper.js/package.nuspec. Missing files: /Users/FezVrasta/Projects/popper.js/*
```

Ideas?

cc @Johann-S


Johann-S:
In this PR #297 (which target your current branch `nuget`) I made a quick fix

FezVrasta:
Thank you, with your fix it works, but I actually wanted to include all the files that are shipped with npm to avoid any kind of difference between the two packages 🤔

Is it a good idea or does Nuget work in a completely different way and doesn't make any sense my approach?

Johann-S:
No you can the same they are no problem 👍 
I didn't know which files do you want so I included all `dist` files

Johann-S:
Can you do a list of folders and files you want in your package please  ?
And I'll update the script to include all the files/folders you want 👍

FezVrasta:
I just want to read `.npmignore` and include in the nuget package everything not excluded by it.

I'm trying with this code but without any luck for now:

```js
const fs = require('fs');
const path = require('path');
const xml = require('xml');
const ignoreToGlob = require('gitignore-to-glob');
const glob = require('glob');
const nuget = require('nuget');
const pkg = require('../package.json');

const nuspecPath = path.resolve(__dirname, '../package.nuspec');

// Convert npmignore to list of glob patterns
const npmignore = ignoreToGlob(path.resolve(__dirname, '../.npmignore'))
  .map(
    pattern => (pattern.indexOf('!') === 0 ? pattern.slice(1) : `!${pattern}`)
  )
  .map(pattern => (pattern.endsWith('**') ? `${pattern}/*` : pattern));
npmignore.push('**/node_modules/**/*');

glob(`!(${npmignore.join('|')})`, (err, files) => {
  console.log(`!(${npmignore.join('|')})`, files);
});
```

FezVrasta:
So, fixed it with:

```js
// Convert npmignore to list of glob patterns
const npmignore = ['!**/node_modules/**']
  .concat(ignoreToGlob(path.resolve(__dirname, '../.npmignore')))
  .map(
    pattern => (pattern.indexOf('!') === 0 ? pattern.slice(1) : `!${pattern}`)
  )
  .map(pattern => (pattern.endsWith('**') ? `${pattern}/*` : pattern));

glob(`**/*`, {
  ignore: npmignore,
}, (err, files) => {
  console.log(`!(${npmignore.join('|')})`, files);
});

```

Johann-S:
But `glob` is async so you should move the rest of code in glob's callback

Johann-S:
Sorry @FezVrasta but your script didn't work because you wrote in your `package.nuspec` folder paths for example : `<file src="dist"/>`

FezVrasta:
Yup I'm on it

FezVrasta:
```
$ node ./scripts/nugetPublish.js
Attempting to build package from 'package.nuspec'.
Successfully created package '/Users/FezVrasta/Projects/popper.js/popper.js.1.9.9.nupkg'.
Object reference not set to an instance of an object
Error: Failed to pack file: /Users/FezVrasta/Projects/popper.js/package.nuspec
```

🤔

FezVrasta:
I wonder if it could be a problem with macOS...

Johann-S:
So strange when I ran manually : `npm pack package.nuspec` `nupkg` is created without error
maybe it's on the publish

FezVrasta:
Yup I noticed this as well, but it fails on the pack step. Maybe the npm package I'm using to run nuget isn't good enough..

FezVrasta:
I forked the `nuget` package and fixed it (published as `nuget-in-path`). Now everything works properly.

PR #297, opened 2017-05-30
Fix nugetPublish script to package all dist files

Johann-S:
@FezVrasta a quick fix to your nugetPublish script which build nuget package as expected 👍


FezVrasta:
Cool! Thank you

PR #316, opened 2017-06-19
tooltip.js: added option "popperOptions" to pass directly for popper …

vgavro:
Any purpose not allowing to pass extended options to popper instance? I need some modifiers on my tooltips, and it appears to be pretty tricky to add modifiers on existing instance.
Thanks!


FezVrasta:
I don't quite get what are you doing here. Seems like the `popperOptions` object is being assigned inside the constructor, so you can't change it on an existing istance to change the options..?

vgavro:
Thanks for you quick reply!
Changing `tooltip.popperInstance.options.modifiers` *after* creating `popper` instance is just ignored. *Pushing* modifier directly to `tooltip.popperInstance.modifiers` *always* applies this modifier after setting style (`computeStyle`, `applyStyle`, `offset` modifiers, even with `order: 1`) - obviously because order means something only on Popper instance creation (on instance creation it works as expected). So I don't see any *reasonable* way to update modifiers after instance creation - in the same handy way, as you can pass custom modifiers to Popper constructor right now. So because of it I think it would be handy to pass popper constructor options to tooltip constructor. Have I missed something? popper v1.10.2 (same as fresh master). Thanks for your time! I'll fix IE10/11 build fails if you consider this PR as useful.

FezVrasta:
Ok sorry, now I see what you mean, it looks good!

src/tooltip/index.js
```diff
@@ -199,14 +201,15 @@ export default class Tooltip {
 
     this._append(tooltipNode, container);
 
-    const popperOptions = {
+    const popperOptions = Object.assign(options.popperOptions || {}, {
```
FezVrasta: I'd like to avoid to directly update the existing object, and use spread operator instead of Object.assign.

```
const popperOptions = {...options.popperOptions, placement: options.placement };
```

src/tooltip/index.js
```diff
@@ -199,14 +201,15 @@ export default class Tooltip {
...
       placement: options.placement,
-      modifiers: {
-        arrow: {
-          element: this.arrowSelector,
-        },
+    });
+
+    popperOptions.modifiers = Object.assign(popperOptions.modifiers || {}, {
```
FezVrasta: Same here, I'd like the spread operator.

```
popperOptions.modifiers = {
  ...popperOptions.modifiers,
    arrow: {
       element: this.arrowSelector,
     },	
};
```

src/tooltip/index.js
```diff
@@ -200,13 +202,16 @@ export default class Tooltip {
     this._append(tooltipNode, container);
 
     const popperOptions = {
+      ...options.popperOptions || {},
```
FezVrasta: According to the [Babel REPL](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=react%2Cstage-2&targets=&browsers=&builtIns=false&debug=false&experimental=false&loose=false&spec=false&code_lz=DYUwLgBARg3AUKSBDCBeCBvAdDqBfIA) there's no need to fallback to an empty object if the first is undefined.

vgavro: thanks, sorry!

PR #318, opened 2017-06-20
Move aria describedby to reference

### LINKED ISSUE #317: aria-describedby is placed on tooltip instead of reference element
### CodePen demo

Haven't created a codepen – as it is on your demo as well. See this screenshot
image

1.  Inspect tooltip after it is initialised
2.  Tooltip has both aria-describedby and an id
3.  Nothing gets added to the reference element (the button) 

I would expect that the button had the aria-describedby, and the tooltip the id – and thereby tie them together

### What went wrong?

I think that on line 194 in ```/src/tooltip/index.js``` should be changed from

``` js
// Add `aria-describedby` to our reference element for accessibility reasons
tooltipNode.setAttribute('aria-describedby', tooltipNode.id);
```

to

``` js
// Add `aria-describedby` to our reference element for accessibility reasons
reference.setAttribute('aria-describedby', tooltipNode.id);
```

apdrsn:
Fixes  #317.
Now tooltip and reference element will have a link.
Before tooltip had both aria-describedby and the id.
Now tooltip will have a id and tooltip a aria-describedby.


apdrsn:
I've tried to sign the CLA but nothing happens?

FezVrasta:
Have you tried to click on "recheck"?

apdrsn:
Yes, I have.

FezVrasta:
That's weird 🤔 Let's try to give it some time

apdrsn:
Wow, that was weird.
Tried several times. It seemed to work when I had my console open at the same time :)

Issue #277, opened 2017-05-19
Arrow is off by 0.5px (1px rendered) when popper is an odd number of pixels wide


atomiks:
### CodePen demo

https://codepen.io/anon/pen/eWPMRd

### Steps to reproduce the problem

1.  Closely look at the arrow
2.  Notice when inspecting the element it is 43.5px left

### What is the expected behavior?

The arrow should be 43px left not 43.5px

### Any other comments?

Demonstration:

![Image](https://i.gyazo.com/7f6d47855cb55bb595a8ed9c38cfa93b.png)

Can you simply apply Math.floor() to the positioning of the arrow to round it down?

I'm on a retina display so I think it renders things differently, is this reproducible on a low DPI screen?

FezVrasta:
I don't have any low DPI screen to reproduce, I'll check it out ASAP to understand what's happening.

atomiks:
The codepen automatically used the latest unpkg version 1.9.6 and it's been fixed. Works fine for both even and odd number of pixels now.

atomiks:
Sadly there's still an issue... if it's vertical it's not correct.

However it's an easy fix:

```javascript 
data.offsets.arrow[side] = isVertical ? Math.round(sideValue) : Math.floor(sideValue);
```

FezVrasta:
Why `round` in one case and `floor` in the other? Wouldn't make sense to use `ceil` and `floor`?

atomiks:
I suppose... it's usually `x.5` so it rounds up to the next integer anyway?

Are there tests for this so you can calculate the midpoint of the reference and test if the arrow is properly positioned? That way it doesn't have to be guessing by eye.

FezVrasta:
I don't have such kind of tests ATM, if you are interested in this I'd love a PR to introduce them.

atomiks:
It seems it's not consistent. Sometimes rounding down is better, other times rounding up is better (in terms of properly centering it). I'm not sure what you can do. Perhaps just leave it as `round` for now... not sure

FezVrasta:
I forked your pen to better show the problem and hack on it.

https://codepen.io/FezVrasta/pen/mmojyP

atomiks:
The latest version still uses `floor()` right? I'm changing values and it seems to be perfectly centered each time. Maybe the CSS triangle's point is the problem?

FezVrasta:
I notice that if you zoom it gets decentered sometimes.

FezVrasta:
[...]

PR #319, opened 2017-06-24
[fix] #277

FezVrasta:
cc @atomiks, please make sure it solves your problem once for all.


atomiks:
I hope so, can't tell for certain but at least vertically the arrow seems better positioned.

I'm not sure if it renders perfectly on all browsers, have you tried more tests to see?

PR #353, opened 2017-07-13
Options.title should override title from attribute

websirnik:
Right now, if element has `title` attribute like `<a href=".." title="google.com">Link</a>`, `title="google.com"` will override the title set through options, but it should not.

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -182,7 +182,7 @@ export default class Tooltip {
     }
 
     // get title
-    const title = reference.getAttribute('title') || options.title;
+    const title = options.title || reference.getAttribute('title');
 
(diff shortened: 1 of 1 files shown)


FezVrasta:
The way it works is the same described in the Bootstrap documentation, so it's the expected behavior.

GaryChangCN:
I have the same problem

FezVrasta:
What's wrong with the answer I given?

GaryChangCN:
option.title will override from html attribute

image

FezVrasta:
> The way it works is the same described in the Bootstrap documentation, so it's the expected behavior.

**this is the expected behavior**

GaryChangCN:
okey thanks

PR #356, opened 2017-07-19
Use correct casing for JavaScript vendor-prefixes

### LINKED ISSUE #355: Incorrect casing for vendor-prefixed style properties.
[`getSupportedPropertyName`](https://github.com/FezVrasta/popper.js/blob/08c5d6010346bf9df06e9f81a54fa6c2c51e3639/packages/popper/src/utils/getSupportedPropertyName.js) attempts to return the correct (and potentially vendor-prefixed) style attribute to set based on the browser's support. But using vendor-prefixes in javascript is case-sensitive, and the casing used here is incorrect in all cases except for the `ms` prefix (and, one could argue, 'webkit' though this still causes problems as I'll explain below).

### References:

* [Vendor Prefixes and JavaScript](https://blogs.msdn.microsoft.com/thebeebs/2012/01/11/vendor-prefixes-and-javascript/)
* [Modernizr Prefixed](http://www.andismith.com/blog/2012/02/modernizr-prefixed/)

### Demo, etc.

I'm not sure how to create a CodePen demo to demonstrate this because it's a browser-specific problem, and most browsers are quite happy to ignore it anyway. So while I'm aware of the guidelines for submitting an issue, I don't think they are helpful in this particular case. But I'm happy to add any additional information or details needed.

### Additional comments

Both `webkit` and `Webkit` are apparently supported/acceptable by Chrome and Safari standards. But the lower-cased version actually causes a problem with React. While I realize this isn't in any way related to React, it is what ultimately led me to this problem in the first place.

I am using [React Popper](https://github.com/souporserious/react-popper), which is a React wrapper around this component. It takes the styles supplied by popper.js and hands them to React. When react sees `webkitTransform`, for example, it writes an error to the console, because it expects `WebkitTransform` with a capital "W". There's an argument to be made that this is really a React issue. But since either would be acceptable, it seems to me it would be just as easy to use the capitalized version here.

I'll put in a quick PR for this, if folks are interested in fixing this and I'm not totally missing something here.

joekrill:
Vendor-prefixed names are case sensitive in JavaScript. They should be:

* `ms`
* `Moz`
* `O`
* `Webkit`

Technically `webkit` is also acceptable, but I'm using the capitalized version here because if this is used in a React component, React expects it to be capitalized (and beyond that I'm not aware of the benefits of using one over the other).

Fixes #355

### DIFF
--- a/packages/popper/src/utils/getSupportedPropertyName.js
+++ b/packages/popper/src/utils/getSupportedPropertyName.js
@@ -3,10 +3,10 @@
  * @method
  * @memberof Popper.Utils
  * @argument {String} property (camelCase)
- * @returns {String} prefixed property (camelCase)
+ * @returns {String} prefixed property (camelCase or PascalCase, depending on the vendor prefix)
  */
 export default function getSupportedPropertyName(property) {
-  const prefixes = [false, 'ms', 'webkit', 'moz', 'o'];
+  const prefixes = [false, 'ms', 'Webkit', 'Moz', 'O'];
   const upperProp = property.charAt(0).toUpperCase() + property.slice(1);
 
   for (let i = 0; i < prefixes.length - 1; i++) {


FezVrasta:
released as 1.11.0

PR #363, opened 2017-07-23
[fix] Fixed #276

### LINKED ISSUE #276: Doesn't work inside tables
### CodePen demo
https://codepen.io/anon/pen/RVeQyL

1. Put reference element inside of a table
2. Initialize Popper.js on it

Popper should show up next to reference element.

### What went wrong?

Popper shows up in wrong location, or if page is scrolled, sometimes it gets hidden.

FezVrasta:
I'm a bit concerned about performance problems...

These should/could kick in only when the returned `offsetParent` is a `TD` or a `TABLE`, but still..

### DIFF
--- a/packages/popper/package.json
+++ b/packages/popper/package.json
@@ -35,7 +35,7 @@
     "build": "node bundle.js",
     "lint": "eslint .",
     "test": "popper-karma",
-    "test:dev": "NODE_ENV=development npm run test",
+    "test:dev": "BROWSERS=Chrome NODE_ENV=development npm run test",
     "coverage": "NODE_ENV=coverage npm run test"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


FezVrasta:
I love how people don't give a fuck about all of this..

pinkeen:
It fixed the positioning issue for me 👍 
However the tooltip gets inserted into the parent `<tr>` element what causes some rendering issues with the table.

FezVrasta:
what do you mean "gets inserted"? Popper doesn't touches the position of the nodes in the DOM three

pinkeen:
I'm sorry, I forgot that tooltip.js is a separate project. When using `<td>` as trigger the tooltip is appended to the parent `<tr>`. I don't know if I should file a bug report in tooltip. This being a special case concerning tables and requiring probably a dirty workaround maybe it's better to just force the container.

FezVrasta:
Tooltip.js should append the tooltip to the `parentNode` of the reference element. Please file an issue if you see a different behavior

pinkeen:
That's what I was saying :) - there is no bug, it's appended correctly, although block element inside tr is a no go. This being a special case I will just force body as container.

Keep up the good work!

Issue #302, opened 2017-06-07
Wrong position of popper element with position:fixed & html/body height: 100%


sychevyi:
### CodePen demo

Bad: https://codepen.io/anon/pen/jwPXXq
Without height: https://codepen.io/anon/pen/qjdLwK
Good: https://codepen.io/anon/pen/mwJavg

### Steps to reproduce the problem

1.  Open 'bad' example, scroll content. Popper moves away from reference
2.  Open 'without height' example, scroll content. Popper stays with reference
3.  Open 'good' example, scroll content. Popper stays with reference, as boundariesElement set to 'viewport'

### What is the expected behavior?

Popper stays with reference, without additional settings (?)

### What went wrong?

If html & body are styled with 'height: 100%', content is scrolled and reference is inside 'position: fixed' container, popper moves away from reference.

### Any other comments?

Actually I've found workaround (see 'good' example), just wanted to check if it's a bug or expected behavior.
Thanks for your time!

BTW, thanks for the 'popper-enabled' branch of the 'react-select', works good for me!

nblackburn:
This seems to have been introduced in one of the more recent releases.

FezVrasta:
Looks like [1.9.4](https://github.com/FezVrasta/popper.js/releases/tag/v1.9.4) broken it

hekigan:
Same issue, and as @FezVrasta wrote, if I rollback to v1.9.3, the problem disappears.
In 1.10.8, the problem is still present.

arvinsingla:
I'm also encountering this when using react-popper in our react app. I only have `height: 100%` on body. Unfortunately the workaround used by @sychevyi doesn't suit our UX as I need the popper attached to the scrollParent.

okize:
I'm seeing this issue with `height: 100vh` on body as well. These body height settings are a common technique used to achieve a "sticky" footer.

FezVrasta:
May someone conform it works fine on Safari Mobile? I'm from iPhone now and I see it correctly

okize:
@FezVrasta looking on my iPhone, the good and bad CodePen examples @sychevyi posted look identical. On desktop (Chrome/Mac) the "bad" example has a popover that drifts away from it's ref as you scroll the content.

Johann-S:
Is it a good solution to set : `boundariesElement: 'viewport'` to fix this issue ? 
Because if it's the case we should allow to change boudariesElement in Bootstrap Tooltips 🤔

[...]

PR #368, opened 2017-07-29
Fix #333

### LINKED ISSUE #333: CSS margin not being accounted when positioning arrow
### CodePen demo

https://codepen.io/aristidesfl/pen/dRdLXr

1. Make a popper with an arrow
2. Add a CSS margin to the popper

Arrow should be centered with ref

### What went wrong?

Arrow is not centered.

FezVrasta:
Again I had to add a `getStyleComputedProperty` to properly fix the problem.

I think we should find a way to cache the value during the beginning of each update lifecycle to reduce the DOM accesses..

### DIFF
--- a/packages/popper/src/modifiers/arrow.js
+++ b/packages/popper/src/modifiers/arrow.js
@@ -1,6 +1,7 @@
 import getClientRect from '../utils/getClientRect';
 import getOuterSizes from '../utils/getOuterSizes';
 import isModifierRequired from '../utils/isModifierRequired';
+import getStyleComputedProperty from '../utils/getStyleComputedProperty';
 
 /**
  * @function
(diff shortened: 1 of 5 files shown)


FezVrasta:
ping @aristidesfl, please may you let me know if this fixes your problem?

aristidesfl:
I checked out this branch, built, uploaded popper.js umd and used it in the codepen example: https://codepen.io/aristidesfl/pen/qXbYNY

FezVrasta:
It doesn't look like the right build, there's no instance of `popperMarginSide`, which is a new variable I added to the code to fix this bug.

FezVrasta:
This is a version with the right build of popper.js
https://codepen.io/FezVrasta/pen/vJLjxb

aristidesfl:
Seems to be working, thanks! I don't understand why the version I got after building your PR is not the right own. 

On an unrelated note, should popper follow reference when you scroll in this case? https://codepen.io/aristidesfl/pen/qXbYPy

FezVrasta:
I think you hit #302

PR #375, opened 2017-08-03
Artbitrary typo

dcorb:
getting boundaries right is an art. But in case it was not intended, I created this PR


FezVrasta:
@dcorb tests are failing, may you look at them pls?

PR #394, opened 2017-08-18
fix(popper): Fixed #302

### LINKED ISSUE #302: Wrong position of popper element with position:fixed & html/body height: 100%

FezVrasta:
This should fix #302, I'm not sure if this may introduce unwanted regressions, the whole test suite seem to pass just fine.

I'd love some feedback from all the interested people that reported the bug in #302.

cc @Johann-S because this should affect (possibly positively) Bootstrap v4.

### DIFF
--- a/packages/popper/src/utils/getWindowSizes.js
+++ b/packages/popper/src/utils/getWindowSizes.js
@@ -3,12 +3,14 @@ import isIE10 from './isIE10';
 function getSize(axis, body, html, computedStyle) {
   return Math.max(
     body[`offset${axis}`],
+    body[`scroll${axis}`],
     html[`client${axis}`],
     html[`offset${axis}`],
+    html[`scroll${axis}`],
     isIE10()
       ? html[`offset${axis}`] +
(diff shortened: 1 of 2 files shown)


FezVrasta:
It's failing on Mobile Safari 9 🤔

PR #402, opened 2017-08-24
fix(popper): Fixed regression #401

### LINKED ISSUE #401: Popper position update does not position itself entirely in viewport
### CodePen demo

https://codepen.io/anon/pen/YxvXRe

1.  Resize the browser window to make it smaller (with sufficient speed)

The popper's position should update to the correct position (entirely inside the viewport) instantly

### What went wrong?

It 'lags' and part of the popper is left outside the viewport as the window is resized

### Any other comments?

This seems to be a regression which was fixed by this issue: https://github.com/FezVrasta/popper.js/issues/251

FezVrasta:
This should fix #401 and don't break #302.

Please @atomiks may you verify that both the tickets are properly fixed please?

### DIFF
--- a/packages/popper/src/utils/getBoundaries.js
+++ b/packages/popper/src/utils/getBoundaries.js
@@ -50,7 +50,7 @@ export default function getBoundaries(
 
     // In case of HTML, we need a different computation
     if (boundariesNode.nodeName === 'HTML' && !isFixed(offsetParent)) {
-      const { height, width } = getWindowSizes();
+      const { height, width } = getWindowSizes(false);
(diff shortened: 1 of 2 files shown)


atomiks:
Yes it fixes the issue I made, but can't verify the other issue atm

FezVrasta:
Going to merge, if problems arises people of #302 blame yourself, you had plenty of time to help me verify this PR.

Thank you again atomiks for your help! 🤗

nheath99:
Hi this does indeed break #302

FezVrasta:
Thank you for the heads up, I'll se what I can do.

PR #415, opened 2017-09-07
Support references within iframes

kasperisager:
This PR fixes an issue related to cases where references are contained within iframes, e.g.:

```html
<div class="popper">...</div>
<iframe src="..."></iframe>
```

```js
const popper = document.querySelector('.popper')
const iframe = document.querySelector('iframe')
const reference = iframe.contentDocument.querySelector('.reference')

new Popper(reference, popper, { ... })
```

The `getScrollParent()` and `setupEventListeners()` utilities will now correctly return the `body` and `window` references, respectively, of the associated iframe rather than the host frame.

### DIFF
--- a/packages/popper/src/utils/getScrollParent.js
+++ b/packages/popper/src/utils/getScrollParent.js
@@ -10,11 +10,16 @@ import getParentNode from './getParentNode';
  */
 export default function getScrollParent(element) {
   // Return body, `getScroll` will take care to get the correct `scrollTop` from it
-  if (
-    !element ||
-    ['HTML', 'BODY', '#document'].indexOf(element.nodeName) !== -1
-  ) {
-    return window.document.body;
+  if (!element) {
+    return window.document.body
+  }
+
+  switch (element.nodeName) {
+    case 'HTML':
+    case 'BODY':
+      return element.ownerDocument.body
+    case '#document':
+      return element.body
   }
 
   // Firefox want us to check `-x` and `-y` variations as well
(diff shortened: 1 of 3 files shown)


packages/popper/src/utils/getScrollParent.js
```diff
@@ -10,11 +10,16 @@ import getParentNode from './getParentNode';
  */
 export default function getScrollParent(element) {
   // Return body, `getScroll` will take care to get the correct `scrollTop` from it
-  if (
-    !element ||
-    ['HTML', 'BODY', '#document'].indexOf(element.nodeName) !== -1
-  ) {
-    return window.document.body;
+  if (!element) {
```
FezVrasta: Can't this be the `default:` case of the below switch?

FezVrasta:
Looks good thanks! Waiting for CI to finish its stuff

FezVrasta:
Ok everything looks good, may you increase the file size limit [here](https://github.com/FezVrasta/popper.js/blob/master/package.json#L40) to `7.5` so that we have some time until I decide how to reduce the size of the lib again?

kasperisager:
Sure thing 👌

kasperisager:
⚠️  Heads up: I just realised that this will break if used with sandboxed iframes that don't allow scripts as the scroll listener is run in the context of the iframe.

FezVrasta:
I don't think this is a problem right? I mean, it's a browser limitation, it's up to the consumer to make sure he's not breaking the rules.

kasperisager:
Shouldn't be a problem, no; people indeed have to mind browser restrictions as always 👍

PR #416, opened 2017-09-07
Add callback as param to update() and scheduleUpdate()

atomiks:
Can write tests if need be. I'm not sure how hefty using spread/rest params is here, so feel free to suggest fixes! First ever PR btw, 😨

### DIFF
--- a/packages/popper/src/index.js
+++ b/packages/popper/src/index.js
@@ -90,8 +90,8 @@ export default class Popper {
 
   // We can't use class properties because they don't get listed in the
   // class prototype and break stuff like Sinon stubs
-  update() {
-    return update.call(this);
+  update(callback) {
+    return update.call(this, callback);
   }
   destroy() {
     return destroy.call(this);
@@ -108,7 +108,7 @@ export default class Popper {
    * @method scheduleUpdate
    * @memberof Popper
    */
-  scheduleUpdate = () => requestAnimationFrame(this.update);
+  scheduleUpdate = callback => requestAnimationFrame(() => this.update(callback));
 
   /**
    * Collection of utilities useful when writing custom modifiers.
--- a/packages/popper/src/methods/update.js
+++ b/packages/popper/src/methods/update.js
@@ -8,9 +8,10 @@ import runModifiers from '../utils/runModifiers';
  * the new style.<br />
  * Prefer `scheduleUpdate` over `update` because of performance reasons.
  * @method
+ * @param {Function} callback - Callback to run once position has updated
  * @memberof Popper
  */
-export default function update() {
+export default function update(callback) {
   // if popper is destroyed, don't perform any further update
   if (this.state.isDestroyed) {
     return;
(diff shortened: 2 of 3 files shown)


packages/popper/src/index.js
```diff
@@ -90,8 +90,8 @@ export default class Popper {
 
   // We can't use class properties because they don't get listed in the
   // class prototype and break stuff like Sinon stubs
-  update() {
-    return update.call(this);
+  update(callback) {
```
Johann-S: IMO `callback` should be optional because I don't think everybody will add a listener plus, it add compatibility with older release of Popper.js

packages/popper/src/methods/update.js
```diff
@@ -66,4 +67,8 @@ export default function update() {
   } else {
     this.options.onUpdate(data);
   }
+
+  if (typeof callback === 'function') {
```
FezVrasta: We have a `isFunction` utility for this cases

packages/popper/src/utils/debounce.js
```diff
@@ -13,19 +13,21 @@ for (let i = 0; i < longerTimeoutBrowsers.length; i += 1) {
 export function microtaskDebounce(fn) {
   let scheduled = false;
   let i = 0;
+  let _args = [];
```
FezVrasta: What is this `_args` needed for? I see you assign to it the arguments but never use it?

atomiks: It's used in the mutation observer callback `fn(..._args)`

packages/popper/src/methods/update.js
atomiks: Ah, okay

packages/popper/src/index.js
atomiks: What do you mean optional? If no argument is supplied nothing will occur

FezVrasta: As far I see it's already optional

Johann-S: Yep that's right you check below with `typeof` everything is fine 👍

packages/popper/src/utils/debounce.js
FezVrasta: Ok... Do we actually need all of this?

The callback doesn't seem to be debounced anywhere?

atomiks: `this.update` is debounced. It returns a function so you need to do the spread/rest thingy to allow it to be supplied with arguments, no?

atomiks: Technically you can just use a single param instead of spread rest if you only want to supply 1 single argument, but with more you'd do that. Depends if you'd add more in the future (doubt it)?

FezVrasta: Oh right, I get it now.

Looks good then, may you add some unit test for this new functionality as well? We already have some tests in the units folder.

atomiks:
After playing with my implementation here it doesn't really do what I want so I guess I'll close this now, maybe can resolve it in the future if need be...

PR #418, opened 2017-09-08
fix(popper) guard against null onUpdate + onCreate methods

kybishop:
No-op functions have a measurable memory and runtime impact. Better to check a boolean to see if a function should be run or not.

This has the added benefit of allowing external libs to pass in `null` values, making Popper.js slightly more ergonomic.


kybishop:
Closing this while I investigate some benchmarks

PR #425, opened 2017-09-12
Rely on `ownerDocument` references for both poppers and reference elements

kasperisager:
This is a continuation of #415 that ensures that both poppers and reference elements rely on `ownerDocument` references whenever needed in order to support poppers and reference elements residing within iframes. Upon further testing I encountered some cases that #415 did not address that this PR does.

### DIFF
--- a/packages/popper/src/utils/getWindow.js
+++ b/packages/popper/src/utils/getWindow.js
@@ -1,6 +1,7 @@
 /**
  * Get the window associated with the element
  * @argument {Element} element
+ * @returns {Window}
  */
 export default function getWindow(element) {
   const ownerDocument = element.ownerDocument;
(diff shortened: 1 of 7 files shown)


packages/popper/src/utils/removeEventListeners.js
```diff
@@ -5,8 +5,11 @@
  * @private
  */
 export default function removeEventListeners(reference, state) {
+  const ownerDocument = reference.ownerDocument
```
FezVrasta: This is repeated in `setupEventListeners`, may we move this to its own module?

kasperisager: Certainly!

packages/popper/src/utils/getWindow.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Get the window associated with the element
+ * @argument {Element} element
```
FezVrasta: missing @returns annotation

kasperisager: Added 👌

PR #438, opened 2017-09-29
Remove event listeners properly on .dispose()

matehat:
These event listeners are added at Tooltip construction time, whereas `_tooltipNode` is added the first time the tooltip is shown. That means the listeners will always be added on the target node, whether we end up showing the tooltip or not. We should then remove them regardless of the existence of `_tooltipNode`

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -252,16 +252,17 @@ export default class Tooltip {
       // destroy instance
       this.popperInstance.destroy();
 
-      // remove event listeners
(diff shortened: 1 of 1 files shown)


rstuart85:
I ran into the same problem and submitted a PR with tests in #467

matehat:
Thanks @rstuart85 for landing this fix, for some reason I never got their CLAassistant thingy to work 😆

FezVrasta:
No problem, unfortunately the CLa assistant has some problems...

PR #443, opened 2017-10-04
Fix 372

### LINKED ISSUE #372: Add .d.ts to tooltip.js
Currently popper.js has an .d.ts file attached. Is it possible to have one for tooltip.js also.

Benefit: tooltip.js could be used in Typescript, without workarounds by adding an empty declaration, or even more worse things ;-)

elebetsamer:
My attempt at creating a type definition for tooltip.js. I used the type definition from popper.js as a reference point.

This is my first attempt at writing a type definition, so it may need to be tweaked.

If everything looks good, this should close #372.

### DIFF
--- a/packages/tooltip/index.d.ts
+++ b/packages/tooltip/index.d.ts
@@ -1,0 +1,55 @@
+// Type definitions for tooltip.js 1.1.5
+// Project: https://github.com/FezVrasta/popper.js/
+// Definitions by: elebetsamer <https://github.com/elebetsamer>
+
+// This file only declares the public portions of the API.
+// It should not define internal pieces such as utils or modifier details.
+
+/// <reference types="jquery"/>
+/// <reference types="popper.js"/>
+
+declare namespace Tooltip {
+  export type Placement = 'top-start'
+  | 'top'
+  | 'top-end'
+  | 'right-start'
+  | 'right'
+  | 'right-end'
+  | 'bottom-end'
+  | 'bottom'
+  | 'bottom-start'
+  | 'left-end'
+  | 'left'
+  | 'left-start';
+
+  export interface TooltipOptions {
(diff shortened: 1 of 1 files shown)


elebetsamer:
Not sure why, but it won't let me sign the CLA. It says "Eric Lebetsamer" seems not to be a GitHub user and tells me to add my email address to my account, but I've already done this.

FezVrasta:
Thanks, don't worry about the CLA, sometimes it doesn't work 😓

Is there any way to test this PR to make sure it works properly?

elebetsamer:
Hmm, I'm not really sure how it could be tested. This is the first time I've ever tried to create type definitions, so I'm a bit out of my depth. I tried to base it on the popper type definition though, so if that works, I would hope this one did as well. Maybe we could ask one of the creators from the popper type definitions for some guidance?

elebetsamer:
I tested locally in VS Code and it seems to work as expected when I directly reference the typings in the file I'm working on: `/// <reference path="../../../tooltip-js.d.ts" />`

But it doesn't work when I try to reference it from typings.d.ts:
```
/// <reference path="node_modules/popper.js/index.d.ts" />
/// <reference path="tooltip-js.d.ts" />
```

eaw-tooltip_tsx_ _evolved-intranet-stencil
eaw-tooltip_tsx_ _evolved-intranet-stencil 2
eaw-tooltip_tsx_ _evolved-intranet-stencil 3

FezVrasta:
Sorry for the delay. Any update on this?

Doesn't TypeScript provide a way to test the type definitions like Flow does?

elebetsamer:
I was able to get it working by manually referencing the type definition file. Other than that, I don't know how else to test this.

FezVrasta:
Could something like this be used?

https://www.npmjs.com/package/typescript-definition-tester

FezVrasta:
Closing for #539

PR #445, opened 2017-10-04
fix #437 by not using the reserved keyword function as property

### LINKED ISSUE #437: Reserved Keyword breaks minifying by YUICompressor
We are using popper.js within the CMS System AEM. Because of several technical restrictions, we are forced to use YUIcompressor for asset minification in the backend.

YUIcompressor seems to be more strict about reserved keywords and fails to compress javascript code containing reserved keywords as function / property names.

Popper.js  and popper-utils.js seem to contain such function names both in that function:

```
function runModifiers(modifiers, data, ends) {
  var modifiersToRun = ends === undefined ? modifiers : modifiers.slice(0, findIndex(modifiers, 'name', ends));

  modifiersToRun.forEach(function (modifier) {
    if (modifier.function) {
      console.warn('`modifier.function` is deprecated, use `modifier.fn`!');
    }
    var fn = modifier.function || modifier.fn;
    if (modifier.enabled && isFunction(fn)) {
      data = fn(data, modifier);
    }
  });

  return data;
}
```
Would it be possible to replace them by this code:

```diff
function runModifiers(modifiers, data, ends) {
  var modifiersToRun = ends === undefined ? modifiers : modifiers.slice(0, findIndex(modifiers, 'name', ends));

  modifiersToRun.forEach(function (modifier) {
-   if (modifier.function) {
+   if (modifier['function']) {
      console.warn('`modifier.function` is deprecated, use `modifier.fn`!');
    }

-   var fn = modifier.function || modifier.fn;
+   var fn = modifier['function'] || modifier.fn;
    if (modifier.enabled && isFunction(fn)) {
      data = fn(data, modifier);
    }
  });

  return data;
}
```

This passes the YUIcompressor.

1.  Code can be tested here: https://99webtools.com/yui-compressor.php#output

Minification by the YUIcompressor works without errors

### What went wrong?

Javascript code contains functions named after reserved keywords.

janwidmer:

### DIFF
--- a/packages/popper/src/utils/runModifiers.js
+++ b/packages/popper/src/utils/runModifiers.js
@@ -18,10 +18,10 @@ export default function runModifiers(modifiers, data, ends) {
     : modifiers.slice(0, findIndex(modifiers, 'name', ends));
 
   modifiersToRun.forEach(modifier => {
-    if (modifier.function) {
+    if (modifier['function']) { // eslint-disable-line dot-notation
       console.warn('`modifier.function` is deprecated, use `modifier.fn`!');
     }
-    const fn = modifier.function || modifier.fn;
+    const fn = modifier['function'] || modifier.fn; // eslint-disable-line dot-notation
     if (modifier.enabled && isFunction(fn)) {
       // Add properties to offsets to make them a complete clientRect object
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thanks!

PR #449, opened 2017-10-07
Add jsDelivr links

LukasDrgon:
I added [jsDelivr CDN links](https://www.jsdelivr.com/package/npm/popper.js) to your readme as an alternative CDN. jsDelivr is the fastest opensource CDN available and built specifically for production usage. It can serve any project from npm with zero config and offers a large network and better reliability.


FezVrasta:
Thanks but I had bad experiences with it in the past and I'm not willing to add it to the readme.

MartinKolarik:
@FezVrasta mind sharing what problems you had with it?

FezVrasta:
The service refused to automatically publish new versions after I released the first v1 beta. I reported the problem and nobody replied for several weeks.

I rather point my users to more reliable and professional services like unpkg.com and cdnjs

PR #450, opened 2017-10-07
Fix #328 Tooltip offset

### LINKED ISSUE #328: Tooltip.js: offset is never set (either with "offset" or "modifiers")
- Tooltip.js v1.1.4
- Popper.js v1.10.6

### CodePen demo

https://codepen.io/hekigan/pen/GEyazB/?editors=1011

On the codepen above, you can see that the **offset** is never set.

```
{
    ...
    offset: 100,
    popperOptions: {
      modifiers: {
        offset: {
          offset: 200
        }
      }
    }
```

The tooltip should have shifted from the target element by 'x' pixels.

### What went wrong?

No change whatsoever.

### Any other comments?

I looked up in the 'tooltip.js' source code and I could not find where you set the offset and pass it on to 'popper.js'.
I did see that you merge the options object, but after that, it just does not seem to get used.

Akryum:

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -223,6 +223,21 @@ export default class Tooltip {
       };
     }
 
+    if (options.offset) {
+      let offset = options.offset;
+
+  		// One value -> switch to offset from tooltip target
+      const typeofOffset = typeof options.offset
+  		if (typeofOffset === 'number' || (typeofOffset === 'string' && offset.indexOf(',') === -1)) {
+  			offset = `0, ${offset}`;
+  		}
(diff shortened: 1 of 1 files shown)


FezVrasta:
Hi, sorry for the unresponsiveness!

May you add a test to make sure we don't regress here and the PR does actually fix the problem?

FezVrasta:
Are you still interested in this PR @Akryum?

Akryum:
I didn't have time to write tests yet.

FezVrasta:
Ok no problem! Let me know if you need any help

PR #454, opened 2017-10-12
fix detecting scrollParent: use reference as source, not popper. Fixes #453

### LINKED ISSUE #453: scrollParent should be calculated from reference, not popper
### CodePen demo
Popper should hide when ref is out of scrollable div bounds, instead it gets hidden when ref gets scrolled outside body:
https://codepen.io/anon/pen/dVjRba

1.  Put popper outside of reference's scrollParent
2.  Use scrollParent as boundariesElement

Popper should try to keep inside reference's scrollParent

### What went wrong?
Popper's scrollParent is used as boundariesElement

SpadarShut:
Fix for https://github.com/FezVrasta/popper.js/issues/453

### DIFF
--- a/packages/popper/src/utils/getBoundaries.js
+++ b/packages/popper/src/utils/getBoundaries.js
@@ -33,7 +33,7 @@ export default function getBoundaries(
     // Handle other cases based on DOM element used as boundaries
     let boundariesNode;
     if (boundariesElement === 'scrollParent') {
-      boundariesNode = getScrollParent(getParentNode(popper));
+      boundariesNode = getScrollParent(getParentNode(reference));
       if (boundariesNode.nodeName === 'BODY') {
         boundariesNode = window.document.documentElement;
       }
--- a/packages/popper/tests/functional/core.js
+++ b/packages/popper/tests/functional/core.js
@@ -1615,4 +1615,77 @@ describe('[core]', () => {
       },
     });
   });
+
+  // test for #453
+  // When the reference is inside a scrollParent, it should be used as the
+  // source for choosing the scrollParent rather than the popper.
+  // This is particularly useful when the popper lives outside of the
+  // scrollParent, e.g. in the body. In such cases, this allows the popper to
+  // adjust (e.g. hide) based on the actual scrollParent boundaries rather than
+  // that of the body.
+  it(
+    'reference is used as source for choosing scrollParent',
+    done => {
+      jasmineWrapper.innerHTML = `
+      <style>
+        body {
+          padding: 100px;
+        }
+
+        .scrollParent {
+          height: 300px;
+          width: 300px;
+          overflow: auto;
+          position: relative;
+        }
+
+        .scrollContent {
+          background: gray;
+          padding: 1000px;
+        }
+
+        #reference {
+          background: lightgrey;
+          height: 25px;
+          width: 100px;
+        }
+
+        #popper {
+          background: cyan;
+          height: 150px;
+          width: 150px;
+        }
+
+        [x-out-of-boundaries] {
+          visibility: hidden;
+        }
+      </style>
+
+      <div class="scrollParent">
+        <div class="scrollContent">
+          <div id="reference">ref</div>
+        </div>
+      </div>
+      <div id="popper">pop</div>
+    `;
+
+      const reference = document.getElementById('reference');
+      const popper = document.getElementById('popper');
+      const scrollParent = document.querySelector('.scrollParent');
+
+      new Popper(reference, popper, {
+        onCreate() {
+          // Scroll reference just out of view to the right
+          simulateScroll(scrollParent, { scrollTop: 950, scrollLeft: 700, delay: 100 });
+        },
+        onUpdate(data) {
+          // Popper is hidden when reference is out of view in scrollParent
+          expect(data.hide).toBe(true);
+
+          data.instance.destroy();
+          done();
+        },
+      });
+    }
+  );
 });


SpadarShut:
@FezVrasta I cannot get this licence agreement signed. The button on the page just reload the page. Can this be a problem with this repo?

FezVrasta:
Thanks for the PR, nevermind the CLA service, it's plain broken, gonna disable it anytime soon...

brentertz:
I tried this fix locally and this seems to work for my use case. 👍

packages/popper/src/utils/getBoundaries.js
```diff
@@ -33,7 +33,7 @@ export default function getBoundaries(
     // Handle other cases based on DOM element used as boundaries
     let boundariesNode;
     if (boundariesElement === 'scrollParent') {
-      boundariesNode = getScrollParent(getParentNode(popper));
+      boundariesNode = getScrollParent(getParentNode(reference));
```
FezVrasta: I'm not really confident in change this line... It was as it was for a good reason I think.

May you add a test for your changes so that at least we know there aren't regressions on this front?

SpadarShut: Can you give a hint how to write a test for this case?)

FezVrasta: There are plenty of tests you can use as base, for example:
https://github.com/FezVrasta/popper.js/blob/master/packages/popper/tests/functional/core.js#L88

brentertz: I hope that you don't mind, but I took a stab at writing a test for this as I am excited about the fix.  

You can see it on my fork here --> https://github.com/FezVrasta/popper.js/compare/master...brentertz:fix-453

I wasn't sure if there was another assertion that could be made, perhaps around boundary rects, etc, or something else entirely, but nothing else was obvious to me.  I am happy to update if you have further suggestions.

In either case, if you end up wanting the test, I am also not really sure how to best get it to you in terms of a commit?  Should I submit another PR, or somehow add it to this one?

FezVrasta: @brentertz you should be able to send a PR to @SpadarShut's fork
https://github.com/SpadarShut/popper.js/tree/fix-scrollparent-detection

FezVrasta:
This is the last time someone touched this line of code:
https://github.com/FezVrasta/popper.js/commit/491a4e504370a60ed5df29dbed1967b0bdb0473c

Apart from that, this line always used `popper` and not `reference`... So I'm really scared about changing this line 😣

We'll need some heavy manual testing I guess.

SpadarShut:
Looks like all tests that use `appendNewPopper` append it to the same scrollParent as reference, but still all test pass. I guess this is a sign that the system is well written =)

packages/popper/src/utils/getBoundaries.js
brentertz: Hello @SpadarShut , I submitted a PR to your fork - not sure if you noticed.  If you decide to merge it, it will be combined with this PR here and we'll then have the fix and a regression test as well, hopefully unblocking this PR.

SpadarShut: @brentertz Thanks! Didn't get any emails. Will acceplt is of course.

FezVrasta:
Sorry I have to hold on this for a bit until I find some time to properly test the PR manually. I hope to find some time this weekend

brentertz:
I am happy to help where I can if you have suggestions, though I have some difficulty running the repo locally.

FezVrasta:
What happens with the repo?

brentertz:
I'm not exactly sure the source of my issues with the repo.  I am unable to run npm scripts in the root of the package.  Neither npm nor yarn seem to recognize any of the binaries that are installed from the npm packages and the scripts fail with things like `sh: lerna: command not found` or `sh: bundlesize-init: command not found`.  I imagine that this may be a PATH issue on my end, but I haven't figured it out yet.  If I cd into packages/popper, then I can run the npm scripts there just fine.

FezVrasta:
the `pkg.scripts` scripts should automatically add `$(npm bin)/` to your PATH. At least, that's how npm/Yarn work... 🤔

FezVrasta:
Oh, note that the repository makes use of Yarn workspaces, so you need at least Yarn 1.x to have everything working properly.

brentertz:
Hi @FezVrasta Have you had any more thoughts or had a chance to do additional testing on this PR?

After some digging, I was able to get the project building properly.  **I had to downgrade my yarn version from 1.2.1 to 1.0.1.**  I noticed that this was the version that you were using in `.travis.yml`.  When using 1.2.1, `node_modules/.bin` was not created and no binaries were installed, thus package scripts were failing.

Is there anything that I can do to help this PR along?  Further validating the use case that you mentioned above perhaps? https://github.com/FezVrasta/popper.js/commit/491a4e504370a60ed5df29dbed1967b0bdb0473c Adding additional test coverage elsewhere?

I believe that this update will provide more commonly expected behavior when poppers are placed inside scrollable containers, and the popper node is rendered to the body - e.g. when using a portal.

In addition to the aforementioned examples, below is a link that may better illustrate my use case.  Ideally, the popper should hide when the reference element is scrolled out of view.

https://395-popover-portal--mineral-ui.netlify.com/components/dropdown/portal/

I’d appreciate your thoughts…

FezVrasta:
Hey sorry I've been in conferences the whole month so I had little time to dedicate to this project.

I promise I'm going to look at it in the next few days. I'm very sorry for the long delays 😓

brentertz:
No worries,  I understand.  I'll look forward to hearing back from you in several days. :)

brentertz:
@FezVrasta Thanks for merging this PR.  Any idea when this update may be published?

FezVrasta:
Probably this weekend

andrewnicols:
This change has broken Popper for us in Moodle.

In our case it was correct to follow the popper rather than the reference.

Any chance we can add a change to make this a config option somehow? I haven't had a look at exactly how this translates to master.

PR #460, opened 2017-10-16
Test behavior of MutationBehavior instead of polyfill detcting.

### LINKED ISSUE #452: isNative detection fails with NewRelic instrumentation
### CodePen demo

https://codepen.io/jdelStrother/pen/LzrBXE

1. Add NewRelic instrumentation, or any other instrumentation that wraps native functions (MutationObserver in particular)
2. Try to display something with Popper

Popper uses MutationObserver and immediately displays the popped element at the correct position

### What went wrong?

Popper's isNative check fails, causing it to fall back to `setTimeout(fn, 0)`, causing the popped element to appear for a split second at the wrong position, before moving into place once the timeout fires.

### Any other comments?

`window.MutationObserver.toString = function() { return "[native code]" }` is a somewhat awful hack that works around Popper's isNative check.

blainesch:
Fixes #452

~~~ html
<script src='./packages/popper/dist/umd/popper.js'></script>
<script>
   const refEl = document.getElementById('ref');
   const popEl = document.getElementById('pop');
   var i = new Popper(refEl, popEl, {}); // macro
   i.update() // macro
   setTimeout(function () {
      i.update() // micro
   })
</script>
~~~

~~~ html
<script src='./packages/popper/dist/umd/popper.js'></script>
<script>
   const refEl = document.getElementById('ref');
   const popEl = document.getElementById('pop');
   setTimeout(function () {
      var i = new Popper(refEl, popEl, {}); // micro
      i.update() // micro
   })
</script>
~~~

An alternative would be to use `Promise` instead of `MutationObserver` to use the micro task queue. Promise has less browser support, but all polyfills use `setTimeout` as a fallback so you just need to detect if promise is on the page, which means the first example would work better.


.nvmrc
```diff
@@ -0,0 +1 @@
+v7.6.0
```
FezVrasta: nvm is super slow, better **n** 😉

Anyway, please remove this file.

FezVrasta:
Tests are failing 😞

blainesch:
This is passing locally for me. How do I reproduce the error locally? CI doesn't have a long enough stack trace to see what the issue is.

The alternative pull request, using promises, is passing. #461

FezVrasta:
It seems to fail on IE

blainesch:
I have an IE VM, are there instructions on how to run the tests in IE?

I tried replicating the travis file, but I keep getting "firewall" errors, even when I disable mine.

FezVrasta:
You may get a OSS browserstack license. Or if you use a VM you just have to open the karma runner URL in your VM browser

blainesch:
There we go [:

🚀

blainesch:
Closing in favor of #461

PR #461, opened 2017-10-16
Use Promises to debounce instead of MutationObserver.

blainesch:
This is an alternative to #460

This would be completely sync, but relies on Promises instead of MutationObserver, so it has less support.

~~~ html
<script src='./packages/popper/dist/umd/popper.js'></script>
<script>
   const refEl = document.getElementById('ref');
   const popEl = document.getElementById('pop');
   var i = new Popper(refEl, popEl, {}); // micro
   i.update() // micro
   setTimeout(function () {
      i.update() // micro
   })
</script>
~~~


FezVrasta:
I'm not sure if promises have the same effect of the MutationObserver, how did you verify it?

blainesch:
You can test it's using the microtask queue using this code snippet:

~~~js
var mutationFn = mutationDebounce(() => {
  console.log('two')
})
var promiseFn = promiseDebounce(() => {
  console.log('three')
})

setTimeout(() => {
  console.log('four')
})
console.log('one')
mutationFn()
promiseFn()

// Copied from #460
function mutationDebounce (fn) {
  if (!window.MutationObserver) {
    return;
  }
  let scheduled = false;
  let i = 0;
  const elem = document.createElement('span');

  // MutationObserver provides a mechanism for scheduling microtasks, which
  // are scheduled *before* the next task. This gives us a way to debounce
  // a function but ensure it's called *before* the next paint.
  const observer = new window.MutationObserver(() => {
    fn();
    scheduled = false;
  });

  observer.observe(elem, { attributes: true });

  return () => {
    if (!scheduled) {
      scheduled = true;
      elem.setAttribute('x-index', i);
      i = i + 1; // don't use compund (+=) because it doesn't get optimized in V8
    }
  };
}

// Copied from #461
function promiseDebounce (fn) {
  let called = false
  return () => {
    if (called) {
      return
    }
    called = true
    Promise.resolve().then(() => {
      called = false
      fn()
    })
  }
}
~~~

FezVrasta:
Thanks! In your example the promise debouncer runs always after the mutationobserver one tho. Doesn't it proof that the mutationobserver is still "faster"?

blainesch:
It only runs first because I called it first.

```js
var mutationFn = mutationDebounce(() => {
  console.log('three')
})
var promiseFn = promiseDebounce(() => {
  console.log('two')
})

setTimeout(() => {
  console.log('four')
})
console.log('one')
promiseFn()
mutationFn()
```

FezVrasta:
Right, lol

blainesch:
@FezVrasta  Thank!

Do you have a timeline for the next release?

PR #462, opened 2017-10-17
Don't rely only on a defined window to determine if we have a DOM

BowlingX:
I came along this issue where certain server side renderers have window defined but no document.
This will add an additional check if `window.document` is not undefined.

### DIFF
--- a/packages/popper/src/utils/debounce.js
+++ b/packages/popper/src/utils/debounce.js
@@ -1,6 +1,6 @@
 import isNative from './isNative';
 
-const isBrowser = typeof window !== 'undefined' && typeof window.document !== 'undefined';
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thanks!

Any reason why `typeof window.document !== 'undefined'` can't be used?

BowlingX:
window can be undefined. So we have to check on both window and the document.

FezVrasta:
Yeah I mean: `typeof window !== 'undefined' && typeof window.document !== 'undefined';`

BowlingX:
yeah :) this would work too. I just thought the second one is shorter.

FezVrasta:
It's just for consistency with the existing code.

Issue #466, opened 2017-10-22
`dispose()` doesn't work correctly if a tooltip has never been shown.


rstuart85:
### CodePen demo

https://codepen.io/rstuart85/pen/veMmqO

### Steps to reproduce the problem

1.  Wait for page to load.
2.  Hover the green element.

### What is the expected behavior?

The tooltip "Test" renders below the element.

### What went wrong?

Both the tooltips "Test" and "I am very very long =D" renders below the element despite the second having had `dispose()` called on it.

### Any other comments?

I've submitted a PR for this issue.

PR #467, opened 2017-10-22
Fix bug with `dispose()`

rstuart85:
When you call dispose() on a tip that hasn't been shown at least once, it
leaves behind some events that will still render the tip.

See issue #466


rstuart85:
The CLAassistant seems to do nothing. I click sign and nothing happens. I've authorised the application and I agree the the Contributor License Agreement.

PR #487, opened 2017-11-08
add positionFixed property, making popper with position: 'Fixed' as an option

MrFrankel:

### DIFF
--- a/packages/popper/src/modifiers/applyStyle.js
+++ b/packages/popper/src/modifiers/applyStyle.js
@@ -49,7 +49,7 @@ export function applyStyleOnLoad(
   state
 ) {
   // compute reference element offsets
-  const referenceOffsets = getReferenceOffsets(state, popper, reference);
+  const referenceOffsets = getReferenceOffsets(state, popper, reference, options.positionFixed ? window.document.documentElement : undefined);
 
   // compute auto placement, store placement inside the data object,
   // modifiers will be able to edit `placement` if needed
@@ -67,7 +67,7 @@ export function applyStyleOnLoad(
 
   // Apply `position` to popper before anything else because
   // without the position applied we can't guarantee correct computations
-  setStyles(popper, { position: 'absolute' });
+  setStyles(popper, { position: options.positionFixed ? 'fixed' : 'absolute' });
 
   return options;
 }
--- a/packages/popper/src/utils/getBoundingClientRect.js
+++ b/packages/popper/src/utils/getBoundingClientRect.js
@@ -18,8 +18,8 @@ export default function getBoundingClientRect(element) {
   // IE10 10 FIX: Please, don't ask, the element isn't
   // considered in DOM in some circumstances...
   // This isn't reproducible in IE10 compatibility mode of IE11
-  if (isIE10()) {
-    try {
+  try {
+    if (isIE10()) {
       rect = element.getBoundingClientRect();
       const scrollTop = getScroll(element, 'top');
       const scrollLeft = getScroll(element, 'left');
@@ -27,10 +27,12 @@ export default function getBoundingClientRect(element) {
       rect.left += scrollLeft;
       rect.bottom += scrollTop;
       rect.right += scrollLeft;
-    } catch (err) {}
-  } else {
-    rect = element.getBoundingClientRect();
+    }
+    else {
+      rect = element.getBoundingClientRect();
+    }
   }
+  catch(e){}
 
   const result = {
     left: rect.left,
--- a/packages/popper/tests/functional/core.js
+++ b/packages/popper/tests/functional/core.js
@@ -14,1678 +14,1724 @@ import getRect from '../utils/getRect';
 const isIE10 = navigator.appVersion.indexOf('MSIE 10') !== -1;
 const isIPHONE = window.navigator.userAgent.match(/iPhone/i);
 const arrowSize = 5;
+[true, false].forEach(positionFixed => {
 
-describe('[core]', () => {
-  afterEach(function() {
-    jasmineWrapper.scrollTop = 0;
-    jasmineWrapper.scrollLeft = 0;
-  });
 
-  it('can access the AMD module to create a new instance', () => {
-    // append popper element
-    const popper = document.createElement('div');
-    popper.style.width = '100px';
-    popper.style.height = '100px';
-    jasmineWrapper.appendChild(popper);
+  describe('[core]' + (positionFixed ? ' Fixed' : ''), () => {
 
-    // append trigger element
-    const trigger = document.createElement('div');
-    jasmineWrapper.appendChild(trigger);
 
-    // initialize new popper instance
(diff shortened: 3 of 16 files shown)


packages/popper/src/modifiers/applyStyle.js
```diff
@@ -67,7 +67,7 @@ export function applyStyleOnLoad(
 
   // Apply `position` to popper before anything else because
   // without the position applied we can't guarantee correct computations
-  setStyles(popper, { position: 'absolute' });
+  setStyles(popper, { position: options.positionFixed ? 'fixed' : 'absolute' });
```
FezVrasta: May you move this logic to the `computeStyles` modifier please?

packages/popper/tests/functional/core.js
```diff
@@ -1366,6 +1366,33 @@ describe('[core]', () => {
     });
   });
 
+  it('checks that positionFixed works correctly', done => {
```
FezVrasta: I fear we'll need way more tests than just this 😞

packages/popper/src/modifiers/applyStyle.js
MrFrankel: will that not cause computation issues? it specifically says in the comment to apply the position ASAP

packages/popper/tests/functional/core.js
MrFrankel: Yea, For sure, just want'd to make sure this PR is relevant first. will add more

FezVrasta: Ideally the whole test suite (or at least the ones in core) should run again but with this option set to true, to make sure all the edge cases are supported.

MrFrankel: If the rest of the code seems good to you all add more test.
btw, are you ok with the name, 'positionFixed'?

MrFrankel: @FezVrasta Is there a way to pass params to the spec files in the test? I wouldnt want to copy/paste all the tests with positionFixed: true set. it would be better if we can run all test twice, once with the flag on, once without.

That way all future test will apply too.

FezVrasta: I think you may wrap the tests inside a[false, true].forEach(positionFixed =>

MrFrankel: works like a charm, thanks

packages/popper/src/modifiers/applyStyle.js
MrFrankel: will do

MrFrankel:
seems some IE11 test are failing, arggg.... Ill fix them... damn you IE!

MrFrankel:
There seems to be an issue with the test, all are failing due to a some type error in IE.
i cant reproduce it even on Saucelabs directly reproducing the test.

@FezVrasta Is there a way I can run the tests locally for IE with sacuelabs?

FezVrasta:
I think the easiest way is to download a free virtual machine from modern.ie and use it to run the tests.

You can copy the karma runner URL and open it in the VM browser and everything should work.

MrFrankel:
copied the karma URL to browser stack, its an IE issue, not sure why my commit raised it. possibly because we run them twice but im not sure.

https://connect.microsoft.com/IE/feedback/details/829392/calling-getboundingclientrect-on-an-html-element-that-has-not-been-added-to-the-dom-causes-unspecified-error

packages/popper/src/utils/getBoundingClientRect.js
```diff
@@ -18,19 +18,21 @@ export default function getBoundingClientRect(element) {
...
       rect.top += scrollTop;
       rect.left += scrollLeft;
       rect.bottom += scrollTop;
       rect.right += scrollLeft;
-    } catch (err) {}
-  } else {
-    rect = element.getBoundingClientRect();
+    }
+    else {
+      rect = element.getBoundingClientRect();
```
FezVrasta: So is this going to possibly throw error on IE11?

MrFrankel: No, I wrapped both cases in try catch, before the else was not wrapped so it throws an error in ie11.

FezVrasta: Yes that was the question, this means that in some cases this will throw an error in IE11 (that will then be caught by the `try...catch`

MrFrankel: yes, Its a bug in IE apparently which they refuse to fix. I saw that there was I try catch on the IE10 case so I figure wrapping the while thing will led to the same result.

Not sure why this arose now, it seems unrelated to my changes. 
I can only assume its related to running the suites twice.

packages/popper/src/modifiers/applyStyle.js
FezVrasta: I think this is the latest point needed to finish this PR?

MrFrankel: Ive added the case in applyStyles in a57d15560721d115993f673fec923c7d7e4db991.
Did I miss somthing?

MrFrankel: @FezVrasta Ive added the case in applyStyles in a57d155.
Did I miss something?

FezVrasta: We must export the position value in the `computeStyles` modifier to make sure that people that don't rely on `applyStyle` will not miss this piece of information.

MrFrankel: in line 59 of update.js we set the position property according to the positionFixed flag ```data.offsets.popper.position = this.options.positionFixed ? 'fixed' : 'absolute';```

which is later applied in the computeStyle modifier,

```// Styles
  const styles = {
    position: popper.position,
  };
```

So it is exposed to them in the modifier, is this not enough?

FezVrasta: I can't seem to see any edit in the `computStyle` modifier, may you add a commit on the interested line?

MrFrankel: I haven't edited the computedStyle modifier, it is passed the position via data.offsets.popper.position and through options.positionFixed.

Is that not exposed enough? or am I missing somthing

FezVrasta: Uh sorry, I didn't realize we did already handled this case, ok so that's perfect. thanjs.

MrFrankel: Yes, Id like to run in my library for a bit to see how it works in the real world before adding the docs to it.

FezVrasta:
May you please rebase on top of master so that I can merge this?

I'm going to release these changes as popper.js@next so that who is interested can play with this for a while before we release it officially

MrFrankel:
https://en.wiktionary.org/wiki/awsome

FezVrasta:
You can install it with `popper.js@next` now

PR #495, opened 2017-11-19
Fix cases of scroll issues with positionFixed enabled

MrFrankel:


packages/popper/tests/functional/core.js
```diff
@@ -115,10 +115,10 @@ const arrowSize = 5;
       prepend(reference, document.body);
       prepend(popper, document.body);
 
-      new Popper(reference, popper, {
-        onCreate(data) {
-          const bottom = getRect(popper).bottom;
-          expect(bottom).toBeApprox(getRect(reference).top);
+    new Popper(reference, popper, {
+      positionFixed: false,onCreate(data) {
```
FezVrasta: Why?

packages/popper/src/utils/getOffsetRectRelativeToArbitraryNode.js
```diff
@@ -16,6 +16,10 @@ export default function getOffsetRectRelativeToArbitraryNode(children, parent) {
   const borderTopWidth = parseFloat(styles.borderTopWidth, 10);
   const borderLeftWidth = parseFloat(styles.borderLeftWidth, 10);
 
+  if(ignoreNegativeParentScroll) {
+    parentRect.top = parentRect.top > 0 ? parentRect.top : 0;
```
FezVrasta: `Math.max(0, parentRect.top)`?

FezVrasta:
I'm not sure, but it looks like this thing is getting messy..

packages/popper/tests/functional/core.js
MrFrankel: well, it seems like an edge case that only fails on IE.

packages/popper/src/utils/getOffsetRectRelativeToArbitraryNode.js
MrFrankel: good point

MrFrankel:
Actually, considering the change and the benefit derived from having this option available, I think its rather reasonable, 4-5 special cases only effected with the flag on

packages/popper/tests/functional/core.js
FezVrasta: In this case I'd rather have something like `if (positionFixed && !isIE) { /* run test */ }` for just the specific test cases.

packages/popper/src/methods/update.js
```diff
@@ -58,7 +59,7 @@ export default function update() {
   );
   data.offsets.popper.position = this.options.positionFixed ? 'fixed' : 'absolute';
 
-  // run the modifiers
+   // run the modifiers
```
FezVrasta: Unneeeded extra space

packages/popper/src/methods/update.js
```diff
@@ -30,7 +30,8 @@ export default function update() {
     this.state,
     this.popper,
     this.reference,
-    this.options.positionFixed ? window.document.documentElement : undefined
+    this.options.positionFixed ? window.document.documentElement : undefined,
```
FezVrasta: What's the rationale behind these checks? May you add a variable named with some meaningful name and maybe also add a comment to better explain the thing?

FezVrasta: Do we need to pass this as argument if the only needed variable to evaluate this is already passed as argument as well?

packages/popper/tests/functional/core.js
MrFrankel: ok, sounds good

packages/popper/src/methods/update.js
MrFrankel: will fix

packages/popper/src/methods/update.js
MrFrankel: well, me thought is that the parameter 'fixedParent' could be used in the future for other cases where someone would want to force a parent as a reference.
If I move it inside I force a coupling between setting a specific parent and cases where you want to ignore negative offsets. 

Ill try and see if I can make this more elegant without having 2 parameters

MrFrankel:
I am withdrawing this PR in favor of a small refactor.
I have done some re writing according to some of your comments which I think will be more aligned with the approach of the rest of the code.

Ill make a new PR when i'm finished

PR #517, opened 2018-01-05
PopperLite PoC

bodia-uz:
PoC of lite version of popper lib. Without `enableEventListeners` and `disableEventListeners` logic (window scroll and resize events) and  `offset`, `keepTogether`, `arrow`, `inner`, and `hide` modifiers.

Useful for apps that don't depend on window `scroll` and `resize` events and need minimum position logic.

You can try it using package `@bodia_uz/popper.js`:
```
import * as Popper from '@bodia_uz/popper.js/dist/umd/popper-lite';
```

Regular popper still available and all tests are passed:
```
import * as Popper from '@bodia_uz/popper.js/dist/umd/popper';
```


FezVrasta:
thanks for the PR! What's the size difference?

bodia-uz:
lite vs full popper **umd** file sizes:
image

gzip full:
image

gzip lite:
image

NOTE: for gzip info used https://closure-compiler.appspot.com/home

FezVrasta:
You'll be happy to know that the `next` branch right now weights 2.68 kB gzipped 😄 (with computeStyle and applyStyle modifiers included).

The core now is way more reliable so the modifiers should be able to be much more lightweight than before.

bodia-uz:
Ou. Cool. Good job!
Thanks for that!

PR #520, opened 2018-01-12
Correction to conditional. User should only be warned modifier /is/ required.

jonathanconway:


FezVrasta:
The current behavior is correct, please provide a codepen that reproduces the bug you are trying to fix.

jonathanconway:
Thanks and apologies, I guess a PR might've been a bit hasty!

The issue is that when I construct Popper, passing [preventOverflow](https://popper.js.org/popper-documentation.html#modifiers..preventOverflow) as the only modifier, I get a warning in the console, which I don't believe is appropriate:

> `preventOverflow` modifier is required by `hide` modifier in order to work, be sure to include it before `hide`!

When I look at the conditional that wraps the warning, it doesn't make sense to me to *show* the warning if the modifier *isn't* required. I would have thought it's the other way round – warn the user if the modifier *is* required, and hasn't been included.

But perhaps I'm wrong and/or using the modifier incorrectly.

Here's a [CodePen of the 'required' warning](https://codepen.io/anon/pen/opyzZX).

FezVrasta:
The error is correct, you are disabling `preventOverflow`, while you keep all the built-in modifiers at their default settings (most of them enabled, included the `hide` modifier).

You should disable `hide` modifier as well if you want the warning to go away

jonathanconway:
Ok and thanks.

PR #527, opened 2018-01-19
Add check if the parent element has a scrollbar to determine scroll parent

donnysim:
Pretty sure `getScrollParent` tests should not be in `core` but couldn't get them to work in unit tests.

Fixes https://github.com/FezVrasta/popper.js/issues/526

### DIFF
--- a/packages/popper/src/utils/getScrollParent.js
+++ b/packages/popper/src/utils/getScrollParent.js
@@ -24,7 +24,8 @@ export default function getScrollParent(element) {
 
   // Firefox want us to check `-x` and `-y` variations as well
   const { overflow, overflowX, overflowY } = getStyleComputedProperty(element);
-  if (/(auto|scroll)/.test(overflow + overflowY + overflowX)) {
+  if (/(auto|scroll)/.test(overflow + overflowY + overflowX) &&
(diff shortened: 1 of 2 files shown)


packages/popper/src/utils/getScrollParent.js
```diff
@@ -24,7 +24,8 @@ export default function getScrollParent(element) {
 
   // Firefox want us to check `-x` and `-y` variations as well
   const { overflow, overflowX, overflowY } = getStyleComputedProperty(element);
-  if (/(auto|scroll)/.test(overflow + overflowY + overflowX)) {
+  if (/(auto|scroll)/.test(overflow + overflowY + overflowX) &&
+    (element.scrollWidth > element.clientWidth || element.scrollHeight > element.clientHeight)) {
```
FezVrasta: I'm a bit rusty, do `clientWidth` and `clientHeight` take in account possible paddings?

donnysim: Yes, it does.

donnysim: There's no need for a rush on this or if has big impact on positioning might as well close this, can get by by turning off overflow, but appreciate the help.

FezVrasta:
Unfortunately this PR breaks some tests, do you have time to look at them?

PR #529, opened 2018-01-22
Fix a bug in positionFixed where parent is transformed + some code clean up

MrFrankel:
positionFixed has been running in our production app for a while now with no issues other then this one.
Browser calculate position: fixed relative to viewport unless a parent has a transform property defined.

I have added a fix for this with tests included for it.

Also, I would like to add positionFixed to the documents in order for other to be able to make use of it, but I dont know how.


packages/popper/src/utils/isIE.js
```diff
@@ -0,0 +1,14 @@
+/**
```
FezVrasta: 1. Do we really need to check of IE11 as well? If not, can we use the existing `isIE10` function?

FezVrasta:
Documentation is automatically generated from the JSDocs comments

MrFrankel:
yes we do, apparently the transform ancestor thing is on all browsers except < ie11.

we dont actually support ie anymore, it is an absolute pain!

MrFrankel:
ok, missed the second bullet, sure, ill make IE more generic, good point

MrFrankel:
@FezVrasta made the isIE more generic, anything else you see needs addressing?

PR #530, opened 2018-01-23
Adds support for `placement: 'center'`

maxkfranz:
- Add `placement:'center'`, `placement:'center-start'`,
  `placement:'center-end'`
- Updates docs
- Updates TS definition
- Adds a basic test for `placement:'center'` without the arrow

Notes:

- There is only a basic test for `center`.  I didn't bother adding cases like scroll.  Not sure if that would be strictly necessary.
- I'm not sure whether `center-start` and `center-end` are really necessary.  I added them for consistency, but I can't think of an actual usecase.  I'm just interested in `center`.
- I also tested the `center` feature with the `cytoscape-popper` extension.  The usecase for `center` there is where you want to position some dom on top of a node.  See the screenshot for an example where I patched the demo to use `placement: 'center'`:

screen shot 2018-01-23 at 2 06 50 pm

### DIFF
--- a/packages/popper/index.d.ts
+++ b/packages/popper/index.d.ts
@@ -22,7 +22,10 @@ declare namespace Popper {
     | 'bottom-start'
     | 'left-end'
     | 'left'
-    | 'left-start';
+    | 'left-start'
+    | 'center-end'
+    | 'center'
+    | 'center-start';
 
   export type Boundary = 'scrollParent' | 'viewport' | 'window';
 
--- a/packages/popper/src/methods/placements.js
+++ b/packages/popper/src/methods/placements.js
@@ -6,6 +6,7 @@
  * - `right`
  * - `bottom`
  * - `left`
+ * - `center`
  *
  * Each placement can have a variation from this list:
  * - `-start`
@@ -45,4 +46,7 @@ export default [
   'left-end',
   'left',
   'left-start',
+  'center-end',
+  'center',
+  'center-start',
 ];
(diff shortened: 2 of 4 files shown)


FezVrasta:
Why can't you use the inner modifier?

maxkfranz:
I've tried out the inner modifier, but it doesn't seem to allow centring.

For example, these options yield the following screenshot:

```js
// ...
  placement: 'top',
  modifiers: {
    inner: {
      enabled: true
    },
    flip: {
      enabled: false
    },
    offset: {
      enabled: true,
      offset: '0px, 50%r - 50%p'
    }
  }
// ...
```

screen shot 2018-01-29 at 9 23 32 am

I would expect the options I have to give the same result as `placement: 'center'` in the PR.  Additionally, it looks like the tuple form of `offset.offset` is deprecated -- which seems necessary to control both axes.

FezVrasta:
I think I'd rather have this as a 3rd party modifier, I don't want to add more features to the core since we already have enough stability problems. Sorry.

MichaelOstrovsky:
I too think it's a missing placement. 
My take on a modifier:

```
const centerModifier = {
  name: "computeStyles",
  enabled: true,
  fn({ state }) {
    const {
      rects: {
        reference: { width: rW, height: rH, x: rX, y: rY },
        popper: { width: pW, height: pH, x: pX, y: pY }
      }
    } = state;
    state.styles.popper = {
      ...state.styles.popper,
      position: "fixed",
      left: `${rY + (rW - pW) * 0.5}px`,
      top: `${rY + (rW - pW) * 0.5}px`
    };

    return state;
  }
};
```

[Demo](https://codesandbox.io/s/popper-center-placement-7yu5z)

FezVrasta:
@MichaelOstrovsky you could publish it as an npm module, it would be useful for others.

PR #531, opened 2018-01-24
AngularJS led me to a wrapper for Angular +2

kimdah3:
This led me to a wrapper for Angular +2

### [angular-popper](https://www.npmjs.com/package/angular-popper)

This component is in alpha base, you can contribute to it!

### DIFF
--- a/MENTIONS.md
+++ b/MENTIONS.md
@@ -29,12 +29,6 @@ React wrapper around Popper.js, maintained by [Atlassian](https://www.atlassian.
 
 Angular wrapper for Popper.js
 
-## AngularJS
-
-### [angular-popper](https://www.npmjs.com/package/angular-popper)
-
-This component is in alpha base, you can contribute to it!
-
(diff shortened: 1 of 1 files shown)


FezVrasta:
I'm not familiar with the Angular nomenclature, may you simply fix the name instead of delete the whole entry?

Johann-S:
yep that's not `AngularJS` which is related to `AngularJS` v1.x it's `Angular` which is related to `Angular` v2+

kimdah3:
It could be moved to Angular and then remove the AngularJS title because I can't find a wrapper.

MrFrankel:
@kimdah3 if you are looking for an Angular wrapper I can also suggest mine.
[https://github.com/MrFrankel/ngx-popper](https://github.com/MrFrankel/ngx-popper)

MrFrankel:
@FezVrasta Mine is already mentioned 👍

PR #539, opened 2018-01-31
types: tooltip.js typing and remove namespaces/declare module from po…

cwmoo740:
It's no longer best practice to use "declare module" and "declare namespace" in .d.ts files - they can export all of their types at the top level.

Closes issue https://github.com/FezVrasta/popper.js/issues/372

### DIFF
--- a/packages/popper/index.d.ts
+++ b/packages/popper/index.d.ts
@@ -5,128 +5,132 @@
 // This file only declares the public portions of the API.
 // It should not define internal pieces such as utils or modifier details.
 
-declare namespace Popper {
-  export type Position = 'top' | 'right' | 'bottom' | 'left';
+export type Position = 'top' | 'right' | 'bottom' | 'left';
 
-  export type Placement = 'auto-start'
-    | 'auto'
-    | 'auto-end'
-    | 'top-start'
-    | 'top'
-    | 'top-end'
-    | 'right-start'
-    | 'right'
-    | 'right-end'
-    | 'bottom-end'
-    | 'bottom'
-    | 'bottom-start'
-    | 'left-end'
-    | 'left'
-    | 'left-start';
+export type Placement = 'auto-start'
+  | 'auto'
+  | 'auto-end'
+  | 'top-start'
+  | 'top'
+  | 'top-end'
+  | 'right-start'
+  | 'right'
+  | 'right-end'
+  | 'bottom-end'
+  | 'bottom'
+  | 'bottom-start'
+  | 'left-end'
+  | 'left'
+  | 'left-start';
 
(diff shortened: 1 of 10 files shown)


FezVrasta:
Thanks, we had a lot of problems with TypeScript definitions, I'd really like to have some kind of CI test to make sure them work, any idea?

cwmoo740:
I can add some typescript sample files and integrate them into your build plan - it will check that code samples of using your component API compile correctly.

Should I do that in this pull request?

FezVrasta:
Yes please! 🤗

cwmoo740:
There are a few new files:

`packages/(popper|tooltip)/tests/types/typing-test.ts` have ts code that creates some objects and calls the public api to make sure it compiles. These are run as new posttest scripts for each of those two packages.

cwmoo740:
I'm not sure the CI issues are related to my changes. I can back out any non-typescript changes to the yarn.lock if necessary.

cwmoo740:
@FezVrasta is this pr ready to merge?

FezVrasta:
I think so, I'm going to hold on on it until I get back home next week tho, I don't have enough time to dedicate to this atm

cwmoo740:
@FezVrasta any updates?

FezVrasta:
It looks good to me, I guess I'm going to merge this hoping nobody will open issues later... eheh

cwmoo740:
Ping me and I'll help you deal with them too

adidahiya:
This is a breaking change in the typings, and it should be reverted. Previously, the _default export_ was a namespace which included types as members. Now, all the types are exported from the top-level of the module _next to_ the default export. Although this is better aesthetically and probably the right direction, it is a breaking change and cannot be done in the 1.x range of this library. It caused this issue downstream: https://github.com/palantir/blueprint/issues/2226

FezVrasta:
@cwmoo740 is there any way to export the namespace and keep the other changes?

adidahiya:
Yes, I think that should be possible. I'll try to put together a quick PR.

cwmoo740:
@adidahiya there's still no convention of following semver for typing - DefinitelyTyped clearly doesn't follow semver when making breaking changes by upgrading tsc versions, for example. Exporting an additional namespace that rolls up the types again is fine, but don't revert the typings to older syntax discouraged by typescript when you should probably upgrade your library instead.

adidahiya:
This isn't DefinitelyTyped. This is an NPM-published library that is depended on by other TypeScript libraries. Downstream consumers shouldn't be able to pick up a minor release and have it break their build -- it's as simple as that.

adidahiya:
I opened #584 to address this.

PR #540, opened 2018-01-31
Find parents with CSS property overflow: overlay;

kristofferj:
Had some problems with the popups not following the scroll. Turns out we are using `overflow: overlay `. And after some digging, it seems that it is forgotten in this repo. 
(Overlay is the property of overflow that makes the scrollbar not affect the size of the element)

### DIFF
--- a/packages/popper/src/utils/getScrollParent.js
+++ b/packages/popper/src/utils/getScrollParent.js
@@ -24,7 +24,7 @@ export default function getScrollParent(element) {
 
   // Firefox want us to check `-x` and `-y` variations as well
   const { overflow, overflowX, overflowY } = getStyleComputedProperty(element);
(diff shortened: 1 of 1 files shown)


FezVrasta:
Is this even supported on modern browsers?

kristofferj:
It is supported by webkit/blink, but not recommended. 
Seems like it is deprecated too. (just found out) https://developer.mozilla.org/en-US/docs/Web/CSS/overflow
Will change to `overflow: scroll` in our app now that I know.
But maybe it is useful to include it in poppper.js for others, since it is still used out there?
Anyway. Thanks for a great library! :)

FezVrasta:
Merged since it's a small change, but better don't use `overlay` for new code

Issue #542, opened 2018-02-01
Destroy method doesn't remove all the styles


dsasko:
### CodePen demo
https://codepen.io/dsasko/pen/LQpyyr

### Steps to reproduce the problem
1. Initiate a popper instance
2. Popper applies styles to the dropdown/tooltip/whatever
3. Destroy popper instance

### What is the expected behavior?
The destroy method should remove all applied styles.

### What went wrong?
The destroy method removes all applied styles except `will-change`.

### Other comments
This seemed like a silly thing to open up a PR for...

_/packages/popper/src/methods/destroy.js - line 18_
add: `this.popper.style.willChange = '';`

FezVrasta:
Please feel free to send a PR

PR #544, opened 2018-02-02
fix: Add missing will-change cleanup for the Destroy method (#542)

dsasko:
Should I write a test for every possible style modification or the popper element?

### DIFF
--- a/packages/popper/src/methods/destroy.js
+++ b/packages/popper/src/methods/destroy.js
@@ -15,6 +15,7 @@ export default function destroy() {
     this.popper.style.left = '';
     this.popper.style.position = '';
     this.popper.style.top = '';
+    this.popper.style.willChange = '';
     this.popper.style[getSupportedPropertyName('transform')] = '';
   }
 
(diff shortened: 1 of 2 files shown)


FezVrasta:
Thanks for the PR!

Actually the whole way we handle this thing isn't optimal, we should have some kind of `destroy` method in the modifier that behaves similarly to the `onLoad` method of the existing modifiers... I guess we'll be able to do this in the future v2.

Some additional test is super welcome anyway!

FezVrasta:
Is this good to go?

dsasko:
I pushed some more changes to the destroy method. It will now clean up all variants of the sideA and sideB.
Also, I made some changes to the tests to set/check all possible style modifications. The thing that bothers me here is the transform property, I didn't include the prefixing function as it would go outside of the test-utils folder and that seemed like complication I didn't want on my head. But since this is a patchy solution anyways I guess it will suffice.

To conclude, the method passes all tests and I tested it myself to make sure everything works fine. So yes, its good to go.

PR #545, opened 2018-02-02
Fix offset parent calculation for hidden elements

### LINKED ISSUE #537: Incorrect offset container used when first child of common parent is a hidden input
### CodePen demo

https://codepen.io/anon/pen/BYyjNM

1.  Include a hidden input as the first child of the offset container of the elements.

The "reference" and "popper" elements are aligned.

### What went wrong?

A hidden input's `offsetParent` is null, causing incorrect offset container to be chosen for the offsets calculations.

### Any other comments?

[This is the place](https://github.com/FezVrasta/popper.js/blob/150ef056629b171b96e5f2bcf6316d25ff0a146f/packages/popper/src/utils/isOffsetContainer.js#L9) where element's first child is arbitrarily chosen for a test if the element is an offset container. I think, it needs to use a child which is actually rendered.

RKushnir:
Fixes #537

### DIFF
--- a/packages/popper/src/utils/getOffsetParent.js
+++ b/packages/popper/src/utils/getOffsetParent.js
@@ -9,6 +9,7 @@ import getStyleComputedProperty from './getStyleComputedProperty';
 export default function getOffsetParent(element) {
   // NOTE: 1 DOM access here
   let offsetParent = element && element.offsetParent;
+  // Skip hidden elements which don't have an offsetParent
   while (offsetParent === null && element.nextElementSibling) {
     offsetParent = (element = element.nextElementSibling).offsetParent;
   }
--- a/packages/popper/src/utils/getOffsetParent.js
+++ b/packages/popper/src/utils/getOffsetParent.js
@@ -1,4 +1,5 @@
 import getStyleComputedProperty from './getStyleComputedProperty';
+import isIE from './isIE';
 /**
  * Returns the offset parent of the given element
  * @method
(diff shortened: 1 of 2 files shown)


packages/popper/src/utils/getOffsetParent.js
```diff
@@ -8,7 +8,10 @@ import getStyleComputedProperty from './getStyleComputedProperty';
  */
 export default function getOffsetParent(element) {
   // NOTE: 1 DOM access here
-  const offsetParent = element && element.offsetParent;
+  let offsetParent = element && element.offsetParent;
+  while (offsetParent === null && element.nextElementSibling) {
```
FezVrasta: Great! May you just add a small comment to explain why this is needed, for future reference?

packages/popper/tests/unit/getOffsetParent.js
```diff
@@ -24,4 +24,23 @@ describe('utils/getOffsetParent', () => {
...
+
+  it('first child element is hidden', () => {
+    const innerNode = document.createElement('div');
+    innerNode.style.display = 'none';
+    const nextSibling = document.createElement('div');
+    node.style.position = 'absolute';
+    node.appendChild(innerNode);
+    node.appendChild(nextSibling);
+
+    expect(getOffsetParent(innerNode)).to.equal(node);
```
FezVrasta: I see IE10 is failing here, I wonder if it just defines `offsetParent` even to hidden elements.... In that case I think the assertion should check to make sure the returned node is the first with a `offsetParent` property set.

RKushnir: If it defined `offsetParent`, wouldn't the assertion still be correct? It seems to me, that would just behave like the previous test case('element is inside positioned element').

RKushnir: So I figured the `offsetParent` for hidden elements in IE10 is `body`.

FezVrasta: So for IE10 we should have something like:

RKushnir: @FezVrasta A kindly reminder in case you forgot 🙂 I'd like to finish this while it's still fresh.

packages/popper/src/utils/getOffsetParent.js
```diff
@@ -7,8 +8,19 @@ import getStyleComputedProperty from './getStyleComputedProperty';
...
-  const offsetParent = element && element.offsetParent;
+  let offsetParent = element.offsetParent;
+  // Skip hidden elements which don't have an offsetParent
+  while (offsetParent === noOffsetParent && element.nextElementSibling) {
+    offsetParent = (element = element.nextElementSibling).offsetParent;
+  }
+
   const nodeName = offsetParent && offsetParent.nodeName;
 
   if (!nodeName || nodeName === 'BODY' || nodeName === 'HTML') {
```
FezVrasta: Just to make CodeClimate happy, may you refactor the branch below to be:

RKushnir: CodeClimate is green and delighted now.

PR #552, opened 2018-02-12
feat(tooltip): Added function to update the tooltip's content

victor-cordova:
Tests updated, but not sure if the typescript definition should be as well, since this is a tooltip exclusive change. ❓Please let me know  
I'm reusing some code to insert the new String|HTMLElement

DEMO:
video

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,8 @@
+checks:
+  file-lines:
+    enabled: false
+  method-complexity:
+    enabled: false
 engines:
   duplication:
     enabled: false
--- a/docs/_includes/tooltip-documentation.md
+++ b/docs/_includes/tooltip-documentation.md
@@ -10,6 +10,7 @@
         * [.hide()](#Tooltip+hide)
         * [.dispose()](#Tooltip+dispose)
         * [.toggle()](#Tooltip+toggle)
+        * [.updateTitleContent()](#Tooltip+updateTitleContent)
     * _static_
         * [.PlacementFunction](#Tooltip.PlacementFunction) ⇒ <code>String</code>
         * [.TitleFunction](#Tooltip.TitleFunction) ⇒ <code>String</code>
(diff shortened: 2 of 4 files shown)


FezVrasta:
Thanks.

I'm not particularly liking the fact that we now have a second class that needs to be initialized.. I'd rather have few functions for a more functional approach.

victor-cordova:
@FezVrasta I totally agree with you. This was split mainly because of code climate's validation. I have now deleted the utils and tried to add everything in the index.js file with only one blocker.
First of all, I believe to add this enhancement there are 2 steps needed (let me know if you think otherwise).
1. Clean the previous innerHTML or innerText of the tooltip node
2. Add the new content with the same logic as the ```_create``` method (taking into account the innerText vs textContent consideration pointed out by @hekigan in [Issue](https://github.com/FezVrasta/popper.js/issues/342)

In order to do the first one, I've added some lines of codes that get the max lines of code pretty close to the limit. However, for some reason, the second step (located in the ```_addTitleContent``` method) violates the cognitive complexity validation (**even though the _create method has the exact same logic**). This is currently my only blocker.

@FezVrasta I would greatly appreciate any leads on why the cognitive complexity of this new method is increased. I've had a quick look at [CognitiveComplexity](https://www.sonarsource.com/docs/CognitiveComplexity.pdf) but I haven't found any leads.

Cheers

FezVrasta:
Don't worry about codeclimate, if it makes sense as is we can just add an exception

victor-cordova:
Excelent 👍 . I went ahead and disabled the validations to the codeclimate.yml (let me know if this should be done elsewhere) and also updated the docs for tooltip. This should now be ready. As a recap, these are the changes:
1. Refactored all references inside ```tooltip.js``` and ```tests\functional\tooltip.js``` from innerText to textContent (all tests pass).
2. Added a new private method called ```_addTitleContent``` used by ```_create``` and ```_updateTitleContent``` that has all the titleNode insertion logic (innerHTML and textContent).
3. Added ```_updateTitleContent``` and ```_clearTitleContent``` for the updateTitle logic.
4. Updated ```tooltip-documentation.md``` to point out that ```textContent``` is now being used and to add the new ```updateTitleContent``` method

Let me know if all of these are ok, otherwise I'll revert/update as needed

PR #556, opened 2018-02-13
Fix javascript error "cannot find method removeChild of undefined" up…

takinosaji:
Fix javascript error "cannot find method removeChild of undefined" upon disposing  when popper or tooltip parent node doesn't exist anymore. Typical for angular applications.

### DIFF
--- a/packages/popper/src/methods/destroy.js
+++ b/packages/popper/src/methods/destroy.js
@@ -22,7 +22,7 @@ export default function destroy() {
 
   // remove the popper if user explicity asked for the deletion on destroy
   // do not use `remove` because IE11 doesn't support it
-  if (this.options.removeOnDestroy) {
+  if (this.options.removeOnDestroy && this.popper.parentNode) {
     this.popper.parentNode.removeChild(this.popper);
   }
   return this;
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -271,7 +271,7 @@ export default class Tooltip {
       this.popperInstance.destroy();
 
       // destroy tooltipNode if removeOnDestroy is not set, as popperInstance.destroy() already removes the element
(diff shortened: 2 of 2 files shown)


takinosaji:
bump

takinosaji:
guys, bump

FezVrasta:
I don't think I understand the reasoning behind this. Seems like you are removing the popper HTMLElement from the DOM using something different from the built-in Popper.js methods and you then expect Popper.js to simply ignore this, right?

Why?

takinosaji:
You dont understand why you should perform checks on null before invoking methods on the objects? Really?

We are using kendogrid in our SPA project, ok? It has paging, sorting, dynamic data load. So imagine that you are attaching tooltip to something inside kendo template. Once you do this tooltip is working fine until you have datasource updated and projected to dom. The issues is that once it is happened tooltip.js doesnt appear any more. To overcome this we are saving created tooltips and once grid loaded before attaching new ones we are disposing old ones and it is exact time and place where we have an issue with your framework.

Still doesn't sound reasonable for you?

FezVrasta:
First of all, calm your tone, I'm not your friend and you are using my work **for free*. Clear?

Second, you should handle the destroy/unbind of your tooltips before you let any 3rd party code destroy the HTML elements, otherwise you'll likely introduce memory leaks.

takinosaji:
@FezVrasta second of all i am not your friend either, so watch your tone as well. You have people who are interesting in your opensource library who spending time and assets to debug, fix problem and contribute to your project, and it is not a reason for saying things like "oh why are you doing smth like this, not our problem good bye"?

To the point. I agree completely about necessity of preventing memory leaks and about handling of destroy beforehand but it is not sometimes possible even in green and nice written projects, you may imagine how things are in old bullshit legacy code.

PR #559, opened 2018-02-16
using the clip-technique to hide the tooltip

pixelatorz:
Screenreaders can not read content thats hidden with display: none, therefor I changed the show and hide functions in tooltip.js using the clipping technique or the visuallyhidden class (also used in bootstrap).

more info can be found here: https://www.w3.org/WAI/tutorials/fundamentals/hiding/


FezVrasta:
Do we actually want the screen reader to be able to read an hidden tooltip?

pixelatorz:
We checked up with our accessibility excerpt and we can decline this pullrequest

PR #563, opened 2018-02-26
Fix javascript error "cannot find method removeChild of undefined" up…

takinosaji:
@FezVrasta Well. Dont get me wrong. No eagerness to offend anybody, frankly speaking i am surprised that you found my post offensive. Still consider this change useful, and since you didn't even read end of my post because of emotions i will create pr one more time and try to explain. 

I agree completely about necessity of preventing memory leaks and about handling of destroy beforehand but sometimes handling of disposing of parent nodes is not possible even in green and nice written projects, you may imagine how things are in old bullshit legacy code. Third party libraries are the problem, what is we dont have place to intercept datasource change, or reload triggered?

Having this check allows us to ensure that any tooltip we dont need is disposed.

### DIFF
--- a/packages/popper/src/methods/destroy.js
+++ b/packages/popper/src/methods/destroy.js
@@ -22,7 +22,7 @@ export default function destroy() {
 
   // remove the popper if user explicity asked for the deletion on destroy
   // do not use `remove` because IE11 doesn't support it
-  if (this.options.removeOnDestroy) {
+  if (this.options.removeOnDestroy && this.popper.parentNode) {
     this.popper.parentNode.removeChild(this.popper);
   }
   return this;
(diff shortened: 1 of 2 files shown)


FezVrasta:
I don't want to introduce possible memory leaks in my code, and I don't think that a library simple as Tooltip.js should handle special edge cases like this.

For Popper.js, you can disable the "auto destroy" feature (off by default) and that's all you need.

takinosaji:
It doesn't cause ANY memory leak. You call this "parentNode" but it is actually boundNode. In  terms of dom this situation is obvious and happens, doesn't violate any principle.

Can you tell what memory leak do you mean if i dont get it?

FezVrasta:
If the `HTMLElement` used as popper (or tooltip) is destroyed by some 3rd party code, the event handlers will stay attached to the DOM in some browsers, or in the worst case, some other internal Popper.js routine will try to call some method on this destroyed `HTMLElement` and will throw an error.

PR #570, opened 2018-03-02
Fix the statement 2 of the loop to run through all autoprefixer

raffo1234:
Fixes https://github.com/FezVrasta/popper.js/issues/568

- Fix the statement 2 of the loop to run through all autoprefixers.


FezVrasta:
Thanks!

PR #573, opened 2018-03-04
TypeScript definitions: extract Behavior type

denisname:
I extracted the _behaviors_ in their own type.
So they can be used by other libraries, especially Bootstrap v4 TypeScript definition.

@giladgray

Edit: Travis CI fail is not related to this...


denisname:
@MrFrankel, @FezVrasta, @cwmoo740 
Any complaints? Also any chances to have this, and other changes to `index.d.ts`, published on npm soon?
Thanks.

cwmoo740:
this looks fine to me. the test failures are definitely unrelated.

PR #584, opened 2018-03-13
Fix regression in Popper namespace typings

adidahiya:
Fixes the breaking change in https://github.com/FezVrasta/popper.js/pull/539#issuecomment-372728264

### DIFF
--- a/packages/popper/index.d.ts
+++ b/packages/popper/index.d.ts
@@ -1,119 +1,140 @@
-// Type definitions for popper.js 1.10
-// Project: https://github.com/FezVrasta/popper.js/
-// Definitions by: edcarroll <https://github.com/edcarroll>, ggray <https://github.com/giladgray>, rhysd <https://rhysd.github.io>, joscha <https://github.com/joscha>, seckardt <https://github.com/seckardt>, marcfallows <https://github.com/marcfallows>
+/**
+ * @fileoverview This file only declares the public portions of the API.
+ * It should not define internal pieces such as utils or modifier details.
+ *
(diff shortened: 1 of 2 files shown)


FezVrasta:
Isn't there any way to reduce the amount of changes?

adidahiya:
Not really, no. I've left code comments explaining how this can be simplified if you choose to make a breaking change in 2.x

FezVrasta:
May we do the opposite of what you are doing here?

Have all the top level exports, and at the bottom the namespace that re-exports the top-level stuff as props of the namespace?


```
export type xxx = xxx
export type yyy = yyy

export namespace Popper {
  export type xxx
  export type yyy
}
```

adidahiya:
Unfortunately that doesn't work.

```ts
export namespace Popper {
  export type Position; // this isn't valid syntax. you need an assignment. '=' expected by tsc

  export type Position = Position; // this doesn't work because Position becomes a circular reference.
}
```

FezVrasta:
Ok thanks. We should have some kind of types tests, may you update them to add a regression test for this thing?

adidahiya:
Yep, I've updated the tests to verify that the regression is fixed.

FezVrasta:
Ok thanks, waiting for feedbacks from @cwmoo740 before I merge this

cwmoo740:
Sure, it keeps all the top level exports. I'll revisit this and update it again when it's ready for a 2.0 release. I also put a PR into 

https://github.com/palantir/blueprint/pull/2227

to reference the types this way.

Versioning types on a non-typescript project is still not a solved issue, and especially when supporting newer features in ts, type definition files shouldn't necessarily be held back by downstream users. This is one of the reasons TSC has the paths mapping feature, to allow you to hold onto old type definition files in existing projects.

Personally, if it's not causing too many issues for users, I would prefer that you merge my pr into blueprint instead, but if that's not possible, this is fine.

FezVrasta:
Released as https://github.com/FezVrasta/popper.js/releases/tag/v1.14.1

PR #587, opened 2018-03-24
html.topScroll -> body.topScroll  to support safari top scroll issue

shamit6:
Hi, 
I faced with bug when i used Safari. I saw the popper's vertical boundaries was between 0px to 100vh.
I found that in Safari document.documentElement.scrollTop = 0, so i replaced it with the body at getScroll function.
The bug also happened in any preventOverflow.boundariesElement when the target element is lower the 
then the viewport.

here is an example I wrote (:
https://codepen.io/anon/pen/EEvELV


FezVrasta:
Thanks, unfortunately it seems to break IE support, may you look into it? And maybe add a test to cover your specific case?

PR #593, opened 2018-03-29
add typecript types to npm build

epicallan:

### DIFF
--- a/packages/tooltip/.npmignore
+++ b/packages/tooltip/.npmignore
@@ -5,3 +5,4 @@
 # Except the dist files
 #======================
 !dist/**/*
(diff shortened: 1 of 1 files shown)


JeroenVinke:
@FezVrasta I might be wrong, but I believe the index.d.ts file is missing for the tooltip.js package. https://unpkg.com/tooltip.js@1.2.0/index.d.ts does not exist while https://unpkg.com/tooltip.js@1.2.0/package.json does reference it. Could you reevaluate this PR? Let me know if there is anything I can do to help

FezVrasta:
Right, I mixed up things :-)

PR #594, opened 2018-03-29
Make isIE minifier-friendly

atomiks:
Noticed a semi-substantial increase in file size due to the new util...

A closure can be also used instead of the external "cache" object

Old: 472 bytes
New: 132 bytes

### DIFF
--- a/packages/popper/src/utils/isIE.js
+++ b/packages/popper/src/utils/isIE.js
@@ -1,30 +1,19 @@
 /**
- * Tells if you are running Internet Explorer
+ * Determines if the browser is Internet Explorer
  * @method
  * @memberof Popper.Utils
- * @argument {number} version to check
+ * @param {number} version to check
  * @returns {Boolean} isIE
  */
-const cache = {};
+export default (() => {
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thanks!

atomiks:
Managed to get it down to 132 bytes. I noticed the tests didn't fail (running `yarn test`) due to that though?

FezVrasta:
Tests for IE are often disabled because the browser is too unstable to reliably run tests..

atomiks:
You could do a mock with a UA string of IE11 and IE10

FezVrasta:
Feel free to add a test for it if you have time

atomiks:
@FezVrasta does this fail in Node environments without `navigator` global (since it's executed immediately)? Is there any test for that? I think the `isBrowser` var needs to be imported here...

atomiks:
@FezVrasta did you see the above comment?

FezVrasta:
I don't think there are tests for it. Send a PR if you notice any problem please.

Issue #551, opened 2018-02-09
Pixel-perfect positioning: don't floor positions with gpuAcceleration disabled


atomiks:
Currently, you floor the positioning because of blurriness with fractional translation, but this inevitably causes small positioning errors because elements on the page can be positioned using subpixels.

I'm thinking of not flooring when `gpuAcceleration` is disabled. When used with `gpuAcceleration` disabled, you achieve no blurriness (due to the lack of a new layer the popper is promoted to with `translate3d`?) and pixel-perfect positioning at a minor performance cost. 

### Any other comments?

See: #277

FezVrasta:
Fixed

PR #595, opened 2018-03-30
Fix rounding

atomiks:
As mentioned in #551, this will allow for "pixel-perfect" positioning if desired, since subpixel `top/left` positioning doesn't cause blurriness unlike `translate3d`.

### DIFF
--- a/packages/popper/src/modifiers/computeStyle.js
+++ b/packages/popper/src/modifiers/computeStyle.js
@@ -37,7 +37,9 @@ export default function computeStyle(data, options) {
     position: popper.position,
   };
 
-  // round sides to avoid blurry text
+  // Avoid blurry text by using full pixel integers.
+  // For pixel-perfect positioning, top/bottom prefers rounded
+  // values, while left/right prefers floored values.
   const offsets = {
     left: Math.floor(popper.left),
     top: Math.round(popper.top),
--- a/packages/popper/src/modifiers/computeStyle.js
+++ b/packages/popper/src/modifiers/computeStyle.js
@@ -37,14 +37,12 @@ export default function computeStyle(data, options) {
     position: popper.position,
   };
 
-  // floor sides to avoid blurry text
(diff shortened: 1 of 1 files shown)


atomiks:
After doing some tests I think I have a decent solution to ensure precision when using `gpuAcceleration` too.

Allow people to specify a `round()` callback function which gets supplied the current `placement` and uses the return value. Browsers will imperfectly render the position relative to the reference sometimes, no matter what, so we need to manually decide which rounding to choose based on the placement (and maybe other additional checks like which browser/rendering engine is currently being used), and also the specific individual instance.

Something like this:

```js
new Popper(reference, popper, {
  modifiers: {
    computeStyle: {
      round: placement => Math.floor
    }
  }
})

// In source:
const floor = gpuAcceleration 
  ? find(
    data.instance.modifiers, 
    modifier => modifier.name === 'computeStyle'
  ).round(data.placement) 
  : v => v;
```

atomiks:
It seems like top/bottom prefers `round` while left/right prefers `floor`. See:

round top/bottom, floor left/right:

rounding
rounding

all floor (sits too high vertically)

rounding

all round (sits too far horizontally)

rounding

If I change the width from 100x100 px to any combination of even/odd the floor + round combo seems to work every time. I also tried in FF and Safari with the same results.

FezVrasta:
Oh, that's what I wanted to see! I knew there was a magic recipe! Thanks!

packages/popper/src/modifiers/computeStyle.js
```diff
@@ -37,11 +37,11 @@ export default function computeStyle(data, options) {
     position: popper.position,
   };
 
-  // floor sides to avoid blurry text
+  // round sides to avoid blurry text
```
FezVrasta: Care to improve this comment to explain the reasoning behind floor vs round?

FezVrasta:
Great thanks, we just need tests to pass now

PR #596, opened 2018-03-30
fix #251 (finally)

### LINKED ISSUE #251: Popper does not update its position to fit entirely in viewport/document if it is overflowing

atomiks:
#251 has been left open too long I think, and this is the solution that @kybishop suggested, and the one I use personally, but it should be placed in `update()` so it works for resizing too.

Yes, it accesses the DOM and will reduce performance, but I don't think there's another way.  Please test the performance difference before/after to see.

`Object.assign()` is being used, which I believe is already being ponyfilled in the lib.

### DIFF
--- a/packages/popper/src/methods/update.js
+++ b/packages/popper/src/methods/update.js
@@ -29,11 +29,10 @@ export default function update() {
   // NOTE: DOM access here
   // resets the popper's position so that the document size can be calculated excluding
   // the size of the popper element itself
-  Object.assign(this.popper.style, {
-    top: '',
-    left: '',
-    [getSupportedPropertyName('transform')]: '',
-  });
+  const popperStyles = this.popper.style;
+  popperStyles.top = '';
+  popperStyles.left = '';
+  popperStyles[getSupportedPropertyName('transform')] = '';
 
   // compute reference element offsets
   data.offsets.reference = getReferenceOffsets(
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thanks! I'm going to take a look at this soon. How's the end result? No weird visual glitches or jumps?

atomiks:
Looks like Object.assign isn't being transformed, probably need to add that babel plugin...

Test it here: https://codepen.io/anon/pen/Yavaav?editors=1010

FezVrasta:
I see that some tests are failing even on Chrome and Firefox, which do support Object.assign natively.

Anyway maybe we better get rid of it and use basic assignment?

atomiks:
Updated it.

I also updated the CodePen so you can scroll and see how the updates work there as well. (seems to be fine). I'm going to test IE11 in a second.

atomiks:
✖ should update the tooltip position when changing the title
        IE 10.0.0 (Windows 8 0.0.0)
        IE 11.0.0 (Windows 10 0.0.0)

Looks like this test is failing for some reason...

atomiks:
Did you measure the perf impact of it yet?

You mentioned that 1.14.0 updates in 0.3ms instead of 1ms, what is it with this change?

FezVrasta:
It still looks smooth
image

PR #597, opened 2018-04-03
Patch pull #594

atomiks:
Patches pull #594 so errors won't be thrown in Node envs.


atomiks:
According to [this](https://stackoverflow.com/questions/17907445/how-to-detect-ie11) the original detection for IE11 is also true for IE10, so I will try to fix it.

PR #607, opened 2018-04-10
feat: Get elements ownerDocument for correct measurements

ryanseddon:
After some digging I found that when `getWindowSizes` is called it looks at the parent document which get incorrect values when using this inside of an iframe.

I created a test case here: https://codesandbox.io/s/8n31m5132j

First one is in an iframe and always puts it at the bottom even though it doesn't have enough room.

The second one shows the same thing but not wrapped in a iframe. This correctly moved the popper to the top of the element and upon scrolling moves it to the bottom.


FezVrasta:
Thanks!

ryanseddon:
Will you push out a react-popper update soon with this change?

PR #612, opened 2018-04-11
fix(computeStyle): correctly calculate `bottom` when x="top" and `rig…

### LINKED ISSUE #585: Incorrect `bottom` parameter when popper element has no parent (but <html>)
Hey Hey,
I stumbled upon this issue the other day, where `computeStyle` doesn't pick the right `bottom` position for my popper. I've illustrated the bug in [this CodePen demo](https://codepen.io/marcohamersma/pen/XZwLwr).

screen shot 2018-03-13 at 18 06 35

### CodePen demo
https://codepen.io/marcohamersma/pen/XZwLwr

1. Make the body stretch beyond the viewport (like make it 2000px high)
2. Create a reference inside an element with position `fixed`
3. Create a popper outside of this element
4. Open a popper with `placement:'top'` and `computeStyle.x` to 'top'

The popper appears in the expected place (above the reference element)

### What went wrong?
The popper is not visible because it's off screen

### Any other comments?
I know that it's maybe not super conventional to have the popper inside a different container than the reference element, but parts of my application require that I attach the overlay to the body. If there is no easy solution for this, perhaps a way to manually define what the `offsetParent` would be?

bimbiltu:
…ht` when y="left"

Fixes #585

### DIFF
--- a/packages/popper/src/modifiers/computeStyle.js
+++ b/packages/popper/src/modifiers/computeStyle.js
@@ -66,12 +66,22 @@ export default function computeStyle(data, options) {
   // its bottom.
   let left, top;
   if (sideA === 'bottom') {
-    top = -offsetParentRect.height + offsets.bottom;
+    // when offsetParent is <html> the positioning is relative to the bottom of the screen (excluding the scrollbar)
+    // and not the bottom of the html element
+    if (offsetParent.nodeName === 'HTML') {
+      top = -offsetParent.clientHeight + offsets.bottom;
+    } else {
+      top = -offsetParentRect.height + offsets.bottom;
+    }
   } else {
     top = offsets.top;
   }
   if (sideB === 'right') {
-    left = -offsetParentRect.width + offsets.right;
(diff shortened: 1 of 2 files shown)


bimbiltu:
When testing this I noticed that the reproducer in #585 had a couple of extras in it that werent causing the bug. I found this bug appears when:
 - `x` is set to `'top'` or `y` is set to `'left'` in computeStyle
 - the poppers offsetParent is the documentElement (did not matter where the reference was)
 - the page has enough content on it so that it scrolls

bimbiltu:
looks like the tests are failing in IE10, ill try to take a look at that later today

bimbiltu:
Fixed the unit tests, everything seems good now

FezVrasta:
thanks!

bimbiltu:
Hey @FezVrasta any idea when this change will be released? Id love to start using it in my project

FezVrasta:
As soon I have some free time I'll publish it, sorry for the delay

Issue #518, opened 2018-01-07
Confusing documentation with boundaries (how to ensure flip with floating navbar?)


niftylettuce:
I have a `.navbar-fixed-top` (using Bootstrap 3 and implementing a custom Popper instance).

Trying to figure out how to make it so when you scroll down the page, a popover that begins to be hidden underneath a fixed top navbar will get flipped automatically.

I tried passing a `modifiers.flip.boundariesElement` option, but that didn't seem to work.

Is there any way to do a simple "reduce viewport top size by X pixels?" where X is the height of the floating navbar?

FezVrasta:
You should define an offset.

niftylettuce:
@FezVrasta how do I do that btw?  It says its an Object, but is it `offset.top`, etc?  I couldn't find it anywhere in docs and started to search in source code..

mwebb33:
I am also confused by the offset implementation/documentation. Sorry, I tried. I got frustrated.

mwebb33:
I basically want to set a pixel offset from the top of the viewport and the left of the viewport.

beowulfenator:
Same problem here. @FezVrasta, can you please be more specific about there to define the offset? Thanks!

FezVrasta:
So, the issue I was referring to is #9

FezVrasta:
Sorry I thought Popper.js already supported it but it seems like it doesn't.

Ideally we should have an option for `preventOverflow` and `flip` that allows to set offsets on the 4 sides, it shouldn't be difficult to implement.

As workaround you may have your layout structured like this:

```
<body>
  <header />
  <main />
</body>
```

One below the other, so that you can set `main` as `boundariesElement`.

Sorry for the confusion.

BrianBu01:
A real simple way to fix this is to change lines 72-75 in getBoundaries.js lines to:

  boundaries.left += padding.left || padding;
  boundaries.top += padding.top || padding;
  boundaries.right -= padding.right || padding;
  boundaries.bottom -= padding.bottom || padding;

I can push the changes to a branch, if you want to give me limited permissions.

FezVrasta:
@BrianBu01 feel free to fork this repo, do you changes and send a pull request. Thanks! 🤩

beowulfenator:
@BrianBu01 This didn't help in my case. Am I doing something wrong? How does this update work?

BrianBu01:
For me after the change my code using Popper would be:

var popper = new Popper(
            target, 
[...]

PR #615, opened 2018-04-12
feat: Add specific padding to boundaries (#518)

BrianBu01:
There were no tests for this class already, but all existing tests still pass.

### DIFF
--- a/packages/popper/src/utils/getBoundaries.js
+++ b/packages/popper/src/utils/getBoundaries.js
@@ -69,10 +69,12 @@ export default function getBoundaries(
   }
 
   // Add paddings
-  boundaries.left += padding;
-  boundaries.top += padding;
-  boundaries.right -= padding;
-  boundaries.bottom -= padding;
+  padding = padding || 0;
+  const isPaddingNumber = typeof padding === 'number';
+  boundaries.left += isPaddingNumber ? padding : padding.left || 0; 
+  boundaries.top += isPaddingNumber ? padding : padding.top || 0; 
+  boundaries.right -= isPaddingNumber ? padding : padding.right || 0; 
+  boundaries.bottom -= isPaddingNumber ? padding : padding.bottom || 0; 
 
   return boundaries;
 }
--- a/packages/popper/src/utils/getWindowSizes.js
+++ b/packages/popper/src/utils/getWindowSizes.js
@@ -8,10 +8,10 @@ function getSize(axis, body, html, computedStyle) {
     html[`offset${axis}`],
     html[`scroll${axis}`],
     isIE(10)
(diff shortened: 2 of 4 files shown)


packages/popper/src/utils/getBoundaries.js
```diff
@@ -69,10 +69,10 @@ export default function getBoundaries(
...
 
   // Add paddings
-  boundaries.left += padding;
-  boundaries.top += padding;
-  boundaries.right -= padding;
-  boundaries.bottom -= padding;
+  boundaries.left += padding.left || padding;
+  boundaries.top += padding.top || padding;
+  boundaries.right -= padding.right || padding;
+  boundaries.bottom -= padding.bottom || padding;
```
FezVrasta: We should probably fallback to `0` in case one of the sides is not defined (eg: `{top: 10, left: 20 }`

BrianBu01: Ok got it.

BrianBu01: This should be updated now with the last squash+push

FezVrasta:
May you add few tests to cover the new cases please?

BrianBu01:
@FezVrasta After I get these tests done, how long would it be until this is in a version?  I have a week until code freeze, so need to either fork this and go off that, or need to have these changes in.

FezVrasta:
As soon this is merged I'll release the new version with this feature in.

BrianBu01:
OK I am working on a few other higher priority issues, I'll get some tests in as soon as I can.

BrianBu01:
@FezVrasta Tests are added and passing.  Some Tests are not passing for IE10 and IE11, but those tests were a backfill for missing tests for existing functionality.

BrianBu01:
@FezVrasta This is ready for your review.

FezVrasta:
I don't get why the tests are not passing on IE...this isn't any browser specific functionality

BrianBu01:
@FezVrasta I'm not sure either, but those tests were just backfill only related due to proximity.  Can we still get these changes is now, and one of us can take a closer look when we have more time?  I have to pull this into my current project today, and I'll have to go off my clone if it isn't in yours.

FezVrasta:
I'd rather wait until we have a solid pr sorry

BrianBu01:
I tried debugging the tests on another PR, as I am not sure how to get the IE tests to run locally even under a VM.  

Pretty much I think IE is doing something funny in getBoundingClientRect.

BrianBu01:
@FezVrasta Well there you go.

lixiaoyan:
@FezVrasta Could you please publish a new version for this? Thanks!

PR #652, opened 2018-07-20
fix broken table in tooltip-documentation.md

audionerd:


FezVrasta:
This markdown is automatically generated from the code comments, we can't change it that way

PR #655, opened 2018-07-27
fix: add missing props of Data to typescript declarations

inomdzhon:
* Add Data.arrowStyles
* Add Data.attributes


packages/popper/tests/types/type-tests.ts
```diff
@@ -50,6 +50,8 @@ const options: Popper.PopperOptions = {
     console.log(data);
   },
   onUpdate(data) {
+    console.log('x-out-of-boundaries', data.attributes);
```
FezVrasta: Is this used to test the type?

inomdzhon: Yeap

PR #656, opened 2018-07-27
flowtype: add event listener methods to Instance

jquense:


jquense:
The typescript types are already correct

PR #658, opened 2018-07-28
docs: table was not rendered properly

illiteratewriter:
the last few rows of the `new Tooltip` was not rendered as a table. fixed it.


FezVrasta:
#652

PR #659, opened 2018-07-28
fix #653 Skip elements that does not have `.offsetParent` correctly

### LINKED ISSUE #653: Position breaks when the reference has an SVG element directly inside
### CodePen demo

https://codepen.io/anon/pen/djNawe

(remove the SVG element to see the difference)
(if you wrap the SVG in a `div`, it also works)

1.  Put the popover inside its reference;
2.  Make the reference absolutely positioned and give it some `top` and `left` values;
3.  Put an SVG element inside the reference.

Removing the SVG element makes it work as expected:
image

### What went wrong?

When it has an SVG element, it breaks:

screen shot 2018-07-21 at 21 07 14

### Any other comments?

I'm trying to integrate Popper.js on my component library (https://reakit.io).

piglovesyou:
Some element (eg. svg) does not have `.offsetParent` so we can try the
nextElementSibling of it

Closes #653

### DIFF
--- a/packages/popper/src/utils/getOffsetParent.js
+++ b/packages/popper/src/utils/getOffsetParent.js
@@ -15,7 +15,7 @@ export default function getOffsetParent(element) {
   const noOffsetParent = isIE(10) ? document.body : null;
 
   // NOTE: 1 DOM access here
-  let offsetParent = element.offsetParent;
+  let offsetParent = element.offsetParent || null;
   // Skip hidden elements which don't have an offsetParent
   while (offsetParent === noOffsetParent && element.nextElementSibling) {
     offsetParent = (element = element.nextElementSibling).offsetParent;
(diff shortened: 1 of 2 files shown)


piglovesyou:
Boof

piglovesyou:
Ready to be reviewed.

packages/popper/src/utils/getOffsetParent.js
```diff
@@ -15,7 +15,7 @@ export default function getOffsetParent(element) {
   const noOffsetParent = isIE(10) ? document.body : null;
 
   // NOTE: 1 DOM access here
-  let offsetParent = element.offsetParent;
+  let offsetParent = element.offsetParent || null;
```
glebm: If `element.offsetParent` is 0, the `offsetParent` variable will be `null`. Is this the intent?

FezVrasta: offsetParent is not a number, is an Element

glebm: Ah, thanks

PR #661, opened 2018-07-28
Feat: options.closeOnClickOutside to close tooltip on click outside

### LINKED ISSUE #300: Add "click outside" option to hide tooltip
### Suggestion: 

I've made a plugin for Tether (back in the day) for tooltips and I pass an option (Boolean) 
for tooltips (only ones which are opening on "click/tap/manual" event) and it listens
to clicking outside of the tooltip area and then closes them.

I guess it will look like this:

````
var tooltip = new Tooltip(referenceElement, {
    title: "Hey there",
    trigger: "click",
    options: {
       clickOutside : true
   }
});
````

Here's a snippet of my code which will give some idea to what I did (modified for simplicity purposes):

````
if( checkClickOutside ){
    timer = setTimeout(function(){
        // bind a "click outside" event on the document itself
        // (this event MUST be removed when the tooltip is
        // removed to prevent it from being created again and
        // not to pollute the document with events)
        $(document.body).off('click.tooltip_outside')
                        .on('click.tooltip_outside', onClickOutside.bind(that));
    }, 50);
}

function onClickOutside(e){
    if( !$(e.target).closest('.ttip').length ){
        $(document).trigger('clickOutside', [e.target]);
        $(document.body).off('click.tooltip_outside');
        this.hide();
    }
}
````

I've been using Tether for YEARS and only today I've found out about your lib. great work!
I was always sad how Tether got so shitty and abandoned... this seems so much better in terms of code.
Tether's source was so horrible..OMG.. super bad.

---------

BTW - would be nice if the [tooltip example page](https://popper.js.org/tooltip-examples.html) would show an example where some options are passed

piglovesyou:
Fixes #300


piglovesyou:
I used `mousedown` of capture event phase, after the fashion of [Google Closure Library](https://github.com/google/closure-library/blob/08b4f5c1f7e7b1f49036a58c4486f089e0444ac9/closure/goog/ui/popupbase.js#L549), that I'm sort of a believer of ;)

It's ready to be reviewed.

FezVrasta:
Thanks!

atomiks:
Doesn't this have a problem of creating a new document event listener for every single tooltip?

(In tippy there is a single global click listener that handles all clicks).

piglovesyou:
@atomiks Yes, you're right. To be more modest, it can `addEventListener` only on the tooltip gets opened and `removeEventListener` when it disappears, I guess.

bodyflex:
could we get a release with this?

PR #672, opened 2018-08-20
_tooltipNode.parentNode may be 'body'

brucewar:
if the tooltip was appended to `body`, `_tooltipNode.parentNode` may be `body`


FezVrasta:
Thanks for the PR!

May you articulate a bit the problem? I don't quite get it.

brucewar:
image

FezVrasta:
Sorry I don't think I understand, care to provide a codepen to reproduce the problem?

PR #679, opened 2018-09-03
Fix blurry text issue

mikila85:
Fix the blurry text issues caused by the translate3d which is not needed since there is no Z dimension.


FezVrasta:
The z translation is needed to enable the hardware acceleration. There shouldn't be any blurry text because we already round the values. Please open an issue if you are having problems.

PR #697, opened 2018-10-09
Add integrity sha to yarn lockfile

mckenna:
[Yarn 1.10](https://github.com/yarnpkg/yarn/releases/tag/v1.10.0) added integrity SHAs to the lock file. Adding them to the lock file here so people working with Yarn 1.10+ don't get an updated lock file each time they run yarn.


FezVrasta:
thanks!

PR #714, opened 2018-11-16
Use the arrow element's parent to compute margin and border

### LINKED ISSUE #710: Arrow position incorrect when an inner element has a border
### CodePen demo

https://codepen.io/anon/pen/dQpGdm

I noticed this after creating a tooltip theme with a border.

img

1.  Zoom right into both demos

- The top one has an inner element with the arrow inside it, and its position is incorrect (1px too far -- happens for both horiz and vertical placements).

- The bottom one has the arrow directly inside the popper, and its position is correct.

It only happens when the inner element has a `border`.

Correct position of the arrow

### What went wrong?

Arrow is 1px too far (aka the `top` for horizontal placement and `left` for vertical placement is rounded? too much).

atomiks:
Fixes #710 

Can't seem to run the tests.

New: https://codepen.io/anon/pen/yQXRaG (line 1412)
Old: https://codepen.io/anon/pen/dQpGdm (v1.14.5)

I'm kind of worried about this assumption as a change to the behavior but it seems necessary to fix the problem mentioned in the issue

### DIFF
--- a/packages/popper/src/modifiers/arrow.js
+++ b/packages/popper/src/modifiers/arrow.js
@@ -74,8 +74,8 @@ export default function arrow(data, options) {
   // take popper margin in account because we don't have this info available
   const popperCss = getStyleComputedProperty(data.instance.popper);
   const parentCss = getStyleComputedProperty(arrowElementParent);
-  const marginProperty = 'margin' + sideCapitalized;
-  const borderProperty = 'border' + sideCapitalized + 'Width';
+  const marginProperty = `margin${sideCapitalized}`;
+  const borderProperty = `border${sideCapitalized}Width`;
   const popperMarginSide = parseFloat(popperCss[marginProperty], 10);
   const popperBorderSide = parseFloat(popperCss[borderProperty], 10);
   const parentMarginSide = parseFloat(parentCss[marginProperty], 10);
--- a/packages/popper/src/modifiers/arrow.js
+++ b/packages/popper/src/modifiers/arrow.js
@@ -36,8 +36,6 @@ export default function arrow(data, options) {
       return data;
     }
   }
-
-  const arrowElementParent = arrowElement.parentNode;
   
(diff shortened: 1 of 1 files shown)


FezVrasta:
Doesn't this break the other case? (when the popper has a margin)?

atomiks:
In the CodePen, it looks like the arrow is correctly placed when the popper has a border as well still. 

As for margin, well if you add `margin-bottom: 1px` to the `#tip, #pop2` selector, it's incorrectly placed for both `data.instance.popper` and `arrowElement.parentNode`. Not sure if it's expected.

atomiks:
If the arrow was not a direct child of the popper, and the popper has a border/margin, then it changes the behavior. 

No change here:

```html
<div id="pop" style="border: 1px solid black;">
  <div x-arrow></div>
</div>
```

Behavior changes here:

```html
<div id="pop" style="border: 1px solid black;">
  <div>
    <div x-arrow></div>
  </div>
</div>
```

However I don't think this would ever be the case because you need a positioning context  to position the arrow properly which would mean the direct parent would have the styling. 

Generally, most would do this:

```html
<div id="pop">
  <div style="border: 1px solid black;">
    <div x-arrow></div>
  </div>
</div>
```

Maybe to prevent breaking current behavior we need to also check the popper too

atomiks:
@FezVrasta I've made some changes so it's backwards-compatible now

Could you run the tests on it? 🙏

FezVrasta:
@atomiks tests run automatically on Travis CI. They are passing, but I don't think we have an extensive test suite for the arrow modifier in any case.

I will test this manually when I find some time. Thanks!

atomiks:
It would be cool to get this and the rounding fix in a new patch release 💙

atomiks:
Gonna close due to bloat caused by backwards compatibility. Maybe consider it for v2. For now you can use a `translate` rule to correct it by 1px for bordered themes...

PR #719, opened 2018-11-20
Additional rounding patch: round element widths before checking width parity

atomiks:
We forgot to round the element widths before checking their width parity (since elements' widths are usually not always integers)

https://codepen.io/anon/pen/aQVVwO


atomiks:
@FezVrasta any reason this wasn't merged before 1.14.6? All variation placements are properly rounded in 1.14.6, but without this patch, the centering can sometimes be off due to decimal width/height because the parity check doesn't work.

FezVrasta:
Sorry I forgot about this

PR #725, opened 2018-12-03
Fix the documentation for tooltip options

stof:


FezVrasta:
Thanks for the PR.

The docs are auto generated from the comments in the code. You should fix the comment to see the fix.

stof:
@FezVrasta I updated this, using the same syntax than for other properties which have a space in their default value, i.e. wrapping the property in `[ ]` as defined in JSDoc.

This seems to affect the display of the property name, but that's already the case for `options.template` or `options.trigger`.

PR #728, opened 2018-12-06
fix: isFixed with detached DOM element

gierschv:
Hi there,

We use popper.js with Vue.js, and we noticed that it generates a lot of errors noise when a DOM element gets detached and the `preventOverflow` is called: `isFixed` tries to go up the DOM tree and tries to resolve a `nodeName` of an undefined parent.

So here is a fix for the isFixed util to avoid crashing on detached DOM elements (this works perfectly with this patch).

Thanks!


FezVrasta:
Thanks!

gierschv:
@FezVrasta Do you think you could make a patch release with this fix? Thanks by advance!

Issue #687, opened 2018-09-21
Focus tooltips don't show when manually triggerring focus on blur


alexturpin:
In our app, if a user tabs out of a field where there is a validation error, we focus back on the field (only once) to show them the error in a focus tooltip. I've been porting our tooltip library to Tooltip.js and noticed that if have a field with a tooltip that should trigger on `focus` and you try to manually trigger focus on that field during the `blur` event, the tooltip doesn't show.

Here is a demo: http://jsfiddle.net/fkat7m4p/

The cause of this is that even when there is no `delay` configured, `setTimeout` is used to trigger `_show` and `_hide`. It essentially means that the manual focus is two ticks early. One can work around with two `setTimeout` set to `0` as shown here: http://jsfiddle.net/k2onwrxd/

I would be happy to submit a PR that doesn't use `setTimeout` when the delay is 0, but I'd just like to check before if this is something that would be accepted / that you would be interested in.

alexturpin:
Another consequence of using `setTimeout` in such cases is that if the tooltip text is changed on blur, there can be a flash of the new text before the tooltip is properly hidden on the next tick.

FezVrasta:
Hi, I'm interrupting the development of the `tooltip.js` library to focus on the development of `@popperjs/core` (Popper.js v2).

Please consider switching to [Tippy.js](https://github.com/atomiks/tippyjs), which is a way better alternative to Tooltip.js and is well maintained. Tippy.js still uses Popper.js under the hood, so you will find all your favourite features.

I'll close this issue.

PR #742, opened 2019-01-25
Don't use setTimeout when not needed

alexturpin:
Fix for #687


alexturpin:
Sorry I'm bad at pull requests and hadn't realized the changes would all be included everytime. This is the order in which they should be merged and squashed for best commits: #742, #743, #744

FezVrasta:
Thanks for the PR, rather than repeating the logic, it may be useful to create an helper function that encapsulates the condition:

```
const setConditionalTimeout = (fn, delay) => delay === 0 ? fn() : window.setTimeout(fn, delay);
```

PR #743, opened 2019-01-25
Fix manual hide

alexturpin:
Fix for #727

PR #744, opened 2019-01-25
Memory leak

alexturpin:
There is a memory leak happening because the `mousedown` event on document never gets cleared for  each popper. This fixes it.


alexturpin:
Sorry I'm bad at pull requests and hadn't realized the changes would all be included everytime. This is the order in which they should be merged and squashed for best commits: #742, #743, #744

sandstrom:
friendly ping @FezVrasta 😄

FezVrasta:
CodeClimate checks are failing (there's a lot of code repetition in this PR).

If anyone wants to adopt the PR, fix the CodeClimate errors, and maybe add a test to make sure the fix works, I'll happily merge it.

sandstrom:
@FezVrasta Understand, those are valid points! 

Awesome work you've done on popper, it's a beautiful library!

PR #750, opened 2019-01-29
Fix TS: padding types of `flip` and `preventOverflow` #518

kasparsuvi1:
Added types of padding, that are missing since #615 PR.


FezVrasta:
Thanks! May you do the same for the Flow types? (they are in `packages/popper/index.js.flow`) the syntax is the same

kasparsuvi1:
Yes, added Flow types also and made Padding props optional.

FezVrasta:
Thanks!

PR #754, opened 2019-02-14
Add 'flipVariationsByContent' option to flip modifier

### LINKED ISSUE #512: Is there a way to apply `flip` before `preventOverflow`?
### CodePen demo

https://codepen.io/Justineo/pen/QaypBM

1. with `placement` set to `bottom-start`
2. when the reference is on the right side of the container
3. the popper is shifted (`#pop`)

Provide a way to apply `flip` when the popper is placed outside the container rather than shift it. (Like `#pop2` in the CodePen demo.)

image

### What went wrong?

Sometimes we want to flip the popper when it's outside the container, not when it overlaps with the reference element.

### Any other comments?

I've looked into how popper handles this now, but not sure how to handle this.

atomicbrainman:
Should close #512 

Basically, `flipVariations` switches the variation if **reference element** overflows boundaries (good when reference element is larger than popper)

This PR adds `flipVariationsByContent`, which switches the variation if **popper itself** overflows boundaries (good when popper content is larger than the reference element, usually a button or input)

### DIFF
--- a/packages/popper/index.d.ts
+++ b/packages/popper/index.d.ts
@@ -72,6 +72,8 @@ declare namespace Popper {
       behavior?: Behavior | Position[],
       padding?: number | Padding,
       boundariesElement?: Boundary | Element,
+      flipVariations?: boolean,
+      flipVariationsByContent?: boolean,
     };
     inner?: BaseModifier;
     hide?: BaseModifier;
--- a/packages/popper/index.js.flow
+++ b/packages/popper/index.js.flow
@@ -83,6 +83,8 @@ declare module 'popper.js' {
       behavior?: Behavior | Position[],
       padding?: number | Padding,
       boundariesElement?: Boundary | Element,
+      flipVariations?: boolean,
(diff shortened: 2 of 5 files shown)


berdyshev:
this is what I really need. thanks!

xxia34:
I really need this as well!

FezVrasta:
I'm sorry for the delay, this looks great, but without some supporting test to make sure the functionality doesn't regress I find it difficult to merge it as is.

May you add some tests please?

Justineo:
@FezVrasta Is there any recent plan to release a new version with this feature?

FezVrasta:
1.15.0 released

Justineo:
Great, thanks!

Tiano1972:
Cool
Thanks

pleek91:
is there an example of how to use this? I'm having the same issue as #512. Searched the docs for `flipVariations` and `flipVariationsByContent` and didn't find any results.

PR #759, opened 2019-03-12
Fix(types): add missing `reference` and `popper` class properties

medfreeman:
Reference: https://github.com/FezVrasta/popper.js/blob/bd1faa9a27499433443bb4c42b3dea9ecf055777/packages/popper/src/index.js#L37-L38

### DIFF
--- a/packages/popper/index.d.ts
+++ b/packages/popper/index.d.ts
@@ -158,8 +158,8 @@ declare class Popper {
   static Defaults: PopperOptions;
 
   options: PopperOptions;
-  popper: HTMLElement;
-  reference: HTMLElement;
+  popper: Element;
+  reference: Element | ReferenceObject;
 
   constructor(reference: Element | ReferenceObject, popper: Element, options?: PopperOptions);
 
--- a/packages/popper/index.d.ts
+++ b/packages/popper/index.d.ts
@@ -158,6 +158,8 @@ declare class Popper {
   static Defaults: PopperOptions;
 
   options: PopperOptions;
+  popper: HTMLElement;
+  reference: HTMLElement;
 
   constructor(reference: Element | ReferenceObject, popper: Element, options?: PopperOptions);
 
--- a/packages/popper/index.js.flow
+++ b/packages/popper/index.js.flow
@@ -142,6 +142,9 @@ declare module 'popper.js' {
   declare class Popper {
     static placements: Placement;
 
+    popper: Element;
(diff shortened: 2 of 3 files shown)


packages/popper/index.d.ts
```diff
@@ -158,6 +158,8 @@ declare class Popper {
   static Defaults: PopperOptions;
 
   options: PopperOptions;
+  popper: HTMLElement;
+  reference: HTMLElement;
 
   constructor(reference: Element | ReferenceObject, popper: Element, options?: PopperOptions);
```
atomiks: `reference: Element`, not `HTMLElement`?

FezVrasta: it can also be "custom reference object", we should already have a type for that somewhere

medfreeman: Shouldn't have done this at the end of the work day...
It should be:
```
popper: Element;
reference: Element | ReferenceObject;
```

I'll try to test and commit before leaving today.

medfreeman: It is a bit conflicting also, because in the code you have `HTMLElement` types for `popper` and `reference`, and `Element` in type definitions. 

https://github.com/FezVrasta/popper.js/blob/bd1faa9a27499433443bb4c42b3dea9ecf055777/packages/popper/src/index.js#L17-L18

I'm not sure which one should be the correct one.
See this [stackoverflow answer](https://stackoverflow.com/a/6581728) for details.

FezVrasta: The JSDoc types are most likely wrong...  But I don't have much time to verify it

atomiks: It can be an `SVGElement` too. So that's why it's the base `Element` interface

medfreeman:
It seems good now!

FezVrasta:
@medfreeman may you double check the Flow types are in sync please?

medfreeman:
sure, about this feature only or in general ?
i've already seen missing properties, `modifiers`, `Defaults`, etc..
I could edit them, but not so familiar with flow syntax anymore.

medfreeman:
I'll have to do it tomorrow morning anyway.

FezVrasta:
I was thinking just about this specific fix.

But if you find other issues feel free to send PRs :-)

medfreeman:
I won't hesitate!
Ok, all good now!
Thanks for your library and reactivity!

PR #767, opened 2019-04-03
Use translate instead of translate3d

szhu:
In some browsers like Chrome, using `translate3d(x, y, 0)` results in blurry text, while `translate(x, y)` doesn't.

### DIFF
--- a/packages/popper/src/modifiers/computeStyle.js
+++ b/packages/popper/src/modifiers/computeStyle.js
@@ -50,7 +50,7 @@ export default function computeStyle(data, options) {
   const sideB = y === 'right' ? 'left' : 'right';
 
   // if gpuAcceleration is set to `true` and transform is supported,
-  //  we use `translate3d` to apply the position to the popper we
+  //  we use `translate` to apply the position to the popper we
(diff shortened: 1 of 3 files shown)


atomiks:
Have you tested the perf impact? I wonder if it matters much - if not, don't see why it can't be changed

FezVrasta:
I use translate3d for a performance reason. The 2d transformation is not as efficient

atomiks:
@FezVrasta 

I tested with 80 poppers `translate` vs `translate3d` while scrolling, there doesn't seem to be a noticeable difference tested via Performance DevTools. Both take up the same amount of rendering time. Doesn't `will-change: transform` provide the performance boost anyway?

FezVrasta:
Last time I ran my tests, `translate3d` gave a good performance boost, if you'd like to share your benchmark I can look deeper into it.

szhu:
@FezVrasta I would like to run performance testing on this but I don't have a lot of background. Is there a performance test suite for popper.js, or what code did you use to figure out that translate3d was more performant?

PR #777, opened 2019-04-17
Update README.md  err translated into Portuguese would be perfect for reading!

desenvolvedorIcotrade:


FezVrasta:
Thanks!

atomiks:
It was correct before though...

PR #781, opened 2019-04-26
Fix wrong style types in index.d.ts

art5255:

### DIFF
--- a/packages/popper/index.d.ts
+++ b/packages/popper/index.d.ts
@@ -97,6 +97,13 @@ declare namespace Popper {
     height: number;
   }
 
+  type CSSStyleProperties = Pick<CSSStyleDeclaration, Exclude<keyof CSSStyleDeclaration,
+    | "top"
+    | "left"
+    | "width"
+    | "height"
+  >> & Offset;
+
   export interface Data {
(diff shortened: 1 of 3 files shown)


packages/popper/index.d.ts
```diff
@@ -97,15 +97,22 @@ declare namespace Popper {
     height: number;
   }
 
+  type CSSStyleProperties = Pick<CSSStyleDeclaration, Exclude<keyof CSSStyleDeclaration,
```
FezVrasta: What does this do?

art5255: Omit Offset properties from CSSStyleDeclaration.

FezVrasta: Why? One could override them if they wanted

art5255: I don`t understand how. In CSSStyleDeclaration, width and other offset properties are `string | null`, but in Offset they are numbers.

FezVrasta: The `styles` object is literally what gets applied to the HTML element, it can be any valid CSS value. While Offset is what gets computed by Popper.js, and must follow the Popper.js logic

PR #790, opened 2019-06-05
Fixed calculation that results in NaN

### LINKED ISSUE #789: getBoundingClientRect results in NaN width/height for elements with 0 clientHeight or clientWidth
### CodePen demo

https://codepen.io/anon/pen/qGGqpa

In this codepen, the popper still shows up; it's unclear to me how or why this happens. That being said, I debugged via Chrome and still see the calculation resulting in `NaN` (however, it doesn't taint the results, as seen when debugging poppers created in my app via Bootstrap). See comments for more details.

1.  Create an element for which `element.clientWidth` is `0`
2.  Attempt to create a Popper with placement: top

Popper should show up

### What went wrong?

Popper doesn't show up

### Any other comments?

`getBoundingClientRect()` returns `NaN` for the `right` and `width` properties in this case, due to [these lines](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getBoundingClientRect.js#L46-L49). If both `sizes.width` (`undefined` for any element which is not `html`) and `element.clientWidth` (`0` for any very thin element, or at least for mine) are falsy, the expression evaluates `result.right - result.left`. However,  `result` is defined [immediately above](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getBoundingClientRect.js#L37-L42) and will _never_ have `right` or `bottom` properties. I suspect these lines were intended to be `rect.right` and `rect.bottom`, but it's unclear why they couldn't just be initialized to `result.width` and `result.height` for this third case. I was unable to reproduce in a codepen, but editing these lines in `./node_modules/popper.js/dist/esm/popper.js` accordingly resolves the issue in my app.

snoozbuster:
I would add a test for this but it seems like this function doesn't already have tests (or at least I was unable to find them). Please advise if they are in fact just hiding somewhere.

Resolves #789.

### DIFF
--- a/packages/popper/src/utils/getBoundingClientRect.js
+++ b/packages/popper/src/utils/getBoundingClientRect.js
@@ -44,9 +44,9 @@ export default function getBoundingClientRect(element) {
   // subtract scrollbar size from sizes
   const sizes = element.nodeName === 'HTML' ? getWindowSizes(element.ownerDocument) : {};
   const width =
-    sizes.width || element.clientWidth || result.right - result.left;
+    sizes.width || element.clientWidth || result.width;
   const height =
-    sizes.height || element.clientHeight || result.bottom - result.top;
+    sizes.height || element.clientHeight || result.height;
 
   let horizScrollbar = element.offsetWidth - width;
   let vertScrollbar = element.offsetHeight - height;


packages/popper/src/utils/getBoundingClientRect.js
```diff
@@ -44,9 +44,9 @@ export default function getBoundingClientRect(element) {
   // subtract scrollbar size from sizes
   const sizes = element.nodeName === 'HTML' ? getWindowSizes(element.ownerDocument) : {};
   const width =
-    sizes.width || element.clientWidth || result.right - result.left;
+    sizes.width || element.clientWidth || result.width;
```
FezVrasta: AFAIK `width` is not available on some Internet Explorer versions, that's why I didn't rely on it.

May we keep both? `result.right - result.left || result.width`

PR #802, opened 2019-07-09
fix: ESM style export of Padding in TS typedefs

Benjamin-Dobell:
TypeScript definitions only declared `Padding` as available via a namespace, rather than ESM style exports. This is a one-liner to fix that.


FezVrasta:
Thanks!

PR #829, opened 2019-09-17
Fix `.js.flow` definitions

pascalduez:
Hi,

in the current state of affairs, the `umd/popper.js.flow` definitions shipped alongside the `umd/popper.js` (package main entry) just doesn't work. It all resolves to `any`.

That's why you have to do [this sort of thing](https://github.com/FezVrasta/react-popper/blob/fa1b2956255666244ae102ae62f97d521f42a0a7/.flowconfig#L9) which should not be necessary. 

Causes:
1. There's a typo in the file name :) (this PR fixes it)
2. `.js.flow` declaration files have a different behaviour and syntax than "library definitions".
- They need a `// @flow` pragma
- They can't use the `declare modules` wrapper
- Types needs to be exported `export type ...`

So I managed to make this file work locally by applying all the points above, but I'm not sure how would you like to integrate it in the build workflow here. Because it makes this file significantly different from the `index.js.flow` at the package root. So waiting for feedback before pushing something.

My advice would be, don't maintain two files. Either only include the `popper.js.flow` (modified file).
Or move the library definition to `flow-typed`.

### DIFF
--- a/packages/popper/index.js.flow
+++ b/packages/popper/index.js.flow
@@ -1,157 +1,156 @@
-declare module 'popper.js' {
-  declare type Position = 'top' | 'right' | 'bottom' | 'left';
-  declare type Placement =
-    | 'auto-start'
-    | 'auto'
-    | 'auto-end'
-    | 'top-start'
-    | 'top'
-    | 'top-end'
-    | 'right-start'
-    | 'right'
-    | 'right-end'
-    | 'bottom-end'
-    | 'bottom'
-    | 'bottom-start'
-    | 'left-end'
-    | 'left'
-    | 'left-start';
+// @flow
 
-  declare type Offset = {
-    top: number,
-    left: number,
-    width: number,
-    height: number,
-    position: Position,
-  };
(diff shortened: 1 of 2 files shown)


FezVrasta:
The one in `umd` should be a direct copy of the one in the root, if you edit `/packages/popper/src/index.js.flow` you are going to update the one in the UMD folder.

pascalduez:
@FezVrasta Yes, but it would be a "breaking" I guess, after changes this file will not be usable as "library definition" anymore.  But probably not much users try [to use it as such](https://github.com/FezVrasta/react-popper/blob/fa1b2956255666244ae102ae62f97d521f42a0a7/.flowconfig#L9)?

FezVrasta:
It was never intended to be used as library definition, I just wanted it to work as is out of the box, so I'd say it falls in the "breaking change of something that never intended to work that way", which is good to me if you ask me.

pascalduez:
@FezVrasta I pushed the updated file.

PR #840, opened 2019-11-21
fix(detectOverflow): remove boundary margin calc

atomiks:
Didn't do much but I think you wanted a PR, merging into the feature branch you have

I think this is the behavior you want as it matches v1 as far as I can tell regarding body/boundary margins

### DIFF
--- a/src/modifiers/detectOverflow.js
+++ b/src/modifiers/detectOverflow.js
@@ -45,13 +45,12 @@ export function detectOverflow(
     },
   });
 
-  const popperClientRect = {
-    ...popperOffsets,
+  const popperClientRect = Object.assign({}, popperOffsets, {
     top: popperOffsets.y,
     bottom: popperOffsets.y + popperRect.height,
     left: popperOffsets.x,
     right: popperOffsets.x + popperRect.width,
-  };
+  });
 
   state.modifiersData.detectOverflow = {
     top: boundaryClientRect.top > popperClientRect.top,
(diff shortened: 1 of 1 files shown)


src/modifiers/detectOverflow.js
```diff
@@ -55,23 +45,22 @@ export function detectOverflow(
     },
   });
 
-  console.log(boundaryMargins);
-
-  const popperClientRect = Object.assign({}, popperOffsets, {
+  const popperClientRect = {
+    ...popperOffsets,
```
FezVrasta: I think I’m trying to avoid to use es next features to reduce the bundle  size

atomiks: So you wanted people to polyfill things like `Object.assign` if they need to? I still don't know how bundlers will be handling modern (~ES2017) and legacy (ES5) code in the future. 

IIRC they want to make ES2017 default on npm and make people transpile `node_modules` in the future to enable differential bundling, or there will be a `targets` field in package.json, not sure 🤔

atomiks: Regardless, I think it would be ES2017 but the spread op is 2018

FezVrasta: Object.assign is well supported, the spread operator when transpiled generates a lot of code

FezVrasta: I just realized I'm using spread operators all around the code, nevermind, ahah

PR #842, opened 2019-11-21
feat: implement DEV-only messages

### LINKED ISSUE #841: [next] Implement DEV-only messaging
```js
if (__DEV__) {
  console.warn(...);
}
```

https://www.npmjs.com/package/babel-plugin-dev-expression

This way warnings don't end up in production bundles.

atomiks:
You already had `process.env.NODE_ENV` check, but I had to refactor it so it'll work for consumers (and replaced with easy `__DEV__` variable). 

I also added `"sideEffects": false` as I'm not sure if DCE for these branches works without it (it needed to be added anyway).

Resolves #841

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -24,7 +24,7 @@ const createUmdBundle = ({ minify } = {}) => ({
   },
 });
 
-const esBundle = {
+const esmBundle = {
   input: 'src/index.js',
   plugins: [
     babel(),
@@ -35,8 +35,8 @@ const esBundle = {
   ].filter(Boolean),
   output: {
     name: 'Popper',
-    file: `${dir}/es/index.js`,
-    format: 'es',
+    file: `${dir}/esm/index.js`,
+    format: 'esm',
     sourcemap: true,
   },
 };
@@ -67,14 +67,19 @@ const devBundle = {
   ],
   output: {
     name: 'Popper',
-    file: `${dir}/es/index.js`,
-    format: 'es',
+    file: `${dir}/esm/index.js`,
+    format: 'esm',
     sourcemap: true,
   },
 };
 
 const builds = IS_DEV
   ? [devBundle]
-  : [createUmdBundle(), createUmdBundle({ minify: true }), esBundle, cjsBundle];
+  : [
+      createUmdBundle(),
+      createUmdBundle({ minify: true }),
+      esmBundle,
+      cjsBundle,
+    ];
 
 export default builds;
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -11,9 +11,7 @@ const umdBundle = minify => ({
   plugins: [
     babel(),
     replace({
-      'process.env.NODE_ENV': JSON.stringify(
-        process.env.NODE_ENV || 'development'
-      ),
+      __DEV__: "process.env.NODE_ENV !== 'production'",
     }),
     minify && terser(),
     bundleSize(),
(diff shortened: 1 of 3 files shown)


atomiks:
Realizing that the `umd` bundle needs to have 2 versions (DEV and PROD) since theres no Node variables. Will push in a sec

atomiks:
Two remaining things:

- It seems Rollup prefers `esm` over `es` word by default now. Should we use that?
- I think we should call the file `popper` not `index`. It shows up better in DevTools (e.g. perf tools) and stuff like `webpack-sourcemap-analyzer` when you can see the file name.

rollup.config.js
```diff
@@ -6,14 +6,12 @@ import { terser } from 'rollup-plugin-terser';
 const IS_DEV = process.env.NODE_ENV === 'development';
 const dir = IS_DEV ? 'tests/visual/dist' : 'dist';
 
-const umdBundle = minify => ({
+const createUmdBundle = ({ minify } = {}) => ({
```
FezVrasta: Why `createUmdBundle`, but `esBundle` instead of `createEsBundle`?

atomiks: `esBundle` is just a static object

FezVrasta:
1. `esm` is okay to me
2. Maybe we could have `index.js` that only requires `popper.js`, at least we keep the standard entrypoint to help new contributors to see where's the entrypoint?
3. I was considering to get rid of the `.js` suffix in the package name, maybe call it `popperjs` to make everybody happy? We also have the `@popper` namespace we could use to do something like `@popper/core` or whatever;
3b. I'd keep RequireJS support, Bootstrap depends on Popper.js, and their customers use a lot of legacy systems;
4. I like the folder approach, especially because the `main` entrypoint needs to live together with all the `.js.flow` files that need to live in the same folder;

atomiks:
1. Done
2. So you can do `import Popper from 'popper.js/dist/esm'` without specifying any filename, right? I wonder how necessary that is, is it difficult for Rollup to generate a bundle that does that?
3. I do like that! `@popper/core`, `@popper/react` seems nice. I've sort of embraced the `.js` in my lib but it's kind of tacky 😆 
4. Makes sense, we'll leave it

FezVrasta:
I'm not sure if Rollup should be concerned about this, I was mainly thinking about the need to have a `src/index.js` to help people understand where the code starts.

atomiks:
Ah I meant the dist files. So `dist/popper.js` not `dist/index.js`.

Btw what's the purpose of the  `"module": "lib/index.js"`, can't they use the `esm` dist file? If you want to export a bunch of utils, that should be possible with `esm`?

FezVrasta:
I see, yeah let's use `popper.js` instead then.

I'm not sure what you mean with your question, `"module"` should be used by bundlers to resolve what file to import, or not?

atomiks:
You have a separate build script for the `es` version, so it's not using Rollup's bundled file, it's a directory with a bunch of separate files.

FezVrasta:
WHat's the issue? I don't follow 😅

atomiks:
You have this:

```json
  "module": "lib/index.js",
```

and this script

```json
  "build:es": "babel src -d lib --ignore '**/*.test.js','**/__mocks__'",
```

Why? instead of (after the renaming I've done):

```json
  "module": "dist/esm/popper.js",
```

Using Rollup's file?

FezVrasta:
Because that's a bundle, it can't get tree shaked (AFAIK) and you can't import modules singularly.

atomiks:
With `"sideEffects": false` bundled files will be tree-shaken (top-level code iirc). You can export individual modules in the bundled file for the consumer as well. With the current file in `lib/index.js`, none of the utils are exported to use either?

What about a separate directory with its own package.json?:

```js
import {getScrollParent} from 'popper.js/dom-utils';
```

FezVrasta:
What's the problem with the `/lib` approach? It seems easier to use to me.

atomiks:
Yeah I'd say it's fine tbh 👍. Forget what I was saying 😅

atomiks:
@FezVrasta I think this PR is fine now, wanna review?

PR #845, opened 2019-11-21
fix(detectOverflow): wrap warning in __DEV__, tweak wording

atomiks:
I'm not sure how to handle multiple lines. We could use template literal strings, concatenation, or this joining technique. What is best?

### DIFF
--- a/src/modifiers/detectOverflow.js
+++ b/src/modifiers/detectOverflow.js
@@ -34,8 +34,8 @@ export function detectOverflow(
     if (__DEV__) {
       console.error(
         [
-          'PopperJS: The "detectOverflow" modifier\'s `boundaryElement` must',
-          'be a parent of the popper element',
+          'Popper: The "detectOverflow" modifier\'s `boundaryElement` must be',
+          'a parent of the popper element.',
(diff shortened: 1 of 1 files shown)


src/modifiers/detectOverflow.js
```diff
@@ -31,9 +31,15 @@ export function detectOverflow(
   const documentElement = options.boundaryElement.ownerDocument.documentElement;
 
   if (!options.boundaryElement.contains(popperElement)) {
-    console.error(
```
FezVrasta: I think the library should throw an error also in production, this is a pretty important error that shouldn't be ignored.

I know React has shorted error messages for production builds, could we do the same?

atomiks: Maybe a dumb idea but we could have an error code for PROD, and it links to a file that explains each code. That way there's no words to bloat the bundle. 

e.g.

```
Popper: Error 5 {link}
```

I'm not sure how practical that would be to maintain

FezVrasta: React uses bit.ly-like links to point to a documentation page, that would work too

PR #852, opened 2019-11-24
Modified nodeName handling for web pages served as XML

tippisum:
When a web page is served with `Content-Type: application/xhtml+xml`, the DOM property `nodeName` will always have a lowercased value instead of a uppercased one typically seen in normal HTML pages.
This modification is meant to help users who want to serve their web pages with `Content-Type: application/xhtml+xml` and is not supposed to break anything.


FezVrasta:
Thanks for the PR, this issue is fixed in the `next` branch (Popper v2) which will be released in the upcoming weeks. Support for XML is not planned for the current version.

PR #853, opened 2019-11-24
[wip] feat: padding for preventOverflow and flip

atomiks:
(Defaults are for testing - `0` is the new default)

- Don't really know what you want the options destructuring etc to look like
- I don't know how `flip` + `preventOverflow` paddings being different for a single side should work


src/types.js
```diff
@@ -81,3 +81,12 @@ export type ClientRectObject = {|
...
   height: number,
 |};
+
+export type Padding =
+  | number
+  | {|
+      top: number,
+      left: number,
+      right: number,
+      bottom: number,
```
FezVrasta: These properties should be optional (`top?:`)

atomiks: IIRC v1 required you to set them to `0`? I haven't really thought that through at all tho.

src/modifiers/flip.js
```diff
@@ -1,20 +1,27 @@
...
   const placement = state.placement;
-  const behavior =
-    options && options.behavior
-      ? options.behavior
-      : [state.options.placement, getOppositePlacement(placement)];
+  const behavior = options.behavior || [
+    state.options.placement,
+    getOppositePlacement(placement),
+  ];
+  const { padding = 5 } = options;
```
FezVrasta: This could made clearer:

```js
const defaultBehavior = [state.options.placement, getOppositePlacement(placement)];
const { behavior = defaultBehavior, padding = 5 } = options;
```

Also, I think it makes sense to keep padding to `0` by default.

atomiks:
By the way I have no idea if the types are correct because idk how to use Flow in VSCode 😳

src/types.js
atomiks: Can you check new commit? It should allow optional properties

src/modifiers/flip.js
```diff
@@ -1,20 +1,27 @@
...
+  const isNumberPadding = typeof padding === 'number';
   const overflow = state.modifiersData.detectOverflow;
 
-  const flippedPlacement = behavior.find(
-    newPlacement => overflow[getBasePlacement(newPlacement)] <= 0
-  );
+  const flippedPlacement = behavior.find(newPlacement => {
+    const basePlacement = getBasePlacement(newPlacement);
+    const paddingValue = isNumberPadding ? padding : padding[basePlacement];
```
FezVrasta: We need to expand a single value into an object with N keys in a few places, 2 in this PR, and one in the `enableEventListeners`.

We could define an utility to simplify our lives:

```js
function expandToHashMap<T: number | string>(value: T, keys: Array<string>): { [key: string]: T } {
	return keys.reduce((hashMap, key) => {
      hashMap[key] = value;
      return hashMap;
    }, {});
}
```

This way you can change this code to:

```js
import { top, right, bottom, left } from '../enums';

const paddingConfig = typeof padding === 'number' ? padding : expandToHashMap(padding, [top, right, bottom, left]);
```

src/types.js
FezVrasta: Looks good now, thanks!

I'm not sure how v1 worked, but defaulting to `0` seems the most reasonable thing.

src/modifiers/flip.js
atomiks: I think the code in the second example is opposite of what it should be, but yeah I agree that would be much better

FezVrasta:
> By the way I have no idea if the types are correct because idk how to use Flow in VSCode 😳

You can install this extension https://github.com/flowtype/flow-for-vscode

Or just run `yarn flow` in the repository root.

src/modifiers/flip.js
```diff
@@ -4,6 +4,8 @@ import type { State, Modifier, Rect, Padding } from '../types';
 import getOppositePlacement from '../utils/getOppositePlacement';
 import getBasePlacement from '../utils/getBasePlacement';
 import mergePaddingObject from '../utils/mergePaddingObject';
+import expandToHashMap from '../utils/expandToHashMap';
+import { right, bottom, left } from '../enums';
```
FezVrasta: you forgot to import `top`

src/modifiers/preventOverflow.js
```diff
@@ -5,6 +5,7 @@ import getBasePlacement from '../utils/getBasePlacement';
 import getMainAxisFromPlacement from '../utils/getMainAxisFromPlacement';
 import getAltAxis from '../utils/getAltAxis';
 import mergePaddingObject from '../utils/mergePaddingObject';
+import expandToHashMap from '../utils/expandToHashMap';
```
FezVrasta: need to also import `top,right,bottom,left`

src/modifiers/flip.js
atomiks: Lol `top` is a property on the `window`? VSCode auto imports didn't work cuz of that I guess 🤦‍♂

src/utils/expandToHashMap.js
```diff
@@ -0,0 +1,9 @@
+export default function expandToHashMap<T: number | string>(
```
FezVrasta: files should start with `// @flow` (I need to add a linting rule for this)

src/modifiers/flip.js
FezVrasta: lol yeah it looks like so

src/modifiers/preventOverflow.js
```diff
@@ -1,45 +1,56 @@
...
 
 export function preventOverflow(
   state: State,
-  options?: { mainAxis: boolean, altAxis: boolean } = {}
+  options?: { mainAxis: boolean, altAxis: boolean, padding: Padding } = {}
 ) {
   const {
     mainAxis: checkMainAxis = true,
     altAxis: checkAltAxis = false,
+    padding = { left: 50, top: 20, right: 0, bottom: 0 },
```
FezVrasta:
```suggestion
    padding = 0,
```

src/modifiers/flip.js
```diff
@@ -1,20 +1,33 @@
...
   const placement = state.placement;
-  const behavior =
-    options && options.behavior
-      ? options.behavior
-      : [state.options.placement, getOppositePlacement(placement)];
+  const defaultBehavior = [
+    state.options.placement,
+    getOppositePlacement(placement),
+  ];
+  const { behavior = defaultBehavior, padding = 5 } = options;
```
FezVrasta:
```suggestion
  const { behavior = defaultBehavior, padding = 0 } = options;
```

atomiks:
There are some flow errors but idk why

FezVrasta:
I manually merged it and fixed the type errors. Thanks!

atomiks:
You accidentally kept the non-`0` default value for padding

FezVrasta:
Fixed! Thanks!

PR #855, opened 2019-11-25
fix: variation placement logic

atomiks:


src/utils/computeOffsets.js
```diff
@@ -65,18 +63,18 @@ export default ({
...
   }
 
-  const mainAxis = placement ? getMainAxisFromPlacement(placement) : null;
-  const altAxis = mainAxis ? getAltAxis(mainAxis) : null;
-  const len = altAxis === 'x' ? 'width' : 'height';
-  const axis = [right, left].includes(basePlacement) ? mainAxis : altAxis;
+  const mainAxis = basePlacement
+    ? getMainAxisFromPlacement(basePlacement)
+    : null;
+  const len = mainAxis === 'y' ? 'height' : 'width';
```
FezVrasta: I think this could live in the if clause below.

PR #856, opened 2019-11-25
fix: blurry text

atomiks:
Let's discuss techniques here before merging

We could:

1. Force `top/left` for  <`@2x` DPI displays (current PR)
2. Round offsets but keep `gpuAcceleration`, however sometimes this still causes blurry text (seen in several issues) due to the accelerated layer or whatever, and can cause 0.5px rounding errors for centered poppers

### DIFF
--- a/src/modifiers/computeStyles.js
+++ b/src/modifiers/computeStyles.js
@@ -29,8 +29,10 @@ export const computePopperStyles = ({
   strategy: PositioningStrategy,
   gpuAcceleration: boolean,
 }) => {
+  // @1x displays in Chrome can be blurry due to non-rounded offsets, but this
+  // introduces slight positioning errors. TODO: somehow solve this better
   // by default it is active, disable it only if explicitly set to false
-  if (gpuAcceleration === false) {
+  if (gpuAcceleration === false || window.devicePixelRatio < 2) {
     return {
       top: `${offsets.y}px`,
       left: `${offsets.x}px`,


atomiks:
## Pros for 1.

- Blurry text should never be an issue on low DPI display debices since there is no GPU acceleration.
- Super easy/smallest code to make it work.

## Cons for 1.

- Worse perf (by how much though?) for low DPI display devices. 
- No subpixel translation possible for low DPI display devices for popper updates if there is a transition applied (looks slightly jerky like `height` or `width` changes for example).

## Pros for 2.

- Consistent attribute styles across all devices.
- Best perf for all devices.

## Cons for 2.

- There will _always_ be the potentiality for blurry text on low DPI displays because of the popper being on an accelerated layer. Especially regarding zooming. 
- High DPI displays now have 0.5px rounding errors when it can be avoided on them (not possible on low DPI unless you want blurry text almost all the time + a slightly blurry edge because the popper sits inside a subpixel).
- More code and bundle bloat.

---

I don't like either of them but I feel like 1. is better overall. There will never be any blurry issues (crisp text always) at the cost of slightly worse perf and different styles applied based on the device. 

UX > DX

atomiks:
## Pros for 1.

- Blurry text should never be an issue on low DPI display devices since there is no GPU acceleration.
- Super easy/smallest code to make it work.

## Cons for 1.

- Worse perf (by how much though?) for low DPI display devices. 
- No subpixel translation possible for low DPI display devices for popper updates if there is a transition applied (looks slightly jerky like `height` or `width` changes for example).

## Pros for 2.

- Consistent attribute styles across all devices.
- Best perf for all devices.

## Cons for 2.

- There will _always_ be the potentiality for blurry text on low DPI displays because of the popper being on an accelerated layer. Especially regarding zooming. 
- High DPI displays now have 0.5px rounding errors when it can be avoided on them (not possible on low DPI unless you want blurry text almost all the time + a slightly blurry edge because the popper sits inside a subpixel).
- More code and bundle bloat.

---

I don't like either of them but I feel like 1. is better overall. There will never be any blurry issues (crisp text always) at the cost of slightly worse perf and different styles applied based on the device. 

UX > DX

FezVrasta:
I will run some benchmarks with the gpu acceleration disabled, maybe the perf is not that bad with v2

FezVrasta:
I ran some tests and the gpu acceleration is still giving a nice perf boost.

PR #857, opened 2019-11-26
docs: add usage for different formats

atomiks:
I also renamed `popper-minimal` to `popper-base` because I kept getting "minimal" confused with "lite". IMO this is much less confusing naming.

I kind of want to add little picture examples at the top of what Popper solves in the introduction as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,23 @@
 
 Positioning tooltips (_but also dropdowns, popovers, and more_) is difficult. **Popper** is here to help!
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -21,13 +29,13 @@ Since we write UIs using powerful abstraction libraries such as React or Angular
 yarn add @popperjs/core@next
 
 # With npm
-npm install --save @popperjs/core@next
+npm i @popperjs/core@next
```
FezVrasta: —save is required

atomiks: It hasn't been since npm 5 or 6

FezVrasta: Ah 🧐

PR #858, opened 2019-11-26
fix: 1px rounding error for variation placements

atomiks:


FezVrasta:
It's for things like that that the v1 code base is so complicated 😭

We need to take some time and figure out a clean solution to the problem.

atomiks:
Well I think the issue is reference element's margins shouldn't be included in its rect (only popper should).

And we need to round the values to prevent 1px errors for variation placements.

FezVrasta:
I manually merged the fix for the reference margins. Could we keep this PR just for the rounding issue?

PR #859, opened 2019-11-27
[wip] fix: flip flicker

atomiks:
The bug: https://i.imgur.com/a397z83.mp4
Issue for v1: https://github.com/FezVrasta/popper.js/issues/720

When the popper changes margins depending on the placement, its size can change, sending the flip modifier into an infinite loop where it thinks it will fit, but doesn't, because it doesn't know ahead of time that the margin is there when it computes the offsets.


atomiks:
I worked something out 🎉 

In the `detectOverflow` modifier, we need to store overflow data about each different placement so that `flip` can know the correct overflows (which can change for each different placement). The only way `flip` can know is if the overflows have been tested ahead of time by `detectOverflow`.

The flicker is gone, but there is a 1 update jerk (once it switches to a new placement) remaining. Once we solve that we can refactor what I've done to actual decent code 😬

atomiks:
I came up with something much better and more performant 🙌 

 There is still the jitter so we need to call `.forceUpdate()` to fix it, which isn't really a problem but it's a hack, it would be nice if we could figure out how to fix that properly.

I've had to remove `auto` placements for this to work appropriately regarding types, but I don't think they were even implemented yet anyway? Maybe the user should implement auto placements themselves?

atomiks:
I think in the docs we will recommend not using conditional margins/paddings based on placement, and instead recommend `offset`:

```js
  {
    name: 'offset',
    options: {
      // I think this should allow a plain value as well, not functions
      offset: [10, 0] // distance: 10px
    }
  }
```

This prevents the need for handling this using hacks

Issue #617, opened 2018-04-17
Popper.js v2


FezVrasta:
If anyone wants to contribute to the Popper.js v2 rewrite please do so sending PRs against the [`next`](https://github.com/FezVrasta/popper.js/tree/next) branch.

It's a completely new codebase, I think the current codebase is too messy to continue to work on it, it started as a little proof of concept and ended up being a full fledged library, I think it deserves a complete rewrite.

The idea is to copy the good code from the old version and write new one to make the new code more stable.

I'd like to make the library even more performant by better leveraging to an internal cache the modifiers can use to get most of the values they need that have been already read from the DOM.

The only decisions I made so far are:

- prettier to format the code
- Flow to type check it
- New system to order modifiers [WIP] ([discussion](https://spectrum.chat/?t=f04a2b1d-6375-492a-884a-5b94bd65d984))

To do:

- CI
  - [x] add a built step (Rollup + Babel)
  - [x] add a CI setup
  - [x] add unit tests (Jest)
  - [x] add functional/integration cross-browser tests (Karma?)
- Core
  - [x] rewrite the positioning engine to work on all the supported cases (and more?)
  - [x] rewrite the modifiers API
- Modifiers
  - [x] computeStyles
  - [x] applyStyles
  - [x] arrow
  - [x] preventOverflow
  - [x] flip
  - [x] offset/shift (merge them together)
  - [ ] other modifiers will be added later, if anyone wants them please send a PR to implement them
- Miscellaneous
  - [x] Implement the changes labelled as [`BREAKING CHANGE`](https://github.com/FezVrasta/popper.js/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22%23+BREAKING+CHANGE%22)

If you'd like to contribute please write here so others know who's working on what.

lukebatchelor:
Commenting here to follow progress.

Exciting stuff!

inomdzhon:
Do we really need set `undefined` for property in [getElementClientRect.js](https://github.com/FezVrasta/popper.js/blob/2ac44d0d640f1732071c200ca8f3322e7fbedff7/src/dom-utils/getElementClientRect.js#L6) when no element?

Why we can't return `0` or make `element` required, that is can't be `null`.

And we need set return types there. `getBoundingClientRect` return `ClientRect | DOMRect`.

FezVrasta:
[...]

PR #861, opened 2019-11-29
fix: positioning logic for transforms and SVG elements

atomiks:
This fixes bug 6 in #617

Use `getBoundingClientRect` instead of `offset` properties, this handles transforms and SVG elements. 

It's possible summing the common scrollParents scroll is redundant in relation to logic outside of the function, but this seems to mimic `offset{Left,Top}` while using `getBoundingClientRect`.


src/dom-utils/getElementClientRect.js
```diff
@@ -1,13 +1,29 @@
...
-  const x = element.offsetLeft;
+// Returns the width, height and offsets of the provided element relative to the
+// offsetParent
+export default (element: Element): Rect => {
+  const rect = getBoundingClientRect(element);
+  const offsetParent = getOffsetParent(element);
+  const scrollParentsScroll = getScrollSum(listScrollParents(element));
+  const windowScroll = getWindowScroll(element);
+  const scrollLeft = scrollParentsScroll.scrollLeft - windowScroll.scrollLeft;
+  const scrollTop = scrollParentsScroll.scrollTop - windowScroll.scrollTop;
```
atomiks: Should `getScrollSum` handle `window` as well? It uses `scrollLeft` / `scrollTop` in the reducer but that doesn't work for `window` (despite window being part of the array returned from `listScrollParents`

atomiks:
I think it handles everything properly now. Review @FezVrasta?

FezVrasta:
That's great, thanks so much for the PR.

Would you please add a functional test for this?

You just need to add a new visual test, and then create a test file in `tests/functional` where you do something like:

```js
/**
 * @jest-environment jest-environment-puppeteer
 * @flow
 */
import { screenshot } from '../utils/puppeteer.js';

it('should position the popper on the right', async () => {
  const page = await browser.newPage();
  await page.goto('http://localhost:5000/your-visual-test.html');

  expect(await screenshot(page)).toMatchImageSnapshot();
});
```

Then, running `BROWSER=chromium jest tests/functional` will run the functional tests, and a new screenshot will be added to `__image_snapshots__`.

atomiks:
I don't really know what's happening with the functional tests :-(

When I visit the visual tests manually, everything works as expected in Chrome. And ironically, Firefox is now working fine lol

atomiks:
Okay now everything is working except for the SVG test 🙄. Is there anything that is making them flaky (_apart_ from the pixel diffs since that actually makes sense)? I don't get how they can run and work and other times just not. What is happening?

Anyway, we need to increase the pixel diff threshold a bit:

### Failed

> Expected image to match or be a close match to snapshot but was 0.005% different from snapshot (24 differing pixels).

img

FezVrasta:
Could we get rid of the round shape and use a rectangle instead?

atomiks:
Finally a green build with the rectangle 🥳, good idea to prevent any pixel differences

tests/visual/svg.html
```diff
@@ -0,0 +1,38 @@
...
+    background-color: rebeccapurple;
+    box-shadow: inset 0 0 0 1px black;
+  }
+</style>
+
+<svg
+  id="reference"
+  viewBox="0 0 100 100"
+  xmlns="http://www.w3.org/2000/svg"
+  fill="red"
```
FezVrasta: Shouldn’t we fill the rect rather than the whole svg box? Could we maybe make the svg box larger than the rect so that we make sure we are positioning the popper relative to the rect and not just coincidentally to the svg box?

atomiks: It depends what node is being passed. We'd need to put `id="reference"` on the `<rect>` node in that case.

We should add a bunch more SVG elements?

atomiks:
img

🙄

FezVrasta:
You can increase the threshold for a specific test.

```
expect(image).toMatchImageSnapshot({
    failureThreshold: 0.01
  });
```

atomiks:
The typechecking for that has red squiggles. Is it because of Flow?

FezVrasta:
I fixed the types on `next`

PR #862, opened 2019-11-30
feat: arrow modifier

atomiks:
I hope this approach is good, need some help with the types

### DIFF
--- a/src/modifiers/applyStyles.js
+++ b/src/modifiers/applyStyles.js
@@ -13,6 +13,11 @@ export function applyStyles(state: State) {
     const attributes = data.attributes[name] || {};
     const element = state.elements[name];
 
+    // arrow is optional
+    if (!element) {
+      return;
+    }
+
     Object.assign(element.style, style);
 
     Object.entries(attributes).forEach((args: [string, any]) =>
@@ -41,6 +46,11 @@ export function onDestroy(state: State) {
       }),
       {}
     );
+
+    // arrow is optional
+    if (!element) {
+      return;
+    }
 
     // Flow doesn't support to extend this property, but it's the most
     // effective way to apply styles to an HTMLElement
--- a/src/modifiers/arrow.js
+++ b/src/modifiers/arrow.js
@@ -58,12 +58,19 @@ export function arrow(state: State, options?: Options = {}) {
 
   const startDiff = popperOffsets[axis] - state.measures.reference[axis];
 
+  // Only center to reference if the popper is longer than reference along the
+  // axis
+  const centerToReference =
+    state.measures.popper[len] > state.measures.reference[len]
+      ? endDiff / 2 -
+        startDiff / 2 -
+        commonTotalScroll[axis === 'y' ? 'scrollTop' : 'scrollLeft']
+      : 0;
+
   let center =
     state.measures.popper[len] / 2 -
     arrowElementRect[len] / 2 +
-    endDiff / 2 -
-    startDiff / 2 -
-    commonTotalScroll[axis === 'y' ? 'scrollTop' : 'scrollLeft'];
+    centerToReference;
 
   // Make sure the arrow doesn't overflow the popper if the center point is
   // outside of the popper bounds
--- a/src/modifiers/arrow.js
+++ b/src/modifiers/arrow.js
@@ -37,7 +37,6 @@ export function arrow(state: State, options?: Options = {}) {
   const isVertical = ['left', 'right'].includes(basePlacement);
   const axis = getMainAxisFromPlacement(basePlacement);
   const len = isVertical ? 'height' : 'width';
-  const side = isVertical ? 'top' : 'left';
 
   const arrowElementRect = addClientRectMargins(
     getElementClientRect(arrowElement),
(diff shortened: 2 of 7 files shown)


src/types.js
```diff
@@ -30,6 +30,7 @@ export type State = {|
   elements: {|
     reference: HTMLElement,
     popper: HTMLElement,
+    arrow: HTMLElement,
```
FezVrasta: `arrow?: HTMLElement`

src/modifiers/arrow.js
```diff
@@ -0,0 +1,82 @@
...
+  // Make sure the arrow doesn't overflow the popper if the center point is
+  // outside of the popper bounds
+  center = within(
+    0,
+    center,
+    state.measures.popper[len] - arrowElementRect[len]
+  );
+
+  // Flow: How to use computed property like {[axis]: center}?
+  state.modifiersData.arrowOffsets = { x: 0, y: 0 };
```
FezVrasta: The property should have the same name of the modifier

src/types.js
atomiks: We'll need to account for that in `applyStyles` I guess (I did try that)

src/modifiers/arrow.js
atomiks: Hmm so should we separate it then like `popperOffsets`? As `computeStyles` assumes `arrowOffsets` name. IIRC some people didn't like how variation placements centered the arrow relative to the reference (they wanted it to be at the start/end always) - they could adjust it there?

src/modifiers/applyStyles.js
```diff
@@ -13,6 +13,11 @@ export function applyStyles(state: State) {
     const attributes = data.attributes[name] || {};
     const element = state.elements[name];
 
+    // arrow is optional
+    if (!element) {
```
FezVrasta: Why is this needed? `state.elements.arrow` should not exist if the arrow modifier didn't run.

src/modifiers/arrow.js
```diff
@@ -0,0 +1,89 @@
...
+  // outside of the popper bounds
+  center = within(
+    0,
+    center,
+    state.measures.popper[len] - arrowElementRect[len]
+  );
+
+  // Flow: How to use computed property like {[axis]: center}?
+  state.modifiersData.arrow = {};
+  state.modifiersData.arrow[axis] = center;
```
FezVrasta:
```js
  state.modifiersData.arrow = {
    x: 0,
    y: 0,
    [(axis: string)]: center,
  };
```

src/modifiers/applyStyles.js
atomiks: It's for the types to stop complaining

FezVrasta: 🤦‍♂️

src/modifiers/arrow.js
```diff
@@ -0,0 +1,89 @@
...
+
+  // Make sure the arrow doesn't overflow the popper if the center point is
+  // outside of the popper bounds
+  center = within(
+    0,
+    center,
+    state.measures.popper[len] - arrowElementRect[len]
+  );
+
+  // Prevents breaking syntax highlighting...
```
FezVrasta: Are you using [Babel JavaScript](https://marketplace.visualstudio.com/items?itemName=mgmcdermott.vscode-language-babel) syntax? That supports this kind of annotations.

atomiks: I don't really want to have to change syntax highlighter packages, I already have to turn off a setting while developing with Flow in VSCode :-(

atomiks:
A good thing about this approach is it won't ever affect the Popper's position. In v1, the arrow can affect the Popper's position which causes weird bugs :-(

FezVrasta:
Shouldn't the arrow attempt to stay centered within the popper element rather than the reference?

image

atomiks:
No the whole point is that it stays centered relative to the reference element, that's how it works in v1. 😅 

example

FezVrasta:
But if the reference is overflowing, the other way around doesn't look nicer? 🌈

atomiks:
Yeah but how would that work? It's centered to the reference until it starts overflowing, then "jerks" down suddenly to center in the popper? 

Seems complex to handle regardless. :-\ I think the current v1 behavior has been good enough anyway

atomiks:
Thought about that a bit more. I'm guessing it would only work if the popper is smaller than the reference along that axis - it would look better to keep the arrow centered when the reference is overflowing 🤔

atomiks:
We need to know the `tether` value from `preventOverflow` modifier. How do we access options from other modifiers?

As we know, `tether` allows the popper to overflow if it's about to detach from the reference element.

If we envision the following values:

```js
{
  // the popper never overflows, always stays visible
  tether: false,
  // default is "center" - this should work well for most cases
  tether: true,
  // popper starts overflowing once it reaches the edge of the ref
  // this works well if the ref is like a line of text or something
  tether: "edge",
  // popper starts overflowing once the surfaces of the boxes touch
  // this is how it currently works
  tether: "surface"
}
```

The `arrow` modifier will need to know these to decide what to do.

FezVrasta:
I added a functional test for it, and merged. I will improve the behavior later.

PR #863, opened 2019-12-01
feat: new tether options

atomiks:
We can discuss which names or API are best. The comments explain how each value works.

Video demos:
`false`: https://i.imgur.com/fmcJ0Bj.mp4
`"center"`: https://i.imgur.com/IUjTVMI.mp4
`"edges"`: https://i.imgur.com/seoUz47.mp4
`"surfaces"`: https://i.imgur.com/ZWJ4YuC.mp4

### DIFF
--- a/src/enums.js
+++ b/src/enums.js
@@ -15,6 +15,11 @@ export const start: 'start' = 'start';
 export const end: 'end' = 'end';
 export type VariationPlacement = typeof start | typeof end;
 
+export const center: 'center' = 'center';
+export const edges: 'edges' = 'edges';
+export const surfaces: 'surfaces' = 'surfaces';
+export type Tether = false | typeof center | typeof edges | typeof surfaces;
+
 export type Placement =
   | 'auto'
   | 'auto-start'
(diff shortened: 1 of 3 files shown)


FezVrasta:
That’s so cool!! Thank you so much for all your help.

Shouldn’t true and false be reversed though?

atomiks:
> Shouldn’t true and false be reversed though?

`false` means it's not tethered at all (can detach), which the videos show. Maybe you accidentally reversed the links when viewing the vids? 😄 

> Also, probably it would be easier to understand if we replaced true with another enum. So that we give names to all the “on” behaviors, and leave false just for the off

Yeah, agreed. I like `"center"` more than `true`. Also, it would be cool if they could supply a number as the offset too, but that might need a function that can passes the state/ current offsets so it's more useful.

atomiks:
I'll also have to work out how to handle the arrow. I think `edges` in that case will mean "center of the arrow", which is offset up a bit from the bottom.

FezVrasta:
Please let’s leave the arrow to me, I still need to do some work on its modifier. We could merge this PR once you rename true to center

atomiks:
Wait there's a bug with `tether` names conflicting I think. 🤦‍♂

FezVrasta:
Is this failure expected with the new behavior?

image

atomiks:
Yes, since the default tether is `center` not `surfaces` anymore, we need to update that snapshot

FezVrasta:
Merged manually

PR #864, opened 2019-12-02
fix: flip should use viewport rect

atomiks:
This fixes bug 8. in #617 

I'm not sure how `flip` could use "clippingParent" or if it even should? Also in v1, you could specify your own boundary for flip but now it's forced to use viewport. Does it even make sense to use anything else - is there a case where it's not suitable?

I'm not good with the functional test thingies if you could add that would be good 🙏

### DIFF
--- a/src/dom-utils/getDocumentRect.js
+++ b/src/dom-utils/getDocumentRect.js
@@ -2,13 +2,19 @@
 import getRectRelativeToOffsetParent from './getRectRelativeToOffsetParent';
 import getWindow from './getWindow';
 import getDocumentElement from './getDocumentElement';
+import getWindowScroll from './getWindowScroll';
 
 export default (element: HTMLElement) => {
   const win = getWindow(element);
+  const winScroll = getWindowScroll(element);
   const documentRect = getRectRelativeToOffsetParent(
     getDocumentElement(element)
   );
+
   documentRect.height = Math.max(documentRect.height, win.innerHeight);
   documentRect.width = Math.max(documentRect.width, win.innerWidth);
+  document.x = -winScroll.scrollLeft;
+  document.y = -winScroll.scrollTop;
+
   return documentRect;
 };
--- a/src/dom-utils/getDocumentRect.js
+++ b/src/dom-utils/getDocumentRect.js
@@ -13,8 +13,8 @@ export default (element: HTMLElement) => {
 
   documentRect.height = Math.max(documentRect.height, win.innerHeight);
   documentRect.width = Math.max(documentRect.width, win.innerWidth);
-  document.x = -winScroll.scrollLeft;
-  document.y = -winScroll.scrollTop;
+  documentRect.x = -winScroll.scrollLeft;
+  documentRect.y = -winScroll.scrollTop;
 
   return documentRect;
 };
--- a/src/dom-utils/getViewportRect.js
+++ b/src/dom-utils/getViewportRect.js
@@ -1,0 +1,12 @@
+// @flow
+import getWindow from './getWindow';
+
+export default (element: HTMLElement) => {
+  const win = getWindow(element);
+  return {
+    width: win.innerWidth,
+    height: win.innerHeight,
+    x: 0,
+    y: 0,
+  };
+};
--- a/src/enums.js
+++ b/src/enums.js
@@ -20,6 +20,13 @@ export const edges: 'edges' = 'edges';
 export const surfaces: 'surfaces' = 'surfaces';
 export type Tether = false | typeof center | typeof edges | typeof surfaces;
 
+export const clippingParent: 'clippingParent' = 'clippingParent';
+export const viewport: 'viewport' = 'viewport';
+export type OverflowArea =
+  | HTMLElement
+  | typeof clippingParent
+  | typeof viewport;
+
 export type Placement =
   | 'auto'
   | 'auto-start'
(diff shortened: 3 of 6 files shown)


src/dom-utils/getViewportRect.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import getWindow from './getWindow';
+
+export default (element: HTMLElement) => {
```
FezVrasta: I’m quite sure we already have a utility that does the same

atomiks: I checked but it doesn't seem like it

FezVrasta: Probably `getDocumentRect` will work.

atomiks: `getDocumentRect` isn't the same as viewport as you're `Math.max`ing it

FezVrasta: image

They should be the same thing, I think I'm Math.max-ing window vs documentElement because in some cases the documentElement can be smaller than the window (for example, on a basic empty page)

atomiks: But the code was already using `documentRect` 😅 

```js
  documentRect.height = Math.max(documentRect.height, win.innerHeight);
  documentRect.width = Math.max(documentRect.width, win.innerWidth);
```

If documentRect is taller, then it uses that instead of the viewport

FezVrasta: I don't get it, this code to me returns the viewport, not the whole page/document height.

atomiks: That only happens when the documentRect is smaller. The code in your screenshot isn't the same as the code we're actually using. If you go to the flip demo and resize, you'll see that documentRect doesn't change when the console window is resized up

atomiks: I'm assuming you mean this right?:

```js
  // this works
  const viewportClientRect = rectToClientRect(
    getViewportRect(options.boundaryElement)
  );

  // this doesn't work properly
  // (based on real document height, not client height)
  const viewportClientRect = rectToClientRect(
    getDocumentRect(options.boundaryElement)
  );
```

atomiks:
If forced viewport is unsuitable in some cases we could add a `viewportElement` option in the future for `detectOverflow`

FezVrasta:
I think we need that namespaced modifiers feature I was talking about with you on Twitter.

So that the user is able to configure each one independently, and the bundle size won't increase.

atomiks:
So would it work like:

```js
state.modifiersData.detectOverflow = {
  flip: {...},
  preventOverflow: {...}
};
```

FezVrasta:
No, the whole modifier will run for each namespace. Each run will not be aware of other instances of itself.

atomiks:
Ah I see. 

Should that be done on this branch or a new one? (If you could do it since I'm not sure of how it will work properly)

FezVrasta:
A new branch, I need to think about it a bit more first, we need to make sure this new API covers all our needs.

atomiks:
There was something odd about `documentRect`. It wasn't calculating anything properly (the document, nor the viewport).

I've fixed it so it calculates the real document rect, while viewport function will be for the viewport.

src/dom-utils/getDocumentRect.js
```diff
@@ -2,11 +2,17 @@
...
   const win = getWindow(element);
+  const winScroll = getWindowScroll(win);
   const documentRect = getElementClientRect(getDocumentElement(element));
+
   documentRect.height = Math.max(documentRect.height, win.innerHeight);
   documentRect.width = Math.max(documentRect.width, win.innerWidth);
+  documentRect.x = -winScroll.scrollLeft;
+  documentRect.y = -winScroll.scrollTop;
+
   return documentRect;
```
Joelkang: Isn't this really just `getWindow(element).document.body.getBoundingClientRect()`?

(it would really only not work if the body has transforms, which seems highly unlikely)

atomiks: It is (`<html>` instead), but we are using the relative to offsetParent function, and returns a `rect` (width, height, x, y) not `clientRect` (which includes all properties)

Joelkang: Oh yea, you're right--I was thinking since the `<head>` is not rendered anyway this is the same, but `<html>` can indeed be taller than `<body>`. lol sigh

atomiks:
I think I came up with something decent

We only really need to know two things:

```js
type Options = {
  clippingArea: HTMLElement | 'viewport',
  visibleArea: HTMLElement | 'viewport',
};
```

`clippingArea` is used by `preventOverflow`, while `visibleArea` is used by `flip`. These are pretty descriptive, and they can also each be the `viewport` enum if the user wants that behavior.

FezVrasta:
Can I merge this?

atomiks:
Check now? We can hardcode which `detectOverflow` property to use, as long as `detectOverflow` can provide enough information for us, so I added the `clippingParent` and `viewport` enums which should cover all cases. The user can therefore configure `flip` and `preventOverflow` implicitly by specifying which areas to use via the `clippingArea` and `visibleArea` options

Joelkang:
Sorry to be a bother again but can you publish a new version with this @FezVrasta ? 🙏

atomiks:
@Joelkang  are you using these unstable versions in prod? 😆

FezVrasta:
2.0.0-next.12

Joelkang:
@atomiks kinda yea, but we're doing a slow staggered, alpha then beta release over the new year. My company builds an embeddable script that runs inside an iframe so that our polyfills/styles don't leak into the host document, but then we render tooltips in the host. v1 causes a lot of problems because of the assumption that `window` is always the host window (when in our case it winds up being the iframe's contentWindow). 

If we run up into many problems as we release we might fork and publish our own fork for consumption, but since the effort to use popper@next was very low, we figured why not try it for now and see how this works!

PR #865, opened 2019-12-02
feat: virtual elements

atomiks:
```js
const virtualElement = {
  // positioning context, real DOM element
  contextElement: reference,
  // custom position
  getBoundingClientRect: () => ({
    width: 200,
    height: 150,
    top: 100,
    right: 300,
    bottom: 250,
    left: 100,
  }),
};
```

### DIFF
--- a/src/dom-utils/getBoundingClientRect.js
+++ b/src/dom-utils/getBoundingClientRect.js
@@ -1,6 +1,9 @@
 // @flow
-import type { ClientRectObject } from '../types';
+import type { ClientRectObject, VirtualElement } from '../types';
 
-export default (element: Element): ClientRectObject => {
-  return JSON.parse(JSON.stringify(element.getBoundingClientRect()));
+export default (element: Element | VirtualElement): ClientRectObject => {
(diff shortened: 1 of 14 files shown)


atomiks:
I don't understand how to use Flow's `%checks` thing for type refinement :| it doesn't work for me.

In TS you can do:

```ts
function isSomething(value): value is Something {
  return value.hasOwnProperty('something');
}
```

I wanted to use this to replace `instanceof` checks but don't know how to do it

FezVrasta:
`Element` is an inexact type, so it may have the `contextElement` property assigned for whatever reason.

This means that if you check for the `contextElement` property existence, it won't tell you if the element you are checking is actually a `VirtualElement` or a generic `Element` with a `contextElement` property.

You can test for `element` to be instance of `Element` though, because this is type safe.

```typescript
export default (element: Element | VirtualElement): Element =>
  element instanceof Element ? element : element.contextElement;
```

Unfortunately this still has the issue of iframes.

I manually merged the PR and changed all the `instanceof` checks to use `getWindow` to get the instance of the element's window.

PR #866, opened 2019-12-03
feat: namespace API for modifiers

FezVrasta:
I recycled this PR for the namespace API, since most of the changes were in common.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,7 +9,7 @@ import addClientRectMargins from './dom-utils/addClientRectMargins';
 
 // Pure Utils
 import unwrapJqueryElement from './utils/unwrapJqueryElement';
-import orderModifiers from './utils/orderModifiers';
+import processModifiers from './utils/processModifiers';
 import expandToHashMap from './utils/expandToHashMap';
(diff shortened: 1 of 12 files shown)


atomiks:
Kinda seems like over-engineering something that should be kept simple to me, but 🤷‍♂

FezVrasta:
Nevermind, easier solution:

```
import detectOverflow from './detectOverflow';

export default {
  ...detectOverflow,
  name: 'detectOverflowFlip'
}
```

And we get "namespaces" for free.

atomiks:
`flip` accesses `detectOverflow` data like this:

```js
const overflow = state.modifiersData.detectOverflow;
```

We can add the `name` to make it dynamic, but still, how would we get `flip` to access `detectOverflowFlip` namespace instead?

```js
const detectOverflowFlip = {
  ...detectOverflow, 
  name: 'detectOverflowFlip',
  options: {
    boundaryElement: someElement,
  }
};

createPopper(ref, pop, {
  modifiers: [detectOverflowFlip]
});
```

PR #867, opened 2019-12-03
feat: use closure instead of class

FezVrasta:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Generally, for CDN users, you'll be using the fully-featured `umd` file:
 ```html
 <script src="https://unpkg.com/@popperjs/core@next"></script>
 <script>
-  // Now the script is loaded, you can use the `Popper` class!
+  // Now the script is loaded, you can use the `createPopper` function!
 </script>
 ```
 
@@ -92,7 +92,7 @@ setup, but is worth it.
 In your app, you can do the following:
 
 ```js
-// Import the core class
+// Import the generator function
 import { popperGenerator } from '@popperjs/core';
 
 // Import the features you need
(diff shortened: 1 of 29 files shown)


src/index.js
```diff
@@ -65,174 +144,98 @@ export default class Popper {
...
-    // Run `onDestroy` modifier methods
-    this.state.orderedModifiers.forEach(
-      ({ onDestroy, enabled }) => enabled && onDestroy && onDestroy(this.state)
-    );
-  }
+    return {
+      forceUpdate,
+      update,
+      destroy,
+    };
```
atomiks: I think this should be stored as a variable and methods are accessed like `instance.update()` - this allows public accessors to monkey-patch the methods if necessary

FezVrasta: Good idea! Done

README.md
```diff
@@ -104,12 +104,14 @@ import {
...
+  defaultModifiers: [
+    detectOverflow,
+    preventOverflow,
+    computeStyles,
+    applyStyles,
+  ],
+});
 ```
 
 Now you can use the `Popper` class with _only_ the features you want. For
```
atomiks: Says `class`

FezVrasta:
I removed the default export, since there can be problems when mixing default and named exports on legacy module systems. So now it's `import { createPopper, popperGenerator } from '@popperjs/core';`

atomiks:
I like the capitalized version. Seems fien to me, although I kind of do like having a default export a bit more.

FezVrasta:
I'm avoiding the mixed export because in some systems you will end up with stuff like this:

```
const createPopper = require('@popperjs/core').default;
```

and:
```
const { default: createPopper, popperGenerator } = require('@popperjs/core');
```

atomiks:
Yeah tippy has that :\, I don't really care about the CJS version because tree-shaking doesn't work anyway (in webpack). It's a last-resort so ugly things like that don't matter too much imo

FezVrasta:
What about UMD? I'm not even sure it can support mixed exports.

atomiks:
What I did was attach it to the function namespace, it's far cleaner to use for UMD:

```js
createPopper.popperGenerator = popperGenerator;
```

FezVrasta:
Isn't a named export clearer? 🤨

atomiks:
I do think `Popper.createPopper` is pretty clean. 

In Tippy it would've been `Tippy.tippy` which is weird since it's a generic constructor instead of one that creates a single instance. In this case it works better 👍

FezVrasta:
Ok cool, then let's stick with named exports 🤘

PR #868, opened 2019-12-03
feat: basic fixed strategy support [2.0]

Joelkang:
For my use case I use the `fixed` strategy since I mount the poppers inside of a portal (to prevent z-index) issues. 

I initially went down this path, but it's beginning to seem that the strategy might need to flow down into `getElementClientRect` instead, which seems like a much bigger feat change


FezVrasta:
Thanks so much for the PR!

Is it really that easy? 🧐

May you try to switch to the `fixed` strategy in the other visual tests and see if they still behave properly?

atomiks:
It's not working with the scrolling container tests

atomiks:
Our current `getElementClientRect` behavior works for `fixed` for all scrolling tests except for `offset-parent.html`. 

I managed to solve it like this for `fixed`:

- Don't take subtract `directOffsetParent` rect
- Include the first `offsetParent` in the `ancestorOffsetParents` array

```js
export default (
  element: Element | VirtualElement,
  // pass in the strategy
  isFixed: boolean = false
): Rect => {
  // ...
  // I think we only need one window stored here to check
  const win = getWindow(unwrappedElement);

  const directOffsetParentRect =
    directOffsetParent instanceof win.Element && !isFixed
      ? getBoundingClientRect(directOffsetParent)
      : { left: 0, top: 0 };

  const ancestorOffsetParents = isFixed ? [directOffsetParent] : [];
  let currentOffsetParent = directOffsetParent;

  while (currentOffsetParent instanceof win.Element) {
    currentOffsetParent = getOffsetParent(currentOffsetParent);
    ancestorOffsetParents.push(currentOffsetParent);
  }

  // ...
};

```

atomiks:
Should `fixed` be the default positioning strategy? What are the differences besides the popper not being affected by clipping parents (which seems like a good thing)? That behavior may be wanted in some cases where the user can choose `absolute`.

FezVrasta:
absolute is generally more preformant, in the simplest cases it doesn’t need to reposition the popper on each scroll.

We could automatically change strategy based on the detected environment though.

Joelkang:
Actually the more I think about it, the more it's unclear to me what `getElementClientRect does--getBoundingClientRect already gets the position relative to the viewport, taking into account scroll, while for offsets relative to the parent, there's `offsetLeft` and `offsetTop`. What does the summing of the ancestral scroll do?

atomiks:
`offsetTop` and `offsetLeft` don't work when the element has transforms or for SVG elements

FezVrasta:
`getBoundingClientRect` returns the offsets relative to the viewport, not to the whole document.

FezVrasta:
I'm going to close this since I just merged #868 that does the same thing, thank you again for the help with the initial implementation!!

Joelkang:
thanks so much @FezVrasta  and @atomiks !

PR #870, opened 2019-12-04
fix: prevent arrow from going outside ref bounds

atomiks:
In the case where we want to center the arrow if it's smaller than the reference, we don't want the arrow to go outside of its bounds

https://i.imgur.com/jfHfmra.mp4

There seems to be a problem when the popper is larger than the reference. When it reaches the "diff" point in size, it starts "accelerating" too fast (2x?) toward the center of the reference. Any help fixing that would be great 🙏


atomiks:
There's also this problem: https://i.imgur.com/jpoFI3q.mp4

The tether option should decide to activate taking into account the arrow size, as the arrow goes outside of the reference bounds

atomiks:
I think I fixed both issues

- There's a `$FlowFixMe` that needs fixing
- There are some ambiguous cases like if the arrow is larger than the popper and/or reference on the axis, what happens?
- We need visual tests covering various cases for this as the conditions have gotten quite complex

The behavior overall is now:

- If the popper is larger than the reference, the arrow will stay centered to the reference always, but it will never overflow the popper bounds to stay centered to it
- If the popper is smaller than the reference, the arrow will stay centered to the popper unless the arrow is about to leave the reference bounds, in which case it will stay relative to the reference edge
- The tether `edges` option includes the arrow in the calculation to decide when to activate

src/modifiers/preventOverflow.js
```diff
@@ -76,14 +78,27 @@ export function preventOverflow({
...
+    // For the "edges" value, we need to include the arrow in the calculation
+    // so the arrow doesn't go outside the reference bounds
+    const arrowElementRect =
+      state.elements.arrow && tether === edges
+        ? addClientRectMargins(
+            getRectRelativeToOffsetParent(state.elements.arrow),
+            // $FlowFixMe
+            state.elements.arrow
+          )
+        : { width: 0, height: 0 };
```
FezVrasta:
```js
    const arrowElement = state.elements.arrow;
    const arrowElementRect =
      arrowElement && tether === edges
        ? addClientRectMargins(
            getRectRelativeToOffsetParent(arrowElement),
            arrowElement
          )
        : { width: 0, height: 0 };
```

FezVrasta:
> Also: if the popper is only slightly larger or slightly smaller than the reference (by a couple px), why or how do we decide which one of 1 or 2 to use? Does it really make sense to even do 2 ever? I know we thought it looked better but it's highly dependent on the setup as to which one appears better, which is getting way too complex to code at that point.

Could you propose an alternative behavior that would cover the most common cases while looking good and simple to implement?

I suppose if people need a more advanced arrow positioning they could swap out the default arrow modifier and use a custom one.

atomiks:
> Could you propose an alternative behavior that would cover the most common cases while looking good and simple to implement?

That would be the simple `centerToReference` behavior we had before, which works like v1

FezVrasta:
It's up to you, you are the arrows expert here 😂

atomiks:
OK I reverted all that logic, but kept the new tether logic.

We can figure out the ambiguous edge cases later 😭

PR #871, opened 2019-12-04
fix(debounce): simplify debounce to always return the last resolved value

Joelkang:
In v1 we had the `onUpdate` and `onCreate` callbacks. I previously used this to read the state so that I can update my tooltip (which is rendered in react).

Given that now we no longer have these callbacks, we should be able to do `tooltip.update().then(lastRenderedState => ...)`.

However, the debounce function was swallowing the last resolved value of the debounced function. So let's fix this, simplifying the implemementation of debounce.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -132,9 +132,9 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
     // debounced, so that it only runs at most once-per-tick
     instance.update = debounce(
       () =>
-        new Promise<State>(success => {
+        new Promise<State>(resolve => {
           instance.forceUpdate();
-          success(state);
+          resolve(state);
         })
     );
 
--- a/src/utils/debounce.js
+++ b/src/utils/debounce.js
@@ -1,15 +1,16 @@
 // @flow
 export default function microtaskDebounce(fn: Function) {
-  let called = false;
-  return () =>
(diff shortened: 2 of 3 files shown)


FezVrasta:
Thanks!

FezVrasta:
Should we add back he `onUpdate` and `onCreate` callbacks? 

I think the `onCreate` could be replaced by a promise `createPopper(...).then(...)`.

But the `onUpdate` seems something we need to reimplement.

atomiks:
I feel like the encourage mutation outside of the modifiers lifecycle when everything should be implemented in modifiers, but I could be wrong

FezVrasta:
The only alternative at the moment is to define a custom modifier like this:

```
createPopper(x, y, {
  modifiers: [
    {
      name: 'lifecycleHook',
      fn: ({ state }) => { 
        if (state.modifiersData.lifecycleHook.created === false) {
		  state.modifiersData.lifecycleHook.created = true;
          // onCreate
        } else {
          // onUpdate
        }
        return state;
      },
      data: { created: false },
    }
  ]
});
```

Which isn't too bad to be honest.

Joelkang:
@FezVrasta do you mind publishing to npm a new build with these changes? thanks!

FezVrasta:
2.0.0-next.11

PR #872, opened 2019-12-04
feat: change modifiers signature

FezVrasta:
Changes the modifiers signature to accept an object rather than multiple arguments.

This should make it easier to extend the API if needed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -119,10 +119,10 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
           continue;
         }
 
-        const { fn, enabled, options } = state.orderedModifiers[index];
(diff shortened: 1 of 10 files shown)


src/index.js
```diff
@@ -122,7 +122,7 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
         const { fn, enabled, options } = state.orderedModifiers[index];
 
         if (enabled && typeof fn === 'function') {
-          state = fn((state: State), options);
+          state = fn({ state, options });
```
atomiks: Should we pass `name` for the dynamic name setting modifiersData I mentioned on Twitter?

PR #873, opened 2019-12-05
feat: move event listeners to modifier

FezVrasta:
With this we lose the `instance.enableEventListeners` method.

We have 3 options:

1. add it back (kinda hack-ish)
2. tell users to destroy + create when they need different options
3. create a global method to update an instance options

3 is something some people asked in the past, but considered how lightweight the popper instantiation is, I don't know if it's worth it.

(with this PR, popper-base goes below the 2kB (1.92 kB))


FezVrasta:
Yes that's right.

I think option 3 is the best one, we should have something like `instance.updateOptions(...)`

FezVrasta:
I was experimenting with option 3, it's a very small code change actually.

PR #874, opened 2019-12-05
Update arrow.js

XhmikosR:
Remove superfluous argument passed to `parseFloat()`

@FezVrasta found with https://lgtm.com/projects/g/popperjs/popper.js/?mode=list which is now owned by GitHub so you could add it to the repo :) Personally I just disable PR comments only.


FezVrasta:
Thank you!

PR #875, opened 2019-12-06
fix: add offset to detectOverflow, run before flip and preventOverflow

atomiks:


src/modifiers/offset.js
```diff
@@ -52,6 +63,6 @@ export function offset({ state, options = {} }: ModifierArguments<Options>) {
 export default ({
   name: 'offset',
   enabled: true,
-  phase: 'main',
+  phase: 'beforeMain',
```
FezVrasta:
```suggestion
  phase: 'main',
  optionalRequires: ['detectOverflow']
```

?

PR #876, opened 2019-12-07
fix: run arrow modifier after preventOverflow

atomiks:
After a nightmarish bug hunt I found out why the arrow randomly broke after a build commit change - it was being placed before `preventOverflow`, but it relies on `preventOverflow`'s mutations to `popperOffsets` to work.

However, this uncovered a separate bug: `preventOverflow` relies on `state.elements.arrow` for its `edges` tether option, but it doesn't exist until after load. So the first load will have the arrow overflow the ref bounds if the popper is coincidentally placed in a location where that will happen, since it's being placed before `arrow`.

I'm not sure if it's possible to allow `arrow` to run before `preventOverflow` though, because it needs to know the real `popperOffsets` value that gets altered due to overflow prevention... D; so maybe the arrow element shouldn't be specified in the modifier, rather like this maybe:

```js
createPopper({
  reference,
  popper,
  arrrow,
  options: {...}
});
```

### DIFF
--- a/src/modifiers/arrow.js
+++ b/src/modifiers/arrow.js
@@ -89,4 +89,5 @@ export default ({
   phase: 'main',
   fn: arrow,
   requires: ['popperOffsets'],
+  optionallyRequires: ['preventOverflow'],
 }: Modifier<Options>);


atomiks:
Also maybe the two requires should be :

`explicitRequires`: this needs to read the modifier's data from `modifiersData` set before hand
`implicitRequires`: this relies on mutations/effects from this modifier to data e.g. `popperOffsets` running before hand

And I feel like we maybe should be using some kind of different architecture to run mutations (e.g. `preventOverflow` just directly mutates the `popperOffsets` values) so we don't get weird bugs like this that are hard to track down.

PR #877, opened 2019-12-07
feat: hide modifier

atomiks:
The `clippingParent` technique we have is too naive, as it doesn't consider multiple nested scrolling containers which can clip the element, we effectively need to get the minimum bounds of all the scrolling clipping parents and use that as a virtual rect instead.

I tried working on the above but I discovered a positioning bug when the popper is within the same parent as the reference if the parent has `position: relative` which we need to fix first before we can solve the clipping issue.

I'm not sure about how we should store the data for both the `reference` and `popper` overflows, and how the user can configure different ones for either element. They both have to use the same options.


src/dom-utils/getClippingParent.js
```diff
@@ -6,21 +6,34 @@ import getWindow from './getWindow';
...
+  element: Element,
+  offsetElement?: Element,
+  canEscapeClipping?: boolean
 ): HTMLElement {
   const scrollParent = getScrollParent(element);
   const offsetParent = offsetElement || getOffsetParent(element);
   const win = getWindow(element);
 
+  if (canEscapeClipping === undefined) {
+    canEscapeClipping = ['absolute', 'fixed'].includes(
```
FezVrasta: Absolute elements can’t escape the clipping parent described by this function.

atomiks: The variable is poorly named 🤦‍♂ , it means "this element can escape its scrolling container"

FezVrasta: But they can’t, this function specifically returns containers that will clip absolutely positioned elements.

atomiks: Yes I've tested it, it works. It handles the case where the reference element (usually not absolutely positioned) is clipped by its direct scrolling container as well. The popper element is still clipped appropriately

atomiks: I've made the explanation clearer.

atomiks:
The arrow is broken again since it's placed before `preventOverflow`. I think there's a bug with the modifier ordering

atomiks:
I came up with something to be able to specify custom detectOverflows. It's fairly awkward, especially the naming, so any help improving it would be nice.

It's also really cumbersome to require specifying custom detectOverflow objects just to use the `hide`, so building it in would be better hopefully ensuring it doesn't bloat bundle size too much :\

FezVrasta:
If the issue is defining a custom detect overflow to make hide work, why don’t we export hide + custom detectOverflow from the hide module, so that importing it will include two modifiers?

We could easily flatten the modifiers list to make it work

atomiks:
> modifiers.flat()

I've been wondering about using super modern methods – shouldn't we stick to ES2017, because if we continue to use new methods, the consumers will need to "keep up" with the polyfills and their apps will break for old browsers when updating to new versions if they aren't polyfilling our newly used methods.

FezVrasta:
`.reduce((flat, x) => [...flat, x], [])`?

atomiks:
I don't think that will automatically merge the inner `custom` object option unfortunately

FezVrasta:
I was proposing an alternative that doesn't need a new option.

You should use the plain `detectOverflow` modifier. Give it a unique name, and then people that want to customize it, will do it by referring to it.

atomiks:
That means we need to make a PR that uses the name passed to the modifier, `state.modifiersData[name] = {...}` right? Currently the names are hardcoded

FezVrasta:
Correct, I thought I did already 🤦‍♂️

FezVrasta:
(I just pushed a commit to fix that)

atomiks:
I exported with custom detectOverflow as an array but it won't work yet obvs, I'll rebase with master if it gets implemented?

FezVrasta:
That's great! I'll add support for nested modifiers in a bit.

FezVrasta:
Typing this change is quite a mess. Maybe it would be easier to tell users to do:

```
createPopper(x, y, {
  modifiers: [...flip]
});
```

So we can keep the library lean.

atomiks:
The new modifier is simple anyway unlike before, they can just add this when using `hide`

```js
{
  ...detectOverflow,
  name: 'detectOverflow',
  options: {
    context: 'reference'
  }
}
```

tests/visual/modifiers/hide/index.html
```diff
@@ -0,0 +1,109 @@
...
+      },
+      {
+        name: 'applyHideStyles',
+        enabled: true,
+        phase: 'write',
+        fn({ state }) {
+          const popper = state.elements.popper;
+
+          if (state.modifiersData.hide.isReferenceHidden) {
+            popper.style.visibility = 'hidden';
```
FezVrasta: Should we change the `applyStyles` modifier to write to `state.` directly so that we don't need a custom modifier here?

atomiks: Yes but I thought that was gonna be a separate PR

FezVrasta: I just pushed the changes

atomiks: Cannot assign `data.isReferenceHidden` to `state.attributes['data-poppe...']` because  boolean [1] is incompatible with  object type [2].Flow(InferError)

atomiks:
Should we enforce a `name` style when using other modifiers as a "namespace" for built-in modifiers?

e.g. instead of

`detectOverflowHide`

- `detectOverflow#hide`
- `detectOverflow:hide`
- `detectOverflow/hide`

src/dom-utils/getClippingRect.js
```diff
@@ -0,0 +1,71 @@
...
+        ? getOffsetParent(element)
+        : element;
+    if (
+      isElement(scrollParent) &&
+      isElement(clipperElement) &&
+      scrollParent.contains(clipperElement)
+    ) {
+      clippingParents.push(scrollParent);
+    }
+  });
```
FezVrasta:
```js
const clippingParents = scrollParents.filter(scrollParent => {
  const clipperElement = canEscapeClipping && isHTMLElement(element) ? getOffsetParent(element) : element;
  return isElement(scrollParent) && isElement(clipperElement) && scrollParent.contains(clipperElement);
});
```

src/modifiers/hide.js
```diff
@@ -0,0 +1,47 @@
...
+  right: number,
+  left: number,
+};
+
+const isAnySideFullyClipped = (overflow: ModifierData): boolean =>
+  [top, right, bottom, left].some(side => overflow[side] >= 0);
+
+export function hide({ state }: ModifierArguments<Options>) {
+  const reference = state.measures.reference;
+  const overflow = state.modifiersData.detectOverflowHide.clippingArea;
```
FezVrasta: I would probably make `detectOverflowHide` a configurable option.

```
  const overflow = state.modifiersData[options.detectOverflowModifierName].clippingArea;
```

src/modifiers/hide.js
```diff
@@ -0,0 +1,47 @@
...
+  const referenceClippingOffsets = {
+    top: overflow.top - reference.height,
+    right: overflow.right - reference.width,
+    bottom: overflow.bottom - reference.height,
+    left: overflow.left - reference.width,
+  };
+
+  const isReferenceHidden = isAnySideFullyClipped(referenceClippingOffsets);
+
+  state.modifiersData.hide = {
```
FezVrasta: `state.modifiersData[name]`

src/modifiers/hide.js
```diff
@@ -0,0 +1,47 @@
...
+    referenceClippingOffsets,
+    isReferenceHidden,
+  };
+
+  state.attributes.popper = {
+    ...state.attributes.popper,
+    'data-popper-reference-hidden': isReferenceHidden,
```
FezVrasta: Doesn't `context` decide if we want to detect `reference` or `popper`?

src/modifiers/hide.js
atomiks: How does that provide any benefit?

src/modifiers/hide.js
atomiks: We still need to do that for every other modifier 😅

FezVrasta:
I changed the visual test to:

```js

  window.instance = createPopper(reference, popper, {
    placement: 'right',
    modifiers: [
      { name: 'preventOverflow', enabled: false },
      hide,
      {
        ...detectOverflow,
        name: 'detectOverflowHide',
        options: {
          context: 'popper',
        },
      },
    ],
  });
```

I would expect the popper element to gain the `data-` attribute to mark it as hidden when the popper overflows the container, but it doesn't happen. What's going on?

src/modifiers/hide.js
FezVrasta: Users can recycle the same detectOverflow modifier for various modifiers, rather than having to define a new one each time.

src/modifiers/hide.js
FezVrasta: Let's start somewhere :-P

atomiks:
The `context` option determines which element we're checking the `clippingArea` and `visibleArea` for.  So the `popper` context is used by `flip` & `preventOverflow`, while `reference` is used for `hide`.

What you're looking for (and what I wanted as well) was the popper overflowing the reference's `clippingArea` (e.g. it looks weird/detached outside), which is like an "alt context". But I haven't worked on it because preventOverflow was breaking it and figured we could solve it sometime in the future...

src/modifiers/hide.js
atomiks: The case where we disable `preventOverflow` and the popper leaves its bounds is not useful for `hide`, the useful one is when the popper leaves the reference's bounds entirely while the reference is still partially visible, it looks odd - seen in the test.

src/modifiers/hide.js
atomiks: A problem with that is how do we specify `requires`?

FezVrasta: For now I’d ask users to make sure they set the proper dependency names, we could build some fancy solution but it’s overkill for now

FezVrasta: For now I’d ask users to make sure they set the proper dependency names, we could build some fancy solution but it’s overkill for now

atomiks: It makes `detectOverflowName` option pointless however?

FezVrasta: Why? We can't read from the `requires` list to know which one to use, it may have a completely different name we can't recognize.

atomiks: The problem is the user needs to spread in the original `requires` as well into their custom specified `requires` property, and it will contain the default name for hide's `detectOverflow`?

src/modifiers/hide.js
atomiks: I wonder if we run `detectOverflow` in the `afterMain` phase, it will work? 🤔

src/modifiers/hide.js
atomiks: I misread your `can't` as `can`, sorry 🤦‍♂.

I see what you're saying.

FezVrasta:
> Because we can't know what name they are passing in?

Why can't we leave the default option name?

atomiks:
If they pass in 

```js
options: {
  detectOverflowName: 'customDetectOverflow'
}
```

Then the `requires` validation will falsely error because our validation is based on the default, but they are passing in one with a different name

FezVrasta:
Can’t they override the requires prop? 🤔

atomiks:
That's what I mentioned here: https://github.com/popperjs/popper.js/pull/877#discussion_r355976032

atomiks:
> Users can recycle the same detectOverflow modifier for various modifiers, rather than having to define a new one each time.

I think this is too brittle, let's just force them to use names that the built-in modifier requires 🤷‍♂  especially now since there's yet another modifier required.

src/modifiers/hide.js
atomiks: ^ It works 🙏

atomiks:
https://i.imgur.com/JMepoP0.mp4

Purple = normal
White = `hasPopperEscaped`
Yellow = `isReferenceHidden`

Unfortunately I think `preventOverflow` still causes problems 🤔

src/modifiers/offset.js
```diff
@@ -47,13 +47,25 @@ export function offset({ state, options }: ModifierArguments<Options>) {
...
+    basePlacements.forEach(placement => {
+      const multiply = [right, bottom].includes(placement) ? 1 : -1;
+      const value =
+        // If it's based on the reference context, we don't add `offset`
+        // TODO: we need to figure out something better entirely here
+        (key.includes('Reference') // detectOverflowReference
+          ? 0
+          : [top, bottom].includes(placement)
+          ? y
+          : x) * multiply;
```
atomiks: We need to figure something out about this problem

atomiks:
I think everything works now

src/modifiers/offset.js
FezVrasta: Would it be too bad if we asked users to provide the list of detectOverflow modifiers in the modifier options?

By default it would be:
```
{
  name: 'offset',
  options: {
    detectOverflowModifiers: ['detectOverflow']
  }
}
```

People can then override it if needed. It's good to provide a good and extensible modifier API, but I don't want to end up with 90% of the library size just for making it work.

atomiks: I really dislike having to do that. I think maybe `offset` shouldn't need to have to alter `detectOverflow` at all, somehow

src/dom-utils/getClippingRect.js
```diff
@@ -0,0 +1,66 @@
...
+    return rectToClientRect(getDocumentRect(documentElement));
+  }
+
+  const clippingRect = restClippingParents.reduce((accRect, clippingParent) => {
+    const rect = getBoundingClientRect(clippingParent);
+    accRect.top = Math.max(rect.top, accRect.top);
+    accRect.right = Math.min(rect.right, accRect.right);
+    accRect.bottom = Math.min(rect.bottom, accRect.bottom);
+    accRect.left = Math.max(rect.left, accRect.left);
+    return accRect;
```
FezVrasta: If here you worked on `x`, `y`, `width`, `height`, on lines 60...63 you could call `rectToClientRect(clippingRect)` to compute `top`, `right`, `bottom`, and `left`.

This should simplify the code and allow us to reuse the `rectToClientRect` utility.

src/modifiers/detectOverflow.js
```diff
@@ -1,18 +1,32 @@
...
+  top,
+  right,
+} from '../enums';
+import unwrapVirtualElement from '../dom-utils/unwrapVirtualElement';
 
 type Options = {
   clippingArea: OverflowArea,
   visibleArea: OverflowArea,
+  context: Context,
+  alt: boolean,
```
FezVrasta: What's `alt`?

src/dom-utils/getClippingRect.js
atomiks: How would it work though? We need to min the right/bottom as well

src/modifiers/detectOverflow.js
atomiks: `alt` means it will check the other element's area (i.e. `clippingParents`)

src/dom-utils/getClippingRect.js
FezVrasta: I suspect the following wouldn't be enough, but we have all the data to make the right operations... 🤔

```js
accRect.x = Math.max(rect.x, accRect.x);
accRect.y = Math.max(rect.y, accRect.y);
accRect.width = Math.min(rect.width, accRect.width);
accRect.height = Math.min(rect.height, accRect.height);
```

src/modifiers/detectOverflow.js
FezVrasta: I don't quite understand why do we need to support two areas in the modifier, if we then need to also specify which one we want to use. Why can't we keep the modifier focused to 1 single area, and use multiple modifiers for each area?

atomiks: Do you mean `clippingArea` and `visibleArea`? For `alt`, it uses the other element's rect but its own measure (how the popper can tell it escaped the reference's boundaries)

FezVrasta: Could you find a more descriptive name for the option? I'm still not understanding what's its purpose. Feel free to add code comments to make it clearer too.

FezVrasta:
Closing as this was manually merged

PR #878, opened 2019-12-07
fix: positioning logic with offsetParents

atomiks:
Added a new test there the direct parent is an offsetParent as well as the root, but not the container in the middle


src/dom-utils/getRectRelativeToOffsetParent.js
```diff
@@ -15,37 +15,41 @@ export default (
...
-  const ancestorOffsetParents = isFixed ? [directOffsetParent] : [];
-  let currentOffsetParent = directOffsetParent;
+  const relevantScrollParents = [];
 
-  while (isElement(currentOffsetParent)) {
-    currentOffsetParent = getOffsetParent(currentOffsetParent);
-    ancestorOffsetParents.push(currentOffsetParent);
+  // We want all the scrolling containers only up to and including the
+  // offsetParent
+  for (let i = 0; i < scrollParents.length; i++) {
```
FezVrasta: The following should do the same:

```
const relevantScrollParents = scrollParents.slice(0, scrollParents.indexOf(offsetParent));
```

PR #879, opened 2019-12-07
fix: blurry text (better)

### LINKED ISSUE #764: Ideas for pixel-perfect, non-blurry display
**Disclaimer:** The pixel-perfect / blurry issue has many other tickets describing a whole variety of possible workarounds and tweaks, but I'm adding this as a new ticket in the hopes that some of these ideas can help further improve or refine the positioning in popper.

First, popper calculates it's initial positioning using [`getOuterSizes.js`](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getOuterSizes.js#L14-L15). Unfortunately, this uses `element.offsetWidth` and `element.offsetHeight` to calculate popper's dimensions. The problem is that these only return integer values, so calculations based on fine positioning (like that which may lead to blurriness) are often  _slightly_ off. It's this kind of "slightly off" stuff that makes you pull your hair out when things don't line up perfectly.

Instead of using these methods, we are calling `element.getBoundingClientRect`, which returns exact fractional (if necessary) values for the position of popper. Doing this allows the initial positioning values to be more accurate and allow better placement (with the chance for less blurriness) later on. We created a modifier, of sort order `50` (so it runs as the first modifier, not sure if this is correct or ideal). It runs before the other modifiers though. Here's what our `popperOffsets` call looks like:

```coffee
            popperOffsets:
              order: 50
              enabled: true
              fn: (data, options) ->
                dom = data.instance.popper
                ref = data.offsets.reference
                pos = data.placement.split('-')[0]

                css = dom.ownerDocument.defaultView.getComputedStyle dom
                bcr = dom.getBoundingClientRect()

                [t, r, b, l, w, h] = 'top right bottom left width height'.split ' '

                vert = pos is t or b # top/bottom?
                main = pos is t or l # top/left?

                [start, half] = if vert then [l, w] else [t, h] # used to center
                [shift, full] = if vert then [t, h] else [l, w] # used to offset

                pop =
                  [t]: ref[t] # align top
                  [l]: ref[l] # align left
                  [w]: bcr[w] + parseFloat(css.marginLeft or 0) + parseFloat(css.marginRight  or 0)
                  [h]: bcr[h] + parseFloat(css.marginTop  or 0) + parseFloat(css.marginBottom or 0)
                  position: data.offsets.popper.position

                pop[start] += (ref[half] - pop[half]) / 2 # center on axis
                pop[shift] += if main then -pop[full] else ref[full] # adjust to edge

                data.offsets.popper = pop
                data
```

Second, the `computeStyle` modifier is used to calculate the actual positioning of the popper before rendering. It's this code that is responsible for figuring out exact placement and whether the GPU should be used (via `translate3d`) or if "normal" positioning (via `top` and `left`) should be used instead. The current code has a whole bunch of magic to try to determine how to correctly position the text to make it line up and not be blurry, etc.

What we've found is that GPU / hardware accelerated positioning (via `transform: translate3d(...)`), which is supposed to be much more efficient and generally way better, always leads to some blurry pixelation unless the positioning occurs on a "proper" pixel boundary. We've found that for a non-retina display, this means that we see pixelation unless the final position is integer pixel-aligned (ie - no decimal). On retina displays (which have a `window.devicePixelRatio` of 2), that we can do GPU position on integer pixel or 0.5 pixel boundaries and not see any blurriness. I assume if `window.devicePixelRatio` is 4, then blurriness would not occur on integer, 0.25, 0.5, or 0.75 pixel boundaries. In our code, if we are "off" by more than some small amount (0.001 of a pixel) from one of these boundaries, then we punt and failover to using "normal" positioning. But, if we are within the small delta or a proper pixel boundary, then we use the GPU via `translate3d`. This means that the GPU can be used dynamically, based on the context of that specific popper. The code for our `computeStyle` modifier is:

```coffee
            computeStyle:
              order: 850
              enabled: true
              fn: (data, options) ->
                dom = data.instance.popper
                ref = data.offsets.reference
                pop = data.offsets.popper

                outer = dom.getBoundingClientRect()
                inner = dom.querySelector('.content-inner').firstChild.getBoundingClientRect()

                [t, r, b, l, w, h] = 'top right bottom left width height'.split(' ')

                x = pop[l] - inner[l] + outer[l]
                y = ref[t] - inner[t] + outer[t]

                # use gpu if possible
                dpr = window.devicePixelRatio
                gpu = ![x, y].find (n) => Math.abs(n - frac n, dpr) > 0.001

                styles =
                  if gpu
                    transform: "translate3d(#{x}px, #{y}px, 0)"
                    top: 0
                    left: 0
                    willChange: 'transform'
                  else
                    top: y
                    left: x
                    willChange: 'top, left'

                data.attributes  = { 'x-placement': data.placement, data.attributes... }
                data.styles      = { data.styles..., styles... }
                data.arrowStyles = { data.offsets.arrow..., data.arrowStyles... }
                data
```

The `frac` helper is used to for fractional steps, so `frac 3.333, 4` would yield `3.25`, since that's the closest 1/4th of an integer. Here's that code:

```coffee
  frac = (val, n=1) -> if n > 1 then Math.round(val * n) / n else if n is 1 then (val | 0) else val
```

The end result, for us at least, has been pixel-perfect poppers that use the GPU when they can and failover to the not-optimized, but not-blurry either "normal" positioning. This gives us the best of both worlds, for only a small amount of tweaking of the code.

Hope this helps and sorry for creating a new ticket. It just didn't seem like this should be "tacked on" to the other closed tickets in the issue tracker.

atomiks:
After testing, subpixel values like `.75px` causes blurry text on my retina display, so we'll need to round the values. We can do this to the nearest suitable subpixel reported by `window.devicePixelRatio`

We still need to disable `transform` on low PPI displays because layer acceleration can disable subpixel rendering which leads to slightly blurry text

Fixes #764 (forever hopefully)

### DIFF
--- a/src/modifiers/computeStyles.js
+++ b/src/modifiers/computeStyles.js
@@ -13,6 +13,15 @@ type Options = {
   gpuAcceleration?: boolean,
 };
 
+// Round the offsets to the nearest suitable subpixel based on the DPR
+const roundOffsets = ({ x, y }) => {
+  const dpr = window.devicePixelRatio;
+  return {
+    x: Math.round(x * dpr) / dpr,
+    y: Math.round(y * dpr) / dpr,
+  };
+};
+
 export const mapToStyles = ({
   offsets,
   position,
@@ -22,20 +31,24 @@ export const mapToStyles = ({
   position: PositioningStrategy,
   gpuAcceleration: boolean,
 }) => {
-  // @1x displays in Chrome can be blurry due to non-rounded offsets, but this
-  // introduces slight positioning errors. TODO: somehow solve this better
-  // by default it is active, disable it only if explicitly set to false
+  const { x, y } = roundOffsets(offsets);
+
+  // Layer acceleration can disable subpixel rendering which causes slightly
+  // blurry text on low PPI displays.
+  // Zooming can change the DPR, but it seems to report a value that will
+  // cleanly divide the values into the appropriate subpixels.
   if (gpuAcceleration === false || window.devicePixelRatio < 2) {
     return {
-      top: `${offsets.y}px`,
-      left: `${offsets.x}px`,
+      top: `${y}px`,
+      left: `${x}px`,
+      transform: '',
       position,
     };
   } else {
     return {
       top: '0px',
       left: '0px',
-      transform: `translate3d(${offsets.x}px, ${offsets.y}px, 0)`,
+      transform: `translate3d(${x}px, ${y}px, 0)`,
       position,
     };
   }
--- a/src/modifiers/__snapshots__/computeStyles.test.js.snap
+++ b/src/modifiers/__snapshots__/computeStyles.test.js.snap
@@ -5,6 +5,7 @@ Object {
   "left": "10px",
   "position": "absolute",
   "top": "5px",
+  "transform": "",
 }
 `;
 
@@ -22,5 +23,6 @@ Object {
   "left": "10px",
   "position": "absolute",
   "top": "5px",
+  "transform": "",
 }
 `;


src/modifiers/computeStyles.js
```diff
@@ -22,20 +31,24 @@ export const mapToStyles = ({
...
+  // blurry text on low PPI displays.
+  // Zooming can change the DPR, but it seems to report a value that will
+  // cleanly divide the values into the appropriate subpixels.
   if (gpuAcceleration === false || window.devicePixelRatio < 2) {
     return {
-      top: `${offsets.y}px`,
-      left: `${offsets.x}px`,
+      top: `${y}px`,
+      left: `${x}px`,
+      transform: '',
```
atomiks: I had to do this because zooming changes the DPR and therefore the positioning technique, so the stale transform attribute gets left behind, breaking the positioning

FezVrasta:
Cool! May you update the unit tests? (`yarn test:unit -u`)

PR #881, opened 2019-12-08
fix: modifier ordering

atomiks:
I used the second algorithm in the link (the one with most upvotes) instead and adapted it - it seems to work.

I have no idea how it works, so I think we should add some unit tests to make sure it works appropriately?

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,9 @@
       "<rootDir>/src/**/*.test.js",
       "<rootDir>/tests/**/*.test.js"
     ],
+    "globals": {
+      "__DEV__": true
+    },
     "globalSetup": "jest-environment-puppeteer/setup",
     "globalTeardown": "jest-environment-puppeteer/teardown",
     "testEnvironment": "jsdom",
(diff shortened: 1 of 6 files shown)


FezVrasta:
What’s wrong with the current one?

atomiks:
It wasn't working properly to put it simply 😆  I don't know the specifics as I don't know how it works

atomiks:
The only reason #877 is failing is because `preventOverflow` is being placed before `arrow` again, which happens when I export the `hide` modifier, so it just isn't ordering properly. This one does order properly.

I should probably make sure the tests I added will fail for the previous algo

atomiks:
Sorry for that complete mess 🤦‍♂ 

It should work now lol

package.json
```diff
@@ -88,6 +88,9 @@
       "<rootDir>/src/**/*.test.js",
       "<rootDir>/tests/**/*.test.js"
     ],
+    "globals": {
```
FezVrasta: What’s this for?

atomiks: So Jest doesn't error for files with `__DEV__`

atomiks:
You already are validating dependencies in `requires` it seems in a commit 3 days ago, so you can remove the validation here and the test

PR #886, opened 2019-12-14
fix: positioning bug

atomiks:
1. There was an index error for the `relevantScrollParents` slicing
2. The popper and arrow need to use `offset` values (before any transforms) while the reference needs to use `getBoundingClientRect` values. The reason is we want to position the popper considering the reference's transforms, but we don't want to position the popper/arrow considering their transforms (since they are positioned using layout)


src/dom-utils/getRectRelativeToOffsetParent.js
```diff
@@ -24,9 +24,10 @@ export default (
 
   // We want all the scrolling containers only up to and including the
   // offsetParent
+  const offsetScrollParentIndex = scrollParents.indexOf(offsetParent);
   const relevantScrollParents = scrollParents.slice(
     0,
-    scrollParents.indexOf(offsetParent) + 1
```
FezVrasta: Maybe `Math.abs(scrollParents.indexOf(offsetParent)) + 1`

It’s shorter

atomiks: `Math.abs(-1) => 1`

FezVrasta: `Math.max(scrollParents.indexOf(offsetParent), 0) + 1`?

I’m thinking about ways to shrink the code 😂

src/dom-utils/getRectRelativeToOffsetParent.js
```diff
@@ -5,14 +5,27 @@ import listScrollParents from './listScrollParents';
...
+  isLayout: boolean = true
 ): Rect => {
+  if (isLayout && isHTMLElement(element)) {
+    return {
+      x: element.offsetLeft,
+      y: element.offsetTop,
+      width: element.offsetWidth,
+      height: element.offsetHeight,
+    };
+  }
```
FezVrasta: I think it would be cleaner if we defined a new utility where we do:

```js
import type { Rect, VirtualElement } from '../types';
import { isHTMLElement } from './instanceOf';
import getRectRelativeToOffsetParent from './getRectRelativeToOffsetParent';

export default (element: Element | VirtualElement, isFixed): Rect => {
  if (isHTMLElement(element) {
    return {
      x: element.offsetLeft,
      y: element.offsetTop,
      width: element.offsetWidth,
      height: element.offsetHeight,
    };
  }

  return getRectRelativeToOffsetParent(element, isFixed);
}
```

Doing so we keep `getRectRelativeToOffsetParent` clean and focused, and we can use this new function where we need it.

atomiks: But they both get it relative to the offset parent, it's just a branched behavior? 🤔

FezVrasta:
Thank you for addressing my comments 🤗

PR #888, opened 2019-12-15
feat: root overflow areas

atomiks:
We need to be able to specify a "root" area that overflow will fallback to.

`flip` goes from `clippingParents` to the `viewport` root, while `preventOverflow` (usually) goes back to the `document` root while "skipping" the viewport. Sometimes users choose viewport for preventOverflow  for the root though.

```ts
type Options = {
  area: OverflowArea,
  rootArea: RootOverflowArea,
  elementContext: Context,
  altArea: boolean,
};
```

There was an idea to specify `[area, ...fallbackAreas]` but it didn't really work.

### DIFF
--- a/src/enums.js
+++ b/src/enums.js
@@ -26,6 +26,7 @@ export type OverflowArea =
   | HTMLElement
   | typeof clippingParents
   | typeof viewport;
+export type RootOverflowArea = typeof viewport | 'document';
 
 export const popper: 'popper' = 'popper';
 export const reference: 'reference' = 'reference';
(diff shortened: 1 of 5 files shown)


src/enums.js
```diff
@@ -26,6 +26,7 @@ export type OverflowArea =
   | HTMLElement
   | typeof clippingParents
   | typeof viewport;
+export type RootOverflowArea = typeof viewport | 'document';
```
FezVrasta: can't we specify `document` as OverflowArea also?

atomiks: Would add more code to do that check. They can use `document.documentElement` easily enough if they want to bypass `clippingParents`? 🤷‍♂

FezVrasta: Ok it makes sense. Thanks

PR #890, opened 2019-12-15
fix: table positioning

atomiks:


src/dom-utils/isTableElement.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+import getComputedStyle from './getComputedStyle';
+
+export default (element: Element) =>
```
FezVrasta: You need to define the return type
```suggestion
export default (element: Element): boolean =>
```

src/dom-utils/getTrueOffsetParent.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+import { isHTMLElement } from './instanceOf';
+
+export default (element: Element) =>
```
FezVrasta:
```suggestion
export default (element: Element): ?HTMLElement =>
```

PR #891, opened 2019-12-16
[wip] fix: alt offsetParent contexts

atomiks:
This fixes the case where the popper and reference are in different `offsetParent` contexts.

Cases where this is broken:
- Tables are broken when there are offsetParents around the reference/table. The offset position seems correct but it's missing certain scrollParents in the calculation.
- The relative offset body test seems to have broken.
- If there are multiple nested containers with offsetParents and the popper is outside of them, it's broken.

I copied/pasted the `findCommonOffsetParent` function from v1 and there were lots of Flow errors that need fixing.

Help finding the magic formula is needed 😥


FezVrasta:
Why `table/offset-parent-multiple` doesn't work when I try it manually, but works on CI?

src/dom-utils/getCommonOffsetParent.js
```diff
@@ -0,0 +1,84 @@
+// @flow
```
FezVrasta: I think this file can be simplifies as follows:

```typescript
// @flow
import getOffsetParent from './getOffsetParent';
import { isHTMLElement, isShadowRoot } from './instanceOf';
import getWindow from './getWindow';

function isOffsetContainer(element: Element): boolean {
  const firstElementChild = element.firstElementChild;
  return !firstElementChild || getOffsetParent(firstElementChild) === element;
}

function getRoot(node: Node): Node {
  const parentNode = node.parentNode;

  if (parentNode != null) {
    return getRoot(parentNode);
  }

  return node;
}

export default function findCommonOffsetParent(
  element1: Element | ShadowRoot,
  element2: Element | ShadowRoot
): Element {
  // If one of the nodes is inside shadowDOM, find which one
  const element1root = getRoot(element1);
  const element2root = getRoot(element2);

  if (isShadowRoot(element1root)) {
    return findCommonOffsetParent(element1root.host, element2);
  } else if (isShadowRoot(element2root)) {
    return findCommonOffsetParent(element1, element2root.host);
  }

  // Here we make sure to give as "start" the element that comes first in the DOM
  const order =
    element1.compareDocumentPosition(element2) &
    Node.DOCUMENT_POSITION_FOLLOWING;
  const start = order ? element1 : element2;
  const end = order ? element2 : element1;

  // Both nodes are inside #document
  // Get common ancestor container
  const range = document.createRange();
  range.setStart(start, 0);
  range.setEnd(end, 0);
  const commonAncestorContainer = range.commonAncestorContainer;

  // If ancestor container is not an element, we must return a safe
  // fallback, body is a good bet
  if (!isHTMLElement(commonAncestorContainer)) {
    return getWindow(commonAncestorContainer).body;
  }

  // if ancestor container is also an offset container, we found our winner
  if (isOffsetContainer(commonAncestorContainer)) {
    return commonAncestorContainer;
  }

  return getOffsetParent(commonAncestorContainer);
}
```

You also need to add a new `isShadowRoot` utility in `instanceOf.js`:

```typescript

/*:: declare function isShadowRoot(node: mixed): boolean %checks(node instanceof
  ShadowRoot); */

function isShadowRoot(node) {
  const OwnElement = getWindow(node).ShadowRoot;
  return node instanceof OwnElement;
}
```

atomiks:
> Why table/offset-parent-multiple doesn't work when I try it manually, but works on CI?

The snapshot was written in a broken state. The code is just wrong and I don't know how to fix it so it works for all the cases 😞 I was hoping you could try to help solve it.

atomiks:
I fixed positioning 🤩 

We were basically trying to "counter-act" other unnecessary logic in `popperOffsets` modifier (`getCommonTotalScroll`), making the problem insurmountable without removing that. Now everything works

PR #892, opened 2019-12-16
Use prepare instead of postinstall for patch-package

benvan:
so that patch-package doesn't run when popperjs imported as a dependency

# Motivation
Installing popper.js@next as a dependency in a project will typically fail, because "patch-package" likely won't be present in the host project

# Justification
It seems that the patches are only needed at build-time - in this case, switching to `prepare` achieves the same effect without affecting downstream dependents


FezVrasta:
Thanks!

PR #894, opened 2019-12-17
[wip] feat: preventOverflow#tetherOffset and flip#checkVariation

atomiks:
### preventOverflow#tetherOffset

Based on the problems `tether` was causing, we need to just use `edges` (or a boolean now) and give the user the ability to define a tetherOffset to customize when it "activates"

### flip#checkVariation

This replicates `flipVariationsByContent` in v1 but it's on by default. When using a variation placement, the popper will attempt to flip to the opposite variation instead of preventing overflow, which keeps the edges of the reference and popper aligned as best as it can which seems like a better default.


src/modifiers/flip.js
```diff
@@ -45,13 +49,50 @@ export function flip({ state, options, name }: ModifierArguments<Options>) {
...
+    /*
+    const fitsOnOriginalSide =
+      overflow[originalSide] + paddingObject[originalSide] <= 0;
+      */
+    const fitsOnOppositeSide =
+      overflow[oppositeSide] + paddingObject[oppositeSide] <= 0;
+
+    if (fitsOnOppositeSide) {
+      flippedPlacement = getOppositeVariationPlacement(flippedPlacement);
+    }
```
atomiks: This should be 

```js
    if (!fitsOnOriginalSide && fitsOnOppositeSide) {
      flippedPlacement = getOppositeVariationPlacement(flippedPlacement);
    }
```

But it causes an infinite loop.

I think I should cache the original placement somewhere but can't get it working, need some help

src/modifiers/flip.js
```diff
@@ -45,13 +49,59 @@ export function flip({ state, options, name }: ModifierArguments<Options>) {
...
+
+    if (
+      !fitsOnOriginalSide &&
+      fitsOnOppositeSide &&
+      oppositeVariation !== data.currentVariation
+    ) {
+      flippedPlacement = oppositeVariation;
+      state.modifiersData[`${name}#persistent`] = {
+        currentVariation: oppositeVariation,
+      };
```
atomiks: We need a better API for persistent data...

PR #897, opened 2019-12-19
refactor: detectOverflow to utility

atomiks:
Instead of making `detectOverflow` a modifier, it's better if it's used as a utility. 

The code to handle variation flipping is simplified and cleaner now,  changing `checkVariations` => `flipVariations` like v1. It merges both `flipVariations` and `flipVariationsByContent` into one by checking the difference in lengths of the reference and popper, so it "just works". The distinction doesn't seem useful and v1 behavior was not intuitive it seems.

There is one low-priority problem: with e.g. clockwise flipping variation placements, if none of the placements fit, it prefers the original placement even if that wouldn't be best. When scrolling, there can be a flicker as the popper decides to flip into view or not. This seems tricky to solve so I've left it out for now, it's not exactly that important, but something to note


src/modifiers/flip.js
```diff
@@ -1,103 +1,135 @@
...
+  const variationPlacement = getVariationPlacement(placement);
+  const oppositeVariationPlacement = variationPlacement === start ? end : start;
+
+  // $FlowFixMe
+  return [
+    `${basePlacement}-${oppositeVariationPlacement}`,
+    // $FlowFixMe
+    `${oppositeBasePlacement}-${variationPlacement}`,
+    `${oppositeBasePlacement}-${oppositeVariationPlacement}`,
+  ];
```
atomiks: Types need fixing

PR #902, opened 2019-12-22
fix: position when offset parent has a border

atomiks:
Seems like we don't want to take into account the `border` of the common offset parent. Works with both `box-sizing: content-box` and `box-sizing: border-box`.

Approach seem right?


atomiks:
While this fixed the offset issue, it looks like we need to also account for this problem...

img

atomiks:
Hmm Popper v1 is broken in this case as well: https://codepen.io/atomiks/pen/gObWgwz

PR #904, opened 2019-12-23
docs: route ordering and utils docs

atomiks:
We want to order like:

- `Home`
- `Documentation`
    - `Modifiers`
        - `Popper Offsets`
        - `...` (based on default dependencies and phase, we can manually specify this?)
    - `Utils`
        - `Detect Overflow`
    - `Typings`

Tried using `index` in the frontmatter but the `MdxRoutes` plugin doesn't seem to pick it up.

`docs/src/utils/processRoutes.js` algorithm is so inefficient/brittle 😥, needs improvement, pls help @FezVrasta 😄

### DIFF
--- a/docs/src/utils/processRoutes.js
+++ b/docs/src/utils/processRoutes.js
@@ -1,0 +1,43 @@
+const BLACKLIST = new Set(['/404/']);
+const DIR_ORDER = ['modifiers', 'utils', 'typings'];
+const MODIFIER_ORDER = [
+  'Popper Offsets',
+  'Offset',
+  'Prevent Overflow',
+  'Arrow',
+  'Flip',
+  'Hide',
+  'Compute Styles',
+  'Apply Styles',
+  'Event Listeners',
+];
+
+// I think this takes 0.5ms - 1ms on my MBP, we need to optimize this...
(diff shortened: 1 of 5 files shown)


docs/src/utils/processRoutes.js
```diff
@@ -0,0 +1,43 @@
...
+  'Prevent Overflow',
+  'Arrow',
+  'Flip',
+  'Hide',
+  'Compute Styles',
+  'Apply Styles',
+  'Event Listeners',
+];
+
+// I think this takes 0.5ms - 1ms on my MBP, we need to optimize this...
```
FezVrasta: What’s the issue? Isn’t this rendered at build time?

atomiks: Well it's called in the JSX, doesn't React call it in hydration phase?

atomiks: I guess let's not care about its perf until it's a problem anyway. Even on super slow hardware it will complete within an animation frame, we probably aren't going to list a huge navigation menu with lots of items in the array 🤷‍♂

atomiks:
Can this be merged?

PR #906, opened 2019-12-24
feat: arrow padding

atomiks:
\+ refactor/fix: remove all margins from rect calculations

There was a problem with the arrow on the `bottom` placement on first update due to arrow margins. I think the easiest thing to do is just not calculate any margins and create an option for padding instead of relying on the computed CSS. If we can figure out a clean way to allow conditional size changes of elements based on placements, we might add margin calcs back in, but right now they only cause problems.

tl;dr

popper margins = use `offset` modifier
arrow margins = use `padding` option

no jitters or wrong positioning will happen with these

---

What about `rem` units? Popper is built using pixel values everywhere. If you want to use `rem` units, you should convert your `rem` values to pixels when passing their values to Popper. This can be done dynamically by getting the root `html` element's font-size value and multiplying it by the rem value.

e.g.

```js
// border-radius: 1rem;
const padding = '1rem';
const paddingInPx = parseFloat(
  getComputedStyle(document.documentElement).fontSize
) * parseFloat(padding);
```


FezVrasta:
Should the website demo get updated too?

src/modifiers/preventOverflow.js
```diff
@@ -104,9 +103,28 @@ function preventOverflow({ state, options, name }: ModifierArguments<Options>) {
...
+        ? arrowPadding
+        : expandToHashMap(arrowPadding, basePlacements)
+    );
+    const arrowPaddingMin = arrowPaddingObject
+      ? arrowPaddingObject[mainSide]
+      : 0;
+    const arrowPaddingMax = arrowPaddingObject
+      ? arrowPaddingObject[altSide]
+      : 0;
+
```
atomiks: Not sure about this? We check for the first `arrow` modifier in `state.orderedModifiers` - in our docs' `usePopper`, I pass in a default modifier for arrow padding to be enabled everywhere, subsequent `arrow` modifiers' padding being specified gets ignored

PR #909, opened 2019-12-26
fix: defends against undefined tooltip.popperInstance

jwakeman-ebsco:
Fixes an issue that if `this.popperInstance` has already been destroyed when the `mousedown` callback is executed, the app crashes with `"cannot read property 'popper' of undefined"`

- [x] the tests are passing by running `yarn test`

NA -- Add any relevant tests to cover the code you have changed and/or added.

NA -- If you change the public API, try to update the Typescript definitions accordingly:
  https://github.com/FezVrasta/popper.js/blob/master/packages/popper/index.d.ts
  This is not required but will help a lot. Mention @giladgray for help as needed.

### DIFF
--- a/packages/tooltip/src/index.js
+++ b/packages/tooltip/src/index.js
@@ -362,9 +362,9 @@ export default class Tooltip {
           if (!this._isOpening) {
             return;
           }
-          const popper = this.popperInstance.popper;
           if (reference.contains(e.target) ||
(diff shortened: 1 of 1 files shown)


FezVrasta:
Hi, thanks for the PR, unfortunately Tooltip.js is discontinued, please use Tippy.js instead.

PR #912, opened 2020-01-03
fix: popper instance

atomiks:
- Runs `onLoad` and `onDestroy` in `.setOptions()`
- I don't see the need to `return state` since the user just modifies a permanent state object. There's no returning of a clone involved, so it's not necessary? It's easy to forget to remember to return the state object too so this seems less bug prone.
- `validateModifiers` was not running for modifiers if they got filtered out of `state.orderedModifiers` (e.g. they have no `phase` property - I kept forgetting it)
- Adds the `state` object to the instance

### DIFF
--- a/src/__snapshots__/index.test.js.snap
+++ b/src/__snapshots__/index.test.js.snap
@@ -1,0 +1,35 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`returns expected instance object 1`] = `
+Object {
+  "destroy": [Function],
+  "forceUpdate": [Function],
+  "setOptions": [Function],
+  "state": Object {
+    "attributes": Object {},
+    "elements": Object {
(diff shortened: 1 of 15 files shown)


FezVrasta:
I think the `return state` is a valid design decision, it allows to take full control over the `state` object without recurring to in place mutation.

atomiks:
> I think the return state is a valid design decision, it allows to take full control over the state object without recurring to in place mutation.

Not sure what this means? 🤔  Can you explain?

We can make returning the state optional like

`state = fn(...) || state` which was already written for the `onLoad` modifier

FezVrasta:
Making it optional probably makes sense.

> Not sure what this means? 🤔 Can you explain?

Say you want to completely replace the `state` object, I don't want to ask consumers to do that by mutating the existing state, I'd rather have them return a new object with just what they need.

PR #915, opened 2020-01-05
refactor(VirtualElement): remove contextElement and add docs

atomiks:
There's no need for `contextElement` anymore since the offsetParent is always relative to the popper element.


atomiks:
Fixed the `$FlowFixMe`

PR #918, opened 2020-01-06
fix: createPopper argument types

### LINKED ISSUE #917: SVGElement reference and popper type
https://github.com/popperjs/popper.js/blob/85bb11cec69f680bf2c0e0dca4b0d219aefb95fc/src/index.js#L51-L52

Could the type for the `reference` and  `popper` element be `Element`? I'm using an `<svg>` element for the reference element which causes typescript warnings. 

Following question, if not any SVG element, how about just `SVGSVGElement` in there too?

atomiks:
Should fix #917 @HHogg 

I removed the whole `jQueryWrapper` handling because I don't think it's useful moving forward with such a modern library, Tippy doesn't handle this at all and doesn't get complaints.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,20 +1,20 @@
 // @flow
 import type {
-  JQueryWrapper,
   State,
   Options,
   Modifier,
   Instance,
+  VirtualElement,
 } from './types';
 import getCompositeRect from './dom-utils/getCompositeRect';
 import getLayoutRect from './dom-utils/getLayoutRect';
 import listScrollParents from './dom-utils/listScrollParents';
 import getOffsetParent from './dom-utils/getOffsetParent';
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -48,24 +48,19 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
   } = generatorOptions;
 
   return function createPopper(
-    reference: HTMLElement | JQueryWrapper,
-    popper: HTMLElement | JQueryWrapper,
+    reference: Element | VirtualElement,
+    popper: HTMLElement,
```
HHogg: I've not used an SVG element as a popper element before, but in theory could it be?

atomiks: I think there are various problems with that, but wrapping an SVG node inside a `<div>` (which becomes the popper) works fine

PR #924, opened 2020-01-09
fix: scrollSum on Edge

atomiks:
cc @donnysim - can you check all of the visual tests in Edge and IE11 to see if they work correctly? [Info available here](https://github.com/popperjs/popper.js/tree/next#hacking-the-library)


donnysim:
Just a note, to be able to run on Windows with things live env changes you need cross-env:
```
"dev": "cross-env NODE_ENV=dev concurrently 'yarn serve' 'yarn build:dev --watch'",
```
---

All visual tests pass on Edge (can't test IE11 without modifying every template because IE does not support script type modules), but at the same time my project use case stays the same - broken on Edge (but works fine on IE11).

The main issue seems to be that every browser (Chrome, Firefox, IE11, Safari?) for `body.scrollTop` returns `0` while only Edge returns the actual scroll position, so based on how all the current calculations are based (`getHTMLElementScroll(node)` returns 0 when element is body, except on Edge), it seems something like:

```js
export default function getNodeScroll(node) {
  if (node === getWindow(node) || !isHTMLElement(node)) {
    return getWindowScroll(node);
  } else if (node.nodeName === 'BODY') {
    return {
      scrollLeft: 0,
      scrollTop: 0,
    };
  } else {
    return getHTMLElementScroll(node);
  }
}
```

fixes the issue on Edge while not breaking anything on other browsers.

---

Also, not really scroll related but 2 tests do have some spacing problems on edge:

chrome_BxPXP4meDR

chrome_narXLpTEeR

chrome_85agP7VvUV

---

I could attempt to test IE11, or add a try adding visual test for my failing case but when I have more free time.

src/dom-utils/getCompositeRect.js
```diff
@@ -24,15 +25,23 @@ export default function getCompositeRect(
...
-      ? getInnerOffsets(offsetParent)
-      : { x: 0, y: 0 };
+
+  let scroll = { scrollLeft: 0, scrollTop: 0 };
+  let offsets = { x: 0, y: 0 };
+
+  if (!isFixed) {
+    if (getNodeName(offsetParent) !== 'body') {
+      scroll = getNodeScroll(offsetParent);
+    }
```
atomiks: Shouldn't this be the same as your code @donnysim?

donnysim:
Oh yeah, it is, when I tested my project I changed the wrong file so I thought it's still broken 😳 everything seems to work fine on edge apart those couple gaps 😅.

atomiks:
It's probably the `getInnerOffsets` function above in the same file, it should be subtracting the borders – what is it reporting in Edge?

donnysim:
Seems like in this case it needs to be:

```
  return {
    top: toNumber(computedStyle.borderTopWidth),
    right: toNumber(computedStyle.borderRightWidth),
    bottom: toNumber(computedStyle.borderBottomWidth),
    left: toNumber(computedStyle.borderLeftWidth)
  };
```

`computedStyle.borderTop` etc. returns an empty string on Edge. The change does not seem to break in any browser (don't have anything to test on for Safari). It fixes all gaps on Edge.

atomiks:
Thanks for all the help 🙂

atomiks:
> Just a note, to be able to run on Windows with things live env changes you need cross-env:

Do you wanna make a PR for this btw?

donnysim:
> > Just a note, to be able to run on Windows with things live env changes you need cross-env:
> 
> Do you wanna make a PR for this btw?

Yeah, I might, currently trying get the visual tests work on IE 11 so if there's no rush I'll do it afterwards.

donnysim:
So IE11 is working (for the most part at least, don't have enough nerves to modify every template 😂).

for IE11 to work it requires:

```
Array.prototype.find
Array.prototype.includes
Object.assign
Object.entries
Promise
String.prototype.includes
Element.prototype.append
```
polyfills though, and you can only run the tests with basic ES5 syntax.

donnysim:
@atomiks I won't be doing a pull request on the run command, windows doesn't support ` ${DEV_PORT:-5000} ` expressions etc. and it just ends up with a bunch of unknown errors, like jest puppeteer throwing an `undefined` error, or serve receiving `null` port and so on 😐

PR #931, opened 2020-01-13
fix: setOptions should accept a partial

atomiks:


FezVrasta:
This would make impossible to unset some options though, or remove modifiers

atomiks:
It doesn't?

It lets the previous properties exist without overriding them - you can still override them with this technique.

Before:

```js
// resets `modifiers` and `strategy` back to default (unexpected) 
popper.setOptions({placement: 'right'});
```

With this change, _only_ placement is changed, as expected. If you want to unset things you can re-specify new values for the options. Also the type becomes correct with this change.

FezVrasta:
What about the modifiers? With this change you don't merge the old ones with the new ones.

atomiks:
Aren't shallow updates expected? It works the same as `this.setState()` in React for example

FezVrasta:
If that's the case I would expect `setOptions` to also accept the following signature: `prevOptions => newOptions` so that one can consume the previous options and alter them.

atomiks:
I don't think we need the function signature actually, since the prevOptions exist at `instance.state.options`:

```js
instance.setOptions({
  modifiers: [...instance.state.options.modifiers, newModifier]
});
```

atomiks:
> But, also, if that's the case, then I wouldn't expect the setOptions function to auto merge things, since the useState API of React doesn't do that.

This is different, because it's `options`, not `state`. Every option is optional (obviously), which means you should be able to pass a partial options object, since you can do that with the main constructor.

PR #938, opened 2020-01-13
fix: ship browser-compatible esm build

### LINKED ISSUE #933: Published NPM code references process.env.NODE_ENV
I am trying out rc1 and I hit an issue where browser code is trying to access `process.env.NODE_ENV`.

[This code](https://github.com/popperjs/popper.js/blob/master/src/index.js#L102) gets built into:

            if (process.env.NODE_ENV !== "production") {

I hit this when importing `@popperjs/core/lib/popper-lite` as described in the documentation [here](@popperjs/core/lib/popper-lite). Since this is intended as a client-side UI library, it would be great to not rely on Node.js variables.

FezVrasta:
fixes #933


cuberoot:
@FezVrasta This looks great. Thanks for the fast turn-around!

Issue #940, opened 2020-01-15
Arrow modifier cannot be used in web component


cuberoot:
### CodePen demo

https://codesandbox.io/s/popper-shadow-dom-uy1uu

### Steps to reproduce the problem

1. Load the codesandbox example above
2. You might have to tweak the code (e.g. change white space) to get around order-of-initialisation issues with the web component

### What is the expected behavior?

The arrow element (a small red square) should be lined up with the button like this

<img width="388" alt="Screen Shot 2020-01-15 at 5 45 41 PM" src="https://user-images.githubusercontent.com/154097/72405879-47235300-37bf-11ea-882b-c8c5c924b47b.png">

### What went wrong?

This console error is printed:

    Popper: "arrow" modifier's `element` must be a child of the popper element.

The red "arrow" is not positioned correctly

<img width="372" alt="Screen Shot 2020-01-15 at 5 46 07 PM" src="https://user-images.githubusercontent.com/154097/72405786-f0b61480-37be-11ea-8b93-3d71283c66e4.png">

### Any other comments?

This check is the issue: https://github.com/popperjs/popper.js/blob/master/src/modifiers/arrow.js#L67

Web Components (aka Custom Elements) use a shadow DOM to encapsulate their contents. The `contains` function does not delve into the shadow DOM.

PR #941, opened 2020-01-15
fix: allow arrow elements to be in shadow DOM

cuberoot:
This is a possible fix for issue #940

### DIFF
--- a/src/modifiers/arrow.js
+++ b/src/modifiers/arrow.js
@@ -64,7 +64,10 @@ function effect({ state, options, name }: ModifierArguments<Options>) {
     }
   }
 
-  if (!state.elements.popper.contains(arrowElement)) {
+  if (
+    !state.elements.popper.contains(arrowElement) &&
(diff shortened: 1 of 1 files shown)


src/modifiers/arrow.js
```diff
@@ -64,7 +64,10 @@ function effect({ state, options, name }: ModifierArguments<Options>) {
     }
   }
 
-  if (!state.elements.popper.contains(arrowElement)) {
+  if (
+    !state.elements.popper.contains(arrowElement) &&
+    (!state.elements.popper.shadowRoot || !state.elements.popper.shadowRoot.contains(arrowElement))
```
atomiks: Maybe extract `popper` to its own variable? For minifcation, unless gzip is larger.

FezVrasta: If we use .contains elsewhere in the code it may make sense to define our own `contains` utility to make sure we address this case everywhere

FezVrasta:
I'm merging this and then refactoring it to define a dedicated utility, thanks for the PR!

PR #942, opened 2020-01-15
Fixed typo in docs

semeleven:
I think, `area` was in the early implementation, but now, modificators don't have area option in the code


FezVrasta:
Thanks!

PR #952, opened 2020-01-18
fix: setOptions() should use forceUpdate

atomiks:
Makes `createPopper` sync

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -58,7 +58,7 @@ export type Instance = {|
   state: State,
   destroy: () => void,
   forceUpdate: () => void,
-  update: () => Promise<void>,
+  update: () => Promise<State>,
   setOptions: (options: $Shape<Options>) => void,
 |};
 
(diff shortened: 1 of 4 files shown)


src/types.js
```diff
@@ -59,7 +59,7 @@ export type Instance = {|
   destroy: () => void,
   forceUpdate: () => void,
   update: () => Promise<void>,
-  setOptions: (options: $Shape<Options>) => void,
+  setOptions: (options: $Shape<Options>) => Promise<void>,
```
FezVrasta: I think this should be Promise<State>

atomiks: Should `update` have that as well?

FezVrasta: yes

FezVrasta: While you are there, could you also update the docs types?

FezVrasta:
Manually merged, thanks

PR #954, opened 2020-01-20
feat: computeStyle dependent on current placement

### LINKED ISSUE #571: computeStyle dependant on current placement
Is it possible somehow to have `computeStyle.x` to be `top` when popper is going to be placed on `top`, and `bottom`, when actual placement is `bottom`?

atomiks:
Resolves #571 

When the popper is on top, and the height changes, there's no need to update its position, since it grows from the bottom. Same with when it's on the left and its width changes, since it grows from the right. 

It doesn't handle the case where its width changes on top/bottom, or height changes on left/right though, as that's not possible without updating.

---

Note, if you had CSS like:

```css
.popper {
  position: absolute;
  /* These break it if the style attr doesn't use them */
  top: 0;
  left: 0;
}
```

This will break it...

---

- Should it be disable-able?
- **How should we test this with the functional tests?**  Most tests are only written using the `right` placement, but it would be good to test top/left as well –– ideally each test would test all placements but then the tests will take forever


FezVrasta:
Considered we already have a flag internally I think it would be good to expose it in the modifier options too

src/modifiers/computeStyles.js
```diff
@@ -4,12 +4,25 @@ import type {
...
+import getOffsetParent from '../dom-utils/getOffsetParent';
+import getWindow from '../dom-utils/getWindow';
+import getDocumentElement from '../dom-utils/getDocumentElement';
+import getBasePlacement from '../utils/getBasePlacement';
 
 type Options = {
   gpuAcceleration: boolean,
 };
 
+const unsetSides = {
```
FezVrasta: Could we set these to `inherit` maybe so that we override eventual page provided styles?

PR #959, opened 2020-01-21
docs: fix links

atomiks:
Appending `v2/` broke these :\

### DIFF
--- a/docs/src/pages/docs/v2/modifiers/flip.mdx
+++ b/docs/src/pages/docs/v2/modifiers/flip.mdx
@@ -67,7 +67,7 @@ doesn't fit, it reverts back to the original placement.
 
 ### `padding`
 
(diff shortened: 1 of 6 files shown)


docs/src/pages/docs/v2/modifiers/flip.mdx
```diff
@@ -67,7 +67,7 @@ doesn't fit, it reverts back to the original placement.
 
 ### `padding`
 
-See [`padding` in Detect Overflow](/docs/utils/detect-overflow/#padding) for
+See [`padding` in Detect Overflow](/docs/v2/utils/detect-overflow/#padding) for
```
FezVrasta: Could you instead use ../utils/etc? So that if the position in the website change they wont break

atomiks: Didn't think that worked for some reason.. 🤦‍♂

Issue #957, opened 2020-01-21
contains test still not right for shadow DOM


cuberoot:
### CodePen demo

https://codesandbox.io/s/popper-shadow-dom-uy1uu

### Steps to reproduce the problem

1. Create a web component (custom element) where there are multiple layers of heirarchy
2. Apply popper to a component where the `tip` element is in the shadow DOM of a child of the popper element
3. Note that the `contains` function fails

### What is the expected behavior?

The tip should be found to be "contained" even if levels deep in the shadow DOM heirarchy

### What went wrong?

Error `Popper: "arrow" modifier's "element" must be a child of the popper element` was thrown

### Any other comments?

I originally submitted PR #941, I realised that it might not handle all cases. Sure enough, I refactored and hit an issue where the tip was two levels deep and my simple fix failed. A better PR will follow momentarily.

PR #961, opened 2020-01-21
fix: add more robust support for shadow dom

cuberoot:
Improve the `contains` utility function so that it can handle shadow DOM heirarchy.
Update jsdom for jest so that we can write shadow DOM tests.

This is a fix for my recently added issue #957


.config/jest.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
   },
   globalSetup: 'jest-environment-puppeteer/setup',
   globalTeardown: 'jest-environment-puppeteer/teardown',
-  testEnvironment: 'jsdom',
+  testEnvironment: 'jest-environment-jsdom-fourteen',
```
cuberoot: This upgrade allows Jest tests to use shadow DOM APIs.

src/dom-utils/contains.js
```diff
@@ -1,8 +1,27 @@
...
 
-export default function contains(parent: Element, child: Element): boolean {
-  return (
-    parent.contains(child) ||
-    (parent.shadowRoot ? parent.shadowRoot.contains(child) : false)
-  );
+export default function contains(parent: Element, child: Element) {
+  if (parent.contains(child)) return true;
+
+  try {
```
cuberoot: This is a trick that we use in our web component library. The easiest way to traverse up through arbitrary shadow DOM parents is to use event bubbling. We can tell whether the event was cancelled immediately which tells us that the parent handler was triggered.

src/dom-utils/getParentNode.test.js
```diff
@@ -28,10 +28,11 @@ it('fallbacks to documentElement if no parentNode can be extracted', () => {
 });
 
 // JSDOM doesn't support Shadow DOM unfortunately, polyfills didn't help
-xit('returns the shadow dom host', () => {
+it('returns the shadow dom host', () => {
```
cuberoot: I saw this test commented out. It works with the new jsdom environment with a few tweaks.

atomiks:
So much code \*__\*

This seems like it also might be expensive runtime-wise since it's in a [hot path](https://github.com/popperjs/popper-core/blob/78de552654dba5bd4a76064c38ef8f83b93f6304/src/dom-utils/getClippingRect.js#L48) - does shadowDOM even matter for this line of code?

If not, we should just remove the check for the arrow as a child of the popper as it's not worth it at all

cuberoot:
Yah it is a lot of code for what should be a simple check. Removing the check solves my problem too.

atomiks:
@cuberoot let's remove the whole `contains` utility and check

cuberoot:
@atomiks I see that the `contains` utility is also used here: https://github.com/popperjs/popper-core/blob/master/src/dom-utils/getClippingRect.js#L48

atomiks:
Yeah that's the line I linked in the previous comment. If you need dynamic clippingParents behavior and this has problems with shadow DOM, you can always specify your own boundary.

cuberoot:
@atomiks I can still rip out the `contains` utility, but I thought I would run this by you first. The `getParentNode` utility is shadow DOM smart, so we could use it to make a more compact `contains` function like this:
```javascript
// @flow
import getParentNode from './getParentNode';

  let last = child, next = child;
  do {
    last = next;
    next = getParentNode(next);
    if (next.isSameNode(parent)) return true;
  } while (next && next !== last);
  return false;
}
```

It's still a bit of code, though. So, still happy to go with the original plan of removing `contains`

cuberoot:
Hmm, I tried this version. It passes the unit tests, but fails some of the function tests. There must be some subtlety I am missing. I will just update the PR as you suggested.

cuberoot:
I am waffling again 😃  So, I dug deeper into `getClippingRect` and I see that it is using `getParentNode` directly and indirectly. As I mentioned, `getParentNode` is already smart about shadow DOMs. So, it is certainly possible for this to break with a hierarchy of web components.

I tweaked the above alternative implementation to one that is even smaller and that works. I now handle the case where a node is expected to contain itself (`contains(nodeA, nodeA) => true`).

```javascript
// @flow
  let next = child;
  do {
    if (next && next.isSameNode(parent)) return true;
    // $FlowFixMe: need a better way to handle this...
    next = next.parentNode || next.host;
  } while (next);
  return false;
}
```

src/dom-utils/contains.js
```diff
@@ -1,8 +1,10 @@
...
-
-export default function contains(parent: Element, child: Element): boolean {
-  return (
-    parent.contains(child) ||
-    (parent.shadowRoot ? parent.shadowRoot.contains(child) : false)
-  );
+export default function contains(parent: Element, child: Element) {
+  let next = child;
+  do {
+    if (next && next.isSameNode(parent)) return true;
```
atomiks: nit: we wrap curly braces around early returns like this for readability

cuberoot: Fixed. I force pushed to keep the commit history clean. So, check the diff again.

atomiks:
Great work! How does this compare perf-wise with regular `.contains()`?

cuberoot:
So, it looks like my solution is an order of magnitude slower than the built-in version with a tree 12-nodes deep: https://jsperf.com/popper-contains

These tests are kind of worst-case as it is unlikely that the parent will be 11 nodes above the child for most calls to contain.

So, this my version is slower, but on my machine I can still do 1.7 million ops per second.

I could add a call to the build-in `contains` first and then fall back to my implementation if that comes back false.

FezVrasta:
The fallback idea is good

cuberoot:
I made the change to try the built-in `contains` first and then fallback to iterating up the DOM if that fails. I also added a test for the case where you call `contains` with the same node as the child and parent.

cuberoot:
I also updated https://jsperf.com/popper-contains and you can see that the performance is pretty close now for the common case.

src/dom-utils/contains.js
```diff
@@ -1,8 +1,14 @@
 // @flow
+export default function contains(parent: Element, child: Element) {
+  if (parent.contains(child)) return true;
```
atomiks: Curly braces missing again

cuberoot: Fixed!

cuberoot:
Seems to be a failing visual test. I'll dig in tomorrow, I'm out of time here in New Zealand.

cuberoot:
Hmm, the same test fails for me locally on `master` without my changes. Not sure what is up there...

FezVrasta:
It just may be a flaky test, I’ll take a look in a bit

FezVrasta:
I think we could use .getRootNode() to check if the child is in a shadow root, and use this info to decide what logic to run

atomiks:
> .getRootNode()

Doesn't work in Edge/IE11, this should be fine. The flaky test is the recent grow one, we can fix it after

FezVrasta:
@atomiks wouldn't this work? It should keep the function fast for non shadow DOM elements, and gracefully fallback to the shadow DOM compatible check if necessary.

```js
// @flow
  const isShadow = child.getRootNode && child.getRootNode().host;

  if (parent.contains(child)) {
    return true;
  } else if (!isShadow) {
    return false;
  } else {
    let next = child;
    do {
      if (next && next.isSameNode(parent)) {
        return true;
      }
      // $FlowFixMe: need a better way to handle this...
      next = next.parentNode || next.host;
    } while (next);
    return false;
  }
}
```

atomiks:
I guess it should work, maybe like this?

```js
// @flow
  // Shadow DOM
  if (child.getRootNode && child.getRootNode().host) {
    let next = child;
    do {
      if (next && next.isSameNode(parent)) {
        return true;
      }
      // $FlowFixMe: need a better way to handle this...
      next = next.parentNode || next.host;
    } while (next);
    return false;
  }

  return parent.contains(child);
}
```

FezVrasta:
Even if we detect the child is in a shadow root, nothing prevents `parent` to be in the same shadow root, making the native `.contains()` work. For this reason I would always first attempt to use the native method as in my code.

PR #966, opened 2020-01-23
docs: contain landing page samples scroll

pascalduez:
hi,

just a small "improvement": prevent landing page samples from scrolling the whole page when their inner scroll reach the sample boundaries.


FezVrasta:
Thanks, good addition!

PR #970, opened 2020-01-25
fix: take into account document element scrollbar

### LINKED ISSUE #969: scrollbar not calculated correctly, causing an overflow
Positioning a tooltip very close to the viewport boundary can cause an overflow.

https://jsfiddle.net/bzs7yfq4/1/

https://imgur.com/a/FPnbu21

## Observation
Positioning a tooltip close to the boundary clips the tooltip and causes an overflow.

## Expectation

The tooltip is flipped to the left to not cause the page to overflow. This happens after resizing the viewport further. 
Coincidentally the issue happens over a span of the scrollbar width. 
Potentially the scrollbar width is not taken into consideration on positioning.

```
popper version: 2.0.0
chrome:  79.0.3945.117 
macOS 10.14.6
```

atomiks:
Fixes #969

Is this approach sound? If the page is horizontally scrollable, `documentElement` doesn't seem to get affected, only `body`. So this should work.

### DIFF
--- a/src/dom-utils/getDecorations.js
+++ b/src/dom-utils/getDecorations.js
@@ -1,15 +1,26 @@
 // @flow
 import type { SideObject } from '../types';
 import getBorders from './getBorders';
+import getNodeName from './getNodeName';
+import getWindow from './getWindow';
 
 // Borders + scrollbars
 export default function getDecorations(element: HTMLElement): SideObject {
   const borders = getBorders(element);
+  const win = getWindow(element);
+
+  let right = element.offsetWidth - element.clientWidth - borders.right;
+  let bottom = element.offsetHeight - element.clientHeight - borders.bottom;
+
+  if (getNodeName(element) === 'html') {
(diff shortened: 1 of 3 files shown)


src/dom-utils/getDecorations.js
```diff
@@ -1,15 +1,24 @@
...
 // Borders + scrollbars
 export default function getDecorations(element: HTMLElement): SideObject {
   const borders = getBorders(element);
 
+  let right = element.offsetWidth - element.clientWidth - borders.right;
+  let bottom = element.offsetHeight - element.clientHeight - borders.bottom;
+
+  if (getNodeName(element) === 'html') {
+    right = window.innerWidth - element.clientWidth;
+    bottom = window.innerHeight - element.clientHeight;
```
atomiks: I don't think taking into account `border` on `<html>` is worth it or makes sense...  since I don't think anyone ever does that

PR #971, opened 2020-01-25
fix(types): typechecking for built in modifier options

atomiks:
Enables autocomplete as well for TS. Would be good to get `modifiersData` types as well 🤔 I don't know why I need to disable eslint for the Options exports...

edit: the `any` modifier type prevents typechecking correctly for other modifiers' options... somehow would need to exclude the built-in modifier name strings but I don't think that is possible

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,9 +2,9 @@
 import type {
   State,
   Options,
-  Modifier,
   Instance,
   VirtualElement,
+  Modifiers,
 } from './types';
 import getCompositeRect from './dom-utils/getCompositeRect';
 import getLayoutRect from './dom-utils/getLayoutRect';
@@ -33,7 +33,7 @@ const DEFAULT_OPTIONS: Options = {
 };
 
 type PopperGeneratorArgs = {
-  defaultModifiers?: Array<Modifier<any>>,
+  defaultModifiers?: Modifiers,
   defaultOptions?: $Shape<Options>,
 };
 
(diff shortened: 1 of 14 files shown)


atomiks:
In TypeScript: I believe we need to let users "opt-in" to strict mode using a generic modifiers type, which will force them to type their custom modifiers. Doing so, we enable typechecking and autocomplete for the built-in modifiers.

```ts
export type Offset = Omit<Partial<Modifier<any>>, 'name' | 'options'> & {
  name: 'offset',
  options?: {
    offset?: [number, number]
  }
};

export type ComputeStyles = Omit<Partial<Modifier<any>>, 'name' | 'options'> & {
  name: 'computeStyles',
  options?: {
    adaptive?: boolean;
    gpuAcceleration?: boolean;
  }
};

// etc for the other modifiers

export type StrictModifiers = Offset | ComputeStyles;

export declare type Options<TModifier = Partial<Modifier<any>> = {
  // ...
  modifiers: Array<TModifier>;
};

export declare const createPopper: <TModifier = Partial<Modifier<any>>>(reference: Element | VirtualElement, popper: HTMLElement, options?: Partial<Options<TModifier>>) => Instance;
```

```ts
import {StrictModifiers} from '@popperjs/core';

// They can extend `StrictModifiers` with their own custom modifier types
createPopper<StrictModifiers>(reference, popper, {
 // typed modifiers
});
```

---

I noticed in VSCode though, that if you got one of the modifier options type wrong, red squiggles appeared at other modifiers even if they were valid. Somehow it needs be "narrowed" correctly.

PR #972, opened 2020-01-25
fix: add margin warning

atomiks:
When upgrading from Popper v1 to Popper v2, the consumer can leave `margin` on their popper and cause problems since we no longer consider it. 

I've seen a library do this when upgrading today when searching on GitHub and they don't realize the bug exists (their margin is somewhat small).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -127,9 +127,20 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
             }
           }
 
+          const {
+            marginTop,
+            marginRight,
+            marginBottom,
+            marginLeft,
+          } = getComputedStyle(popper);
+
           // We no longer take into account `margins` on the popper, and it can
           // cause bugs with positioning, so we'll warn the consumer
-          if (parseFloat(getComputedStyle(popper).margin)) {
+          if (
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -125,6 +126,31 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
...
+          const {
+            marginTop,
+            marginRight,
+            marginBottom,
+            marginLeft,
+          } = getComputedStyle(popper);
+
+          // We no longer take into account `margins` on the popper, and it can
+          // cause bugs with positioning, so we'll warn the consumer
+          if (
```
FezVrasta: Shouldn’t this be on dev only?

atomiks: It is

src/index.js
```diff
@@ -125,6 +126,31 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
               );
             }
           }
+
+          const {
+            marginTop,
```
FezVrasta: I remember IE wanted us to use marginTopWidth I think?

atomiks: That's only for `border`, since it's a shorthand property. This is the full margin name

PR #978, opened 2020-01-27
fix: rtl scrollbar

### LINKED ISSUE #976: Scrollbar not taken into account for RTL direction (scrollbar on the left)
### JSfiddle demo

https://jsfiddle.net/26pu5sjn/

https://i.imgur.com/pjc6yNE.png

1. have css direction: rtl; overflow: scroll; to the body
2. a scrollbar will be shown on the left side of the screen
3. Add a tooltip with positioning bottom-end;
3. the tooltip will be shifted by the width of the scrollbar to the right.

### What is the expected behaviour?

The tooltip should not be shifted by the right if there is a scrollbar on the left side.

### What went wrong?

The scrollbar on the left side is not taken into the position calculation.

### Any other comments?

Thank you very much for fixing https://github.com/popperjs/popper-core/issues/969 so quickly.

atomiks:
Should fix #976

### DIFF
--- a/src/dom-utils/getCompositeRect.js
+++ b/src/dom-utils/getCompositeRect.js
@@ -1,21 +1,9 @@
 // @flow
-import type { Rect, VirtualElement, Offsets } from '../types';
+import type { Rect, VirtualElement } from '../types';
 import getBoundingClientRect from './getBoundingClientRect';
 import getNodeScroll from './getNodeScroll';
 import getNodeName from './getNodeName';
-import getBorders from './getBorders';
 import { isHTMLElement } from './instanceOf';
-
-// offsets without `border`
-function getInnerOffsets(offsetParent: Element): Offsets {
-  const rect = getBoundingClientRect(offsetParent);
-  const borders = getBorders(offsetParent);
-
-  return {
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -182,6 +182,11 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
           return;
         }
 
+        // HACK:
+        // `offsetLeft` must be native on each read, not due to popper's
+        // positioning if `gpuAcceleration` is disabled
+        popper.style.left = '0';
+
```
atomiks: `.offsetLeft` needs to read the default (so it's just the scrollbar, if present), so we need to write to the DOM here to undo the popper offsets. Any ideas for a better technique for this? :\

atomiks: I am guessing if the popper has has scrollbar itself, the arrow also needs this too 🤔

atomiks:
^Found the correct solution, use `clientLeft`

FezVrasta:
TS is failing :-(

atomiks:
Empty force-pushed, it seemed to be a package install error

PR #982, opened 2020-01-28
fix: reduce polyfills needed on IE11

FezVrasta:
This should address the concerns expressed in https://github.com/mui-org/material-ui/issues/19358

The required polyfills for IE11 now are just `Promise` and `Object.assign`.

The brotli-compressed bundles ended up being lighter for some reason, while the gzip ones are ~0.01kB larger

cc @eps1lon and @oliviertassinari

### DIFF
--- a/src/utils/computeAutoPlacement.js
+++ b/src/utils/computeAutoPlacement.js
@@ -43,7 +43,7 @@ export default function computeAutoPlacement(
     ? flipVariations
       ? variationPlacements
       : variationPlacements.filter(
-          placement => placement.indexOf(variation) >= 0
+          placement => getVariation(placement) === variation
         )
     : basePlacements;
 
--- a/src/utils/computeAutoPlacement.js
+++ b/src/utils/computeAutoPlacement.js
@@ -42,7 +42,9 @@ export default function computeAutoPlacement(
   const placements = variation
     ? flipVariations
       ? variationPlacements
-      : variationPlacements.filter(placement => placement.includes(variation))
+      : variationPlacements.filter(
+          placement => placement.indexOf(variation) >= 0
+        )
     : basePlacements;
 
   // $FlowFixMe: Flow seems to have problems with two array unions...
--- a/docs/src/components/Landing.js
+++ b/docs/src/components/Landing.js
@@ -386,7 +386,7 @@ createPopper(popcorn, tooltip, {
             (placements, basePlacement) => [
               ...placements,
               // clockwise tabbing order
-              ...(['bottom', 'left'].includes(basePlacement)
+              ...(['bottom', 'left'].indexOf(basePlacement) >= 0
                 ? [
                     `${basePlacement}-end`,
                     basePlacement,
--- a/docs/src/pages/docs/v2/browser-support.mdx
+++ b/docs/src/pages/docs/v2/browser-support.mdx
@@ -17,7 +17,7 @@ IE11 (and older browsers in general) require polyfills to work. The simplest way
 to make Popper work is to use the following polyfill service:
 
 ```html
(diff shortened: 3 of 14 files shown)


atomiks:
`findIndex` was introduced with `find` though

FezVrasta:
I wanted to use indexOf 🤦‍♂️ I’ll fix that later

atomiks:
We should keep `find` I think, also I think there's a `String.prototype.includes` usage in `computeAutoPlacement` remaining.

FezVrasta:
Maybe we can find a Babel polyfill to convert them at runtime without adding too much bloat, considering the file size is the same I wouldn’t mind

FezVrasta:
Maybe like this one

https://www.npmjs.com/package/babel-plugin-array-includes

atomiks:
> https://www.npmjs.com/package/babel-plugin-array-includes

The "pitfalls" section doesn't look good

Keep `Promise` + `Object.assign` + `find` as required polyfills should be fine imo. `indexOf` to replace `includes` is fine

FezVrasta:
I removed `Object.entries` and `String.prototype.includes` for now, and the polyfills bundle is at 4.9kB gzipped.

I'll remove `Array.prototype.includes` too now.

src/utils/computeAutoPlacement.js
```diff
@@ -42,7 +42,9 @@ export default function computeAutoPlacement(
   const placements = variation
     ? flipVariations
       ? variationPlacements
-      : variationPlacements.filter(placement => placement.includes(variation))
+      : variationPlacements.filter(
+          placement => placement.indexOf(variation) >= 0
+        )
```
atomiks: We can use `placement => getVariation(placement) === variation` I think

atomiks:
> edit: removed it, the polyfill is 4.8kB now

95%+ of apps are polyfilling `Promise` if they need to anyway so it's not really something that should be considered tbh...

FezVrasta:
Looking at the polyfills on Bundlephobia, `promise-polyfill` is 1.1kB, `array-find-polyfill` is 319B, and `object-assign` is 581B, so the total is 2kB of polyfills. I'm not sure why `polyfill.io` adds so much weight.

eps1lon:
I appreciate the effort but we can't just switch to it in v4. We can use it in v5 when we revisit our bundle strategy (and likely ship a separate IE 11 build that requires polyfills).

Object.assign can be transformed at runtime (e.g. babel-transform-object-assign). Array.prototype.find is something that you could inline.

> > edit: removed it, the polyfill is 4.8kB now
> 
> 95%+ of apps are polyfilling `Promise` if they need to anyway so it's not really something that should be considered tbh...

Good point though I'm interested where the number comes from. Overall I'm more interested in how this is used. If these are only new features then we can safely use it.

FezVrasta:
I’m thinking we could make the promise optional but it would confuse the API quite a lot

oliviertassinari:
@FezVrasta Thanks for reconsidering the tradeoff. I'm happy to see that fewer polyfills are required now :).

Have you considered to completely remove the need for polyfills 👌? For instance, the Promise seems to be used exclusively for the update API. Is this for leveraging [micro](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/)[tasks](https://www.youtube.com/watch?v=cCOL7MC4Pl0)? Does it work with the polyfill?

FezVrasta:
We use `Promise` both in the `update` method, as a way to enable consumers to execute code after Popper positioned the tooltip, and as alternative to [`queueMicrotask`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/queueMicrotask) in the `debounce` utility.

The polyfill won't mimic the microtask (AFAIK) but will fall back to the best alternative debounce available in the browser, which is good enough for us.

We could perform feature detection in the `debounce` utility, but the `update()` API is exposed to the consumers, so we can't do anything to avoid the requirement there.

PR #987, opened 2020-01-31
fix: rtl body scrollbar in edge cases and iOS bottom bar

atomiks:
## RTL

#978 worked for scrolling containers. These browsers are fine in all cases: Edge (2019), Safari, and likely IE11 if it works like Edge. 

Chrome and Firefox work by default for `<body>`, assuming their scrollbar is on the right for RTL (it seems to be). But it appears on the left for them in these cases, and is still offset incorrectly by the scrollbar width:

- `<iframe>` body context
- Firefox setting `layout.scrollbar.side = 3` 

## iOS bottom bar

The check `window.innerHeight - document.documentElement.clientHeight` to get the bottom scrollbar height causes a problem on iOS, explained in the comment

### DIFF
--- a/src/dom-utils/getCompositeRect.js
+++ b/src/dom-utils/getCompositeRect.js
@@ -4,12 +4,14 @@ import getBoundingClientRect from './getBoundingClientRect';
 import getNodeScroll from './getNodeScroll';
 import getNodeName from './getNodeName';
 import { isHTMLElement } from './instanceOf';
+import getWindowScrollBarX from './getWindowScrollBarX';
 
 // Returns the composite rect of an element relative to its offsetParent.
 // Composite means it takes into account transforms as well as layout.
 export default function getCompositeRect(
   elementOrVirtualElement: Element | VirtualElement,
   offsetParent: Element,
+  documentElement?: HTMLElement,
   isFixed: boolean = false
 ): Rect {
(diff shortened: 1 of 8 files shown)


src/dom-utils/getCompositeRect.js
```diff
@@ -4,12 +4,14 @@ import getBoundingClientRect from './getBoundingClientRect';
...
 import getNodeName from './getNodeName';
 import { isHTMLElement } from './instanceOf';
+import getWindowScrollBarX from './getWindowScrollBarX';
 
 // Returns the composite rect of an element relative to its offsetParent.
 // Composite means it takes into account transforms as well as layout.
 export default function getCompositeRect(
   elementOrVirtualElement: Element | VirtualElement,
   offsetParent: Element,
+  documentElement?: HTMLElement,
```
FezVrasta: Could we import `getDocumentElement` in this module and run it on `offsetParent`? `popper` and its `offsetParent` should have the same `documentElement` anyways.

This would keep the utility interface clean

atomiks: `offsetParent` can be `Window` which is the problem

FezVrasta: can we change `getDocumentElement` to check for `window.document.documentElement` too?

FezVrasta:
Manually merged

PR #994, opened 2020-02-09
fix(applyStyles): don't remove reference inline styles

atomiks:
Inline styles specified in `initialStyles` are getting removed from the reference element upon destroy. Other line changes are small perf tweaks

### DIFF
--- a/src/modifiers/applyStyles.js
+++ b/src/modifiers/applyStyles.js
@@ -47,21 +47,21 @@ function effect({ state }: ModifierArguments<{||}>) {
   return () => {
     Object.keys(state.elements).forEach(name => {
       const element = state.elements[name];
-      const styleProperties = Object.keys(
-        state.styles.hasOwnProperty(name)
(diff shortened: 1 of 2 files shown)


src/modifiers/applyStyles.js
```diff
@@ -73,9 +73,9 @@ function effect({ state }: ModifierArguments<{||}>) {
       // $FlowFixMe
       Object.assign(element.style, style);
 
-      Object.keys(attributes).forEach(attribute =>
-        element.removeAttribute(attribute)
-      );
+      Object.keys(attributes).forEach(attribute => {
```
FezVrasta: what's the reason for the additional curly braces here?

atomiks: It removes `return` statement when compiled

atomiks:
Should we also be applying `position` to the arrow in the effect like the popper? Currently you need to specify it in CSS first

PR #996, opened 2020-02-09
fix: merge modifiers by name

atomiks:
As mentioned on Spectrum, the new data structure allows for this and should be done by default I think. It's not too large and handles custom modifiers.

Modifiers later in the array override earlier modifiers in the array. It also deep-merges `data` and `options` objects, but not their values. Expected behavior? Will it break existing merge setups?

### DIFF
--- a/src/utils/mergeByName.js
+++ b/src/utils/mergeByName.js
@@ -1,0 +1,35 @@
+// @flow
+import type { Modifier } from '../types';
+
+export default function mergeByName(
+  modifiers: Array<$Shape<Modifier<any>>>
+): Array<$Shape<Modifier<any>>> {
+  const modifiersMap = new Map();
+
+  modifiers.forEach(modifier => {
+    const currentModifier = modifiersMap.get(modifier.name);
+
+    modifiersMap.set(
+      modifier.name,
+      currentModifier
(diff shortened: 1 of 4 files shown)


src/utils/mergeByName.js
```diff
@@ -0,0 +1,35 @@
...
+    modifiersMap.set(
+      modifier.name,
+      currentModifier
+        ? {
+            ...currentModifier,
+            ...modifier,
+            // $FlowFixMe
+            options: { ...currentModifier.options, ...modifier.options },
+            // $FlowFixMe
+            data: { ...currentModifier.data, ...modifier.data },
```
atomiks: Needs fixing

FezVrasta:
| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/umd/popper-base.min.js` | 1.8 kB | +29 B (1%) |  |
| `dist/umd/popper-lite.min.js` | 2.86 kB | +38 B (1%) |  |
| `dist/umd/popper.min.js` | 5.29 kB | +4 B (0%) |  |


<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>

PR #1000, opened 2020-02-10
fix(flip): remove uniqueBy

atomiks:
🚨**I'm not 100% sure of this**

I think I added this due to `auto` but in v1 you could specify a placement again, but I think that was only because v1 didn't revert to the originally preferred placement:

```js
behavior: ['top', 'bottom', 'top']
```

This removes that ability since it only contains unique placements. Does it make sense to filter dupes out or not?

```js
// this expands to the four base placements, and contains another 
// "right" somewhere (where it has the most space)
fallbackPlacements: ['right', 'auto']
```

I think it will just get ignored in v2 so it doesn't affect anything — this change will remove the `uniqueBy` function completely with production DCE and improve bundle+runtime perf (except potentially in the worst case with lots of repeated/dupe placements).


FezVrasta:
I think it’s okay to allow duplicate placements, I’d be in favor to merge it

atomiks:
We can merge then I guess

PR #1004, opened 2020-02-11
perf: compile esm entry files with rollup

atomiks:
With the Babel-compiled modules, webpack adds a bunch of `Object(o.s)`-type calls and is inefficient, leading to a 6.4 kB default size (what bundlephobia reports). This leads to 500B+ of bloat in the resulting bundle. 

If we compile the entry files with Rollup instead, not only will compilation be faster for consumers (single file), the bundle size is also much better.

I'm quite sure this is fully backwards-compatible, after testing with my repo it works the exact same as before but is faster to compile.

Before:

5.81 kB

After:

6.36 kB

---

Note if the user is using tree-shaking then they'll be using the Babel-compiled modules when importing individual modifiers; we probably have to Rollup those files for better perf and to prevent module duplication...

### DIFF
--- a/.config/rollup.config.js
+++ b/.config/rollup.config.js
@@ -9,12 +9,14 @@ import compiler from '@ampproject/rollup-plugin-closure-compiler';
 import pkg from '../package.json';
 
 const getFileName = input => input.split('/')[1].split('.')[0];
+const getDir = (dir, format) => (format === 'esm' ? dir : `${dir}/${format}`);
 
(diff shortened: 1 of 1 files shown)


atomiks:
So we can't do this due to module duplication... even if we only did this to the full file, if they import `detectOverflow` util then there's +200 kB of duplication bloat (and removes all of this benefit). Lite/Base consumers can't do it due to that in the first place due to duplication.

I wonder why webpack is produces that type compilation bloat when compiling the modules together, since Rollup doesn't do it 🤷‍♂

Issue #986, opened 2020-01-30
clippingParents for popper in body


semeleven:
## Feature description
Now `clippingParents` works fine, when popper is siblings with reference. But when popper in the body, `clippingParents` calculate body is a parent. But it is not what I expect. I expect to calculate reference overflow parent in this case. I think will be good to add some flag, like `inBody`, to calculate parents from reference. Now we have `altBoundary` flag, but it is, as far as I understand, for another mission.

## Why should this feature be part of the Popper's core?
Render popper in the body, I think, a common case when people work with overflow containers. Now we have `preventOverflow` modifier for overflow case. If handmade set right overflow parent to `boundary` - all fine. But I think, will be better has current sugar `clippingParents` for this case.

FezVrasta:
Thanks for the report, may you provide an example ([you can use this sandbox](https://codesandbox.io/s/popper-2-playground-hgjuz)) to help me understand what are you proposing?

atomiks:
`"scrollParent"` in v1 was a very problematic default, see: https://github.com/codesandbox/codesandbox-client/issues/3059

`altBoundary` is fine to use if you want it to check the reference's `"clippingParents"` instead.

PR #1010, opened 2020-02-19
added alt boundary options for modifiers (prevent overflow and flip)

semeleven:
In #986 I requested new flag for case, when popper render outside scroll parent, for correct work `clippingParents` options. @atomiks told `altBoundary` fine for this mission, but this option not pass to `detectOverflow` function. I fixed that 😃


FezVrasta:
Thanks this looks great! I'll wait for @atomiks review before merging just to make sure I'm not missing anything.

docs/src/pages/docs/v2/modifiers/prevent-overflow.mdx
```diff
@@ -127,6 +128,24 @@ createPopper(reference, popper, {
...
+    {
+      name: 'preventOverflow',
+      options: {
+        altBoundary: true,
+      },
+    },
+  ],
+});
+```
+
```
atomiks: Should we comment out before release? @FezVrasta have you thought of a solution to the manual npm package releases vs. automated docs deploying? I've discussed it at https://github.com/atomiks/tippyjs/pull/692

src/modifiers/preventOverflow.js
```diff
@@ -28,6 +28,8 @@ type Options = {
   boundary: Boundary,
   /* If the popper is not overflowing the main area, fallback to this one */
   rootBoundary: RootBoundary,
+  /* If popper render outside overflow parent, example */
+  altBoundary: boolean,
```
atomiks: This comment should be something like, `Use the reference's "clippingParents" boundary context`

src/modifiers/preventOverflow.js
```diff
@@ -45,12 +47,18 @@ function preventOverflow({ state, options, name }: ModifierArguments<Options>) {
     altAxis: checkAltAxis = false,
     boundary,
     rootBoundary,
+    altBoundary = false,
```
atomiks: You can leave off the `= false`, as `undefined` gets resolved to the default (`false`) automatically

src/modifiers/preventOverflow.js
semeleven: fixed

src/modifiers/preventOverflow.js
semeleven: fixed

docs/src/pages/docs/v2/modifiers/flip.mdx
```diff
@@ -122,6 +122,24 @@ createPopper(reference, popper, {
 });
 ```
 
+### `altBoundary`
+
+See [`altBoundary` in Detect Overflow](../../utils/detect-overflow/#altboundary)
+for details.
+
```
atomiks: Can we add the following (for `preventOverflow` as well):

> This will check the reference's boundary context (`clippingParents`) instead of the popper's, effectively replicating the `scrollParent` boundary from Popper v1.

docs/src/pages/docs/v2/modifiers/flip.mdx
```diff
@@ -122,6 +122,24 @@ createPopper(reference, popper, {
...
+```js
+createPopper(reference, popper, {
+  modifiers: [
+    {
+      name: 'preventOverflow',
+      options: {
+        altBoundary: true,
```
atomiks: Add `// false by default` comment (inlined next to it) (and to `preventOverflow` as well)

semeleven:
Added for flip, but now I don't sure about altBoundary, because it is not same behavior like popper in the scroll element. I find this rule https://github.com/popperjs/popper-core/blob/master/src/dom-utils/getClippingRect.js#L34 . Result - we have different clipingRect, because popper always has absolute style, but reference usually not have this. In this [example](https://codesandbox.io/s/cool-worker-1wy7s) (hands set boundary, before has altBoundary options) we have horizontal scroll, but when we disabled this, horizontal scroll not have. Maybe it is another bug for fallback rootBoundary, I am not sure.

docs/src/pages/docs/v2/modifiers/flip.mdx
semeleven: done

docs/src/pages/docs/v2/modifiers/flip.mdx
semeleven: done

atomiks:
The reason is unrelated to `altBoundary`, it's because it's being prevented at the scroll boundary, and I believe `top/left` are preferred in ambiguous case where it can't fit over `right/bottom` (so it will prefer to overflow the right, rather than the left scroll boundary). 

I guess it could be smarter in that case...

docs/src/pages/docs/v2/modifiers/flip.mdx
```diff
@@ -122,6 +122,29 @@ createPopper(reference, popper, {
...
+[`scrollParent`](../../../v1/#preventoverflowboundarieselement) boundary from
+Popper v1. See
+[`altBoundary` in Detect Overflow](../../utils/detect-overflow/#altboundary) for
+details.
+
+```js
+createPopper(reference, popper, {
+  modifiers: [
+    {
+      name: 'preventOverflow',
```
atomiks: This should be `flip`

semeleven: sorry, fixed

semeleven:
Can I help with anything for release this PR?

FezVrasta:
Sorry, it slipped through the cracks. I'm merging it now, I'll cut a new release this week.

PR #1029, opened 2020-03-14
fix: arrow position using offsetParent client offset

atomiks:
A tweak to #1023

The arrow's `offsetParent` may not always be the popper, for example with the following DOM:

```html
<div id="popper"> <!-- wrapper positioned node with no styling -->
  <div id="box"> <!-- offsetParent, with all styling/borders and animations -->
    <div id="arrow"></div>
  </div>
</div>
```

This still handles the case where the popper is the `offsetParent` (the existing test)

What it doesn't handle is the case where the popper is not the `offsetParent`, but it has a border/scrollbar. (This also includes any arbitrary parentNodes of the `offsetParent` that have borders/scrollbars I think).

In the cases I am using it, the `offsetParent` is always the one with the border & styling. I'm not sure if handling all arbitrary parents' client offsets is common or worth it...

### DIFF
--- a/src/modifiers/arrow.js
+++ b/src/modifiers/arrow.js
@@ -3,6 +3,7 @@ import type { Modifier, ModifierArguments, Padding } from '../types';
 import getBasePlacement from '../utils/getBasePlacement';
 import getLayoutRect from '../dom-utils/getLayoutRect';
 import contains from '../dom-utils/contains';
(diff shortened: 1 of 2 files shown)


atomiks:
@FezVrasta may you publish a patch release with this? Some bordered arrow styles require this with image snapshot tests 😓

FezVrasta:
2.1.1, done

PR #1032, opened 2020-03-20
feat: arrow edge data

atomiks:
This property allows users to create a modifier to hide the arrow once it reaches the edge of the popper (useful when `preventOverflow#tether = false`). The arrow won't be pointing to the center of the reference anymore once it's prevented from overflowing the popper (at its edge), so it is desirable to hide it. https://twitter.com/hakimel/status/1240927891254378496 (a transition would look nicer, but that's not related to this feature since it's simply pure data)

Maybe there's also a case where they only want to hide it if it's going to leave the reference bounds (even if it's not centered), but that should be applicable with a custom modifier.

I also refactored the client logic to be simpler and fixed an edge case bug with borders (`max` value) I found while developing this feature.

IIRC, Tether also has this feature on its demo page.

```js
    modifiers: [
      {
        name: 'applyArrowHide',
        enabled: true,
        phase: 'write',
        fn({ state }) {
          if (state.modifiersData.arrow.centerOffset !== 0) {
            state.elements.arrow.setAttribute('data-hide', '');
          } else {
            state.elements.arrow.removeAttribute('data-hide');
          }
        },
      },
      {
        name: 'preventOverflow',
        options: {
          tether: false,
        },
      },
    ],
  });
```

Since it describes how much the arrow is offset from the "true" center of the reference due to being prevented from overflowing the popper, this should enable shape interpolation seen in https://i.imgur.com/y4pJsrK.mp4


atomiks:
Anyone know how to create this effect with the arrow: https://i.imgur.com/y4pJsrK.mp4? It twists and distorts the shape such that it can point to the reference, with would not otherwise be possible while the arrow is prevented from overflowing its popper.

Is it possible to add with the existing features? Or do we need to provide some other data to allow the user to do it?

docs/src/pages/docs/v2/modifiers/arrow.mdx
```diff
@@ -89,7 +89,40 @@ createPopper(reference, popper, {
...
+  name: 'applyArrowHide',
+  enabled: true,
+  phase: 'main',
+  fn({ state }) {
+    if (!state.elements.arrow) {
+      return;
+    }
+
+    if (state.modifiersData.arrow.centerOffset !== 0) {
+      state.elements.arrow.removeAttribute('data-hide', '');
```
atomiks:
```suggestion
      state.elements.arrow.setAttribute('data-hide', '');
```

🙄
 
@FezVrasta

docs/src/pages/docs/v2/modifiers/arrow.mdx
```diff
@@ -89,7 +89,40 @@ createPopper(reference, popper, {
 
 ```flow
 type Data = {
+  // Describes the offset of the arrow
   x?: number,
   y?: number,
+  // Describes the offset of the arrow relative to its "true" center (occurs
+  // once the arrow reaches the edge of its popper and cannot overflow it to be
+  // centered to the reference)
+  centerOffset: number,
```
atomiks: `non-zero` instead of `occurs` might be better I guess..

docs/src/pages/docs/v2/modifiers/arrow.mdx
FezVrasta: Could we use the `applyStyles` modifier to set/unset this attribute?

atomiks: Not sure if we should bother, it will add more bloat in the case where the user doesn't need it + the interpolation technique won't use the attribute I think.

Issue #1031, opened 2020-03-19
Flow types do not work automatically


mcous:
The Popper v2 [documentation for usage with Flow](https://popper.js.org/docs/v2/typings/#flow) states:

> Rather than exporting limited type definitions, we link to the Flow-typed source files from the dist/umd folder.
>
> Flow will automatically read the exposed types, so you don't need to change any configuration in your project to take benefit from Popper typings.

In practice, I'm not finding this to be true.

### Steps to reproduce the problem

1. Set up the Popper recommended `.flowconfig`
    ```
    [declarations]
    ; read type signatures from node_modules but ignore flow errors within
    <PROJECT_ROOT>/node_modules/.*
    ```
2. Create a file that imports from Popper core
    ```js
    // @flow
    import { createPopper } from '@popperjs/core'
    ```
3. Hover over `createPopper` or try to assign it to a typed variable and notice that `createPopper` is typed as `any`

### What is the expected behavior?

- `createPopper` should be typed according to the types in its source code as stated in documentation

### What went wrong?

- `createPopper` is not typed

### Any other comments?

I'm pretty sure I'm not doing anything wrong, but please let me know if I am, in fact, doing something wrong!

From what I can tell, this problem is caused by the fact that the main types file `popper.js.flow` is located at `umd/popper.js.flow`, but `package.json::main` is `cjs/popper.js`, and there is no `cjs/popper.js.flow`. This seems like a fairly straightforward build misconfiguration in the rollup config, so I'm happy to PR a fix if this is the case!

I am also able to work around this by making the following changes in my `.flowconfig`:

```
[options]
module.system.node.main_field=module
module.system.node.main_field=main
```

I think this works because `package.json::module` is `lib/popper.js`, and `lib/popper.js.flow` exists. Even with that workaround, however, I'm still not able to do something like this:

```js
import type { Instance } from '@popperjs/core'
```

Instead, I have to do

```js
import type { Instance } from '@popperjs/core/lib'
```

`lib/index.js` / `lib/index.js.flow` exports all the types, but `lib/popper.js` / `lib/popper.js.flow` (the `package.json::module` entry) does not export any types.

Finally, I'm not sure what `dist/esm` is for compared to `lib`, and I don't see it referenced anywhere.

FezVrasta:
Thanks for noticing! The types should be copied to `cjs`, a PR to fix that would be awesome.

mcous:
@FezVrasta great, I can have a PR shortly!

[...]

PR #1036, opened 2020-03-23
fix(types): Export Flow type definitions from entry points

mcous:
As discussed in #1031, this PR makes two changes:

- Updates the Rollup config to ensure Flow types are shipped with the CommonJS build, which is the `main` of `package.json`
- Ensures type definitions, which will be stripped by babel in the build, are exported from the three bundle entry points' source files, so that clients may import type definitions from them
    - @FezVrasta I didn't get a answer back from you about whether or not this change was desired, so I went ahead and made the change
    - Let me know if I need to revert it / replace it with something else

**Before**
```js
// @flow
import { createPopper } from '@popperjs/core'
// createPopper is typed any

import type { Instance } from '@popperjs/core'
// cannot find export Instance
```

**After**

```js
// @flow
import { createPopper } from '@popperjs/core'
// createPopper is properly typed thanks to the build change

import type { Instance } from '@popperjs/core'
// Instance typedef is importable thanks to the types export in the bundle entry
```

### DIFF
--- a/.config/rollup.config.js
+++ b/.config/rollup.config.js
@@ -13,7 +13,7 @@ const getFileName = input => input.split('/')[1].split('.')[0];
 const inputs = ['src/popper.js', 'src/popper-lite.js', 'src/popper-base.js'];
 const bundles = [
   { inputs, format: 'umd', dir: 'dist', minify: true, flow: true },
-  { inputs, format: 'umd', dir: 'dist', flow: true },
+  { inputs, format: 'umd', dir: 'dist' },
   { inputs, format: 'cjs', dir: 'dist', flow: true },
 ];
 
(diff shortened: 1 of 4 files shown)


.config/rollup.config.js
```diff
@@ -14,7 +14,7 @@ const inputs = ['src/popper.js', 'src/popper-lite.js', 'src/popper-base.js'];
 const bundles = [
   { inputs, format: 'umd', dir: 'dist', minify: true, flow: true },
   { inputs, format: 'umd', dir: 'dist', flow: true },
```
FezVrasta: Shouldn't we remove this one? `umd` consumers can't take advantage of Flow anyways.

src/popper-base.js
```diff
@@ -1,5 +1,7 @@
 // @flow
 import { createPopper, popperGenerator } from './index';
 
+export * from './types';
```
FezVrasta: Thanks!

PR #1045, opened 2020-03-31
feat: contextElement for virtual elements

atomiks:
We didn't add `referenceNode` from Popper v1 because we thought it wasn't necessary anymore, but in some cases it still is. (Note Popper v1's implementation was bugged, this works correctly.)

Found via https://github.com/atomiks/tippyjs/issues/746#issuecomment-606275367


src/types.js
```diff
@@ -135,4 +135,5 @@ export type Padding = number | $Shape<SideObject>;
 
 export type VirtualElement = {|
   getBoundingClientRect: () => ClientRect | DOMRect,
+  contextNode?: Element,
```
atomiks: `contextElement` instead maybe?

atomiks:
CI tests aren't running anymore(?)

src/types.js
FezVrasta: Yes I agree, since it expects an `Element`

FezVrasta:
The build is here, but for some reason GitHub doesn't link it to the PR automatically.

https://travis-ci.org/github/popperjs/popper-core/builds/669135144

src/types.js
FezVrasta: Should this be documented in the docs?

atomiks: We have to wait until after release. I'll make a docs PR for v2.2

Issue #837, opened 2019-11-05
Firefox bug in getOffsetParent computation with "fixed" position element as firstElementChild


adilyalcin:
Hey all,

There's a very specific bug I encountered with Firefox where the offset element is computed inside popper.js.

Based on [MDN spec page for DOM offsetParent property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent):

> offsetParent returns null ... if the element  has position property set to fixed. **(firefox returns `<body>`)**

I have a fixed element somewhere along the shared DOM as first child, which gets picked for input to the getOffsetParent method. Chrome initializes the offsetParent to null, and continues finding the correct offsetParent eventually. Firefox initializes the offsetParent variable to **< body >** instead. It messes up the position calculations where I attach popper to a different reference element.

The related function is `getOffsetParent`.

```
  var offsetParent = element.offsetParent || null;
  // Temporary fix
  if( isFirefox && getStyleComputedProperty(element,'position')=="fixed"){
    offsetParent = null;
  }
```

I do not currently have the time to create the CodePen sample to replicate the issue, but I may be able to do it a little later. The bug appears in a really specific DOM configuration! And, I'm using tippy library, so it will take extra effort to isolate the bug to using only popper and/or approved libraries here.

I hope the code & documentation case is clear.

Let me know if you have any questions or concerns!

PR #1046, opened 2020-03-31
fix: firefox offsetParent check

### LINKED ISSUE #1043: [Firefox] wrong positioning after switching from relative to fixed positioned parent
### CodePen demo
https://codepen.io/fothermucker/pen/mdJaRLq

1. Open the demo in firefox
2. Scroll the page

The tooltip should stay close to the element, as it happens in google chrome

### What went wrong?

The tooltip element starts to move down when you scroll

### Any other comments?

I have a component that implements a sticky behavior in order to have better support and it works 
just like this demo. You have a container element that wraps all the sticky elements, when the scroll hits the position of the `sticky-container` it puts a class to the `sticky-element` (`.fixed`) and the element stays at the top of the window and the reference element of the popper is inside of this `sticky-element`.

This bug doesn't happen in chrome or safari.

Thx guys :)

atomiks:
Resolves #1043 (and fixed incorrect behavior for #837)

### DIFF
--- a/src/dom-utils/getOffsetParent.js
+++ b/src/dom-utils/getOffsetParent.js
@@ -15,7 +15,7 @@ function getTrueOffsetParent(element: Element): ?Element {
     !isHTMLElement(element) ||
     !(offsetParent = element.offsetParent) ||
     // https://github.com/popperjs/popper-core/issues/837
-    (isFirefox() && getComputedStyle(offsetParent).position === 'fixed')
+    (isFirefox() && getComputedStyle(element).position === 'fixed')
   ) {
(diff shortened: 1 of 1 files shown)


src/dom-utils/getOffsetParent.js
```diff
@@ -15,7 +15,7 @@ function getTrueOffsetParent(element: Element): ?Element {
     !isHTMLElement(element) ||
     !(offsetParent = element.offsetParent) ||
```
FezVrasta: Do we need this assignment?

PR #1049, opened 2020-04-01
fix: virtual element

FezVrasta:


atomiks:
@FezVrasta I realized #1048 is actually the correct behavior, no need for this PR (or the one I closed 😅)

PR #1050, opened 2020-04-01
fix: virtual element

atomiks:
#1049


atomiks:
#1048 is fine actually 🤦‍♂

PR #1053, opened 2020-04-06
Set default value for getComputedStyle transitionProperty.

lpoulter:
In testing environments, JSDOM, getComputedStyle is returning null. This causes [transitionProperty.indexOf(property)](https://github.com/popperjs/popper-core/blob/master/src/modifiers/computeStyles.js#L128) to throw an error.

I've set a default value to stop the error throwing.


atomiks:
I don't think this works, because default properties only resolve for `undefined` (not `null`):

```js
const { transitionProperty = '' } = { transitionProperty: null };

// still null, not ''
```

PR #1058, opened 2020-04-08
docs: add 3rd party modifiers page

atomiks:


FezVrasta:
Thanks! I updated the text and renamed the section to "Community Modifiers", I think it will help to better engage with the community, may you let me know what do you think about the changes?

atomiks:
Yep, worded much better now! Maybe `popper-modifier` instead of `popperjs-modifier` keyword?

FezVrasta:
I wanted to use the same prefix used by the npm scope, just to keep everything consistent on the platform 🤔

atomiks:
uhmm true 😄 I guess I can rename my package to `popperjs-modifier-max-size`, it's very new anyway.

FezVrasta:
The package name doesn’t matter, is enough to have the keyword in the package list

PR #1061, opened 2020-04-09
fix: track scrolling parent when parent of assigned slot

### LINKED ISSUE #1059: Scroll parents not tracked correctly through slotted elements.
### CodePen demo
https://opensource.adobe.com/spectrum-web-components/components/dropdown

1. Visit the above URL
2. Open the dropdown element in the "Example" section
3. Scroll the content
4. Watch the options list not scroll with this page

Scrolling did not affect the location of the options list.

### What went wrong?
Scrolling for this element occurs in a parent of a `<slot>` element. Currently, the `getParentNode` code will not walk the correct branches of a DOM tree that projects content in this way.

### Any other comments?
Scrolling can be corrected by adding `element.assignedSlot ||` as the first optional return in the `getParentNode` helper:
https://github.com/popperjs/popper-core/blob/90cce7abe24d9fa6015beb7abb9aba0662c03203/src/dom-utils/getParentNode.js#L9-L15
I'd be happy to make a PR that includes this if you agree.

Westbrook:
fixes #1059  

- updates `getParentNode` to step into `assignedSlot` "ancestors"
- adds visual test


src/dom-utils/getParentNode.js
```diff
@@ -7,10 +7,9 @@ export default function getParentNode(element: Node | ShadowRoot): Node {
   }
 
   return (
+    element.assignedSlot || // step into the shadow DOM of the parent of a slotted node
     element.parentNode || // DOM Element detected
     // $FlowFixMe: need a better way to handle this...
-    element.host || // ShadowRoot detected
-    document.ownerDocument || // Fallback to ownerDocument if available
-    document.documentElement // Or to documentElement if everything else fails
+    element.host // ShadowRoot detected
```
FezVrasta, line 13:
```suggestion
    element.host || // ShadowRoot detected
    getDocumentElement(element)
```

I think this is what @atomiks suggested

src/dom-utils/getParentNode.js
```diff
@@ -1,16 +1,19 @@
...
 
   return (
+    ((element: any): HTMLElement).assignedSlot || // step into the shadow DOM of the parent of a slotted node
     element.parentNode || // DOM Element detected
     // $FlowFixMe: need a better way to handle this...
     element.host || // ShadowRoot detected
-    document.ownerDocument || // Fallback to ownerDocument if available
-    document.documentElement // Or to documentElement if everything else fails
+    getDocumentElement(element)
   );
```
FezVrasta, line 18:
```js
  return (
    // $FlowFixMe: this is a quicker (but less type safe) way to save quite some bytes from the bundle
    element.assignedSlot ||
    element.parentNode ||
    element.host ||
    getDocumentElement(element)
  );
```

The alternative would be to do the following, but it would require more logic and a new `isShadowRoot` helper, I don't think it's worth it:

```js
export default function getParentNode(
  element: Node | ShadowRoot
): Node | HTMLSlotElement {
  if (getNodeName(element) === 'html') {
    return element;
  }

  if (isShadowRoot(element)) {
    return element.host;
  }

  if (isElement(element)) {
    return (
      element.assignedSlot || element.parentNode || getDocumentElement(element)
    );
  }

  return getDocumentElement(element); // ShadowRoot detected // Or to documentElement if everything else fails
}
```

src/dom-utils/getParentNode.js
```diff
@@ -1,16 +1,19 @@
 // @flow
 import getNodeName from './getNodeName';
+import getDocumentElement from './getDocumentElement';
 
-export default function getParentNode(element: Node | ShadowRoot): Node {
+export default function getParentNode(
+  element: Node | ShadowRoot
+): Node | Element {
```
FezVrasta, line 7:
```suggestion
): Node {
```

`Element` is an interface of `Node`, we don't need to explicitly list it.

src/dom-utils/getParentNode.js
Westbrook: That stricter type requires a `// $FlowFixMe: HTMLElement is a Node` for the fallback...but agree that the simpler version is preferred.

Do Flow types get baked into the actual JS in some way that changes the bundle size? I've only worked with TypeScript and those sorts of inline casts just get flushed, IIUC.

FezVrasta: The increase of weight would be because of the additional `isShadowRoot` helper, and because of the more verbose checks. Flow gets completely stripped away at build time.

Westbrook:
Thanks @FezVrasta!

So I can inform my team when we might be able to leverage this fix, what's generally the release policy here?

FezVrasta:
Hey @Westbrook, here at Popper, Inc. we use an agile approach, where every release is carefully crafted any time someone is persuasive enough to convince me to press the button.

😂 You should find your patch in the 2.2.2 version I just released.

PR #1066, opened 2020-04-12
fix(flow): remove src folder from npm bundle

FezVrasta:
Fix for https://github.com/popperjs/popper-core/issues/1064

### DIFF
--- a/.config/rollup.config.js
+++ b/.config/rollup.config.js
@@ -31,7 +31,7 @@ const configs = bundles
         minify && compiler(),
         minify && terser(),
         license({ banner: `@popperjs/core v${pkg.version} - MIT License` }),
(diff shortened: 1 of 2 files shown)


TrySound:
In this case flow types will stop working
https://unpkg.com/@popperjs/core@2.2.3/dist/cjs/popper.js.flow

Flow only looks for .js.flow near "main" entry point when importing like this
```
import { createPopper } from '@popperjs/core';
```

FezVrasta:
Right! I suppose I could point to the lib folder though

FezVrasta:
I updated Rollup, now the flow file points to lib
```
// @flow

export * from '../../lib/index.js'
```

PR #1067, opened 2020-04-12
fix: properly compute iOS visual viewport

### LINKED ISSUE #1065: Measure viewport properly on iOS
### CodePen demo

https://6z5bu.csb.app/
https://codesandbox.io/s/loving-feather-6z5bu

1. open demo in xcode simulator
2. hardware -> keyboard -> toggle software keyboard
3. click on input
4. scroll page up
5. flip happens only when input is overlapped by keyboard

Flip should happen when without overlapping by software keyboard

### What went wrong?

Mobile browsers decided to not consider software keyboard in window height and instead overlap part of viewport. At some point this prevents junking of measured layout. Though the problem comes when we try to implement custom autocomplete.

### Any other comments?

I tried to fix this problem in our popup with many ways and end up with hardcoded keyboard size took from article. Though I searched for the problem again and found probably new api which may solve the problem.
https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API

This solved maxSize plugin for me and prevented overlapped items.
```
const keyboardHeight = window.innerHeight - window.visualViewport.height
```

Perhaps popper could use this api for viewport measuring and fallback to window when it does not exist. Solution does not require any iOS detection.

FezVrasta:
fixes #1065

### DIFF
--- a/src/dom-utils/getViewportRect.js
+++ b/src/dom-utils/getViewportRect.js
@@ -3,10 +3,11 @@ import getWindow from './getWindow';
 
 export default function getViewportRect(element: Element) {
   const win = getWindow(element);
+  const visualViewport = win.visualViewport || {};
 
   return {
-    width: win.innerWidth,
-    height: win.innerHeight,
+    width: visualViewport.width || win.innerWidth,
+    height: visualViewport.height || win.innerHeight,
     x: 0,
     y: 0,
   };
--- a/src/types.js
+++ b/src/types.js
@@ -25,6 +25,10 @@ export type Window = {|
   Node: Node,
   toString(): '[object Window]',
   devicePixelRatio: number,
+  visualViewport?: {
+    width: number,
+    height: number,
+  },
 |};
 
 export type Rect = {|


src/dom-utils/getViewportRect.js
```diff
@@ -3,10 +3,11 @@ import getWindow from './getWindow';
 
 export default function getViewportRect(element: Element) {
   const win = getWindow(element);
+  const visualViewport = win.visualViewport;
 
   return {
-    width: win.innerWidth,
-    height: win.innerHeight,
+    width: visualViewport.width || win.innerWidth,
+    height: visualViewport.height || win.innerHeight,
```
atomiks, line 10: This will error if `visualViewport` is undefined

src/dom-utils/getViewportRect.js
```diff
@@ -3,10 +3,11 @@ import getWindow from './getWindow';
 
 export default function getViewportRect(element: Element) {
   const win = getWindow(element);
+  const visualViewport = win.visualViewport || {};
 
   return {
-    width: win.innerWidth,
-    height: win.innerHeight,
+    width: visualViewport.width || win.innerWidth,
```
atomiks, line 9: Is the optional chaining operator supported in flow? `win.visualViewport?.width`

FezVrasta: It is, but we don’t have the right Babel plugin I think. And I think it would produce a larger output

Issue #1062, opened 2020-04-09
Allow auto placements to be specified


atomiks:
## Feature description

`auto` chooses which placement to use (with most space), but you can't restrict which placements to use.

In this example, we only want to use `top` and `bottom`, but we still get the intrinsic "auto" behavior which is not achievable with the default `fallbackPlacements` option.

```js
modifiers: [
  {
    name: 'flip',
    options: {
      autoPlacements: ['top', 'bottom']
    }
  }
]
```

## Why should this feature be part of the Popper's core?

@FezVrasta wanted to use it 😄

FezVrasta:
https://popper.js.org/docs/v2/modifiers/flip/#allowedautoplacements

PR #1070, opened 2020-04-12
feat: add `autoWhitelist` option to `flip` modifier

FezVrasta:
This should implement #1062


atomiks:
Not a fan of the term "whitelist" and it's not descriptive of the type like "placements"

FezVrasta:
Alternatives?

atomiks:
`autoPlacements`?

FezVrasta:
I don’t think it’s very clear, this list is a list of placements that can be resolved to starting from the auto placement.

atomiks:
To me, `autoPlacements` is saying, "use only (or restrict to) these placements for `auto`", which is what the behavior is right?

FezVrasta:
I went with `allowedAutoPlacements`

PR #1072, opened 2020-04-13
fix(types): options types

atomiks:
- ~Need to test with TS as well~ (TS works, although the red squiggles in VSCode don't place themselves in useful locations...)

You can now do this for typechecking:

```js
// @flow
import type {Modifier, StrictModifiers} from '@popperjs/core';

type CustomModifier = $Shape<Modifier<any>> & {
  name: 'custom',
  options?: $Shape<{customOption: boolean}>,
};

type ExtendedModifiers = StrictModifiers | CustomModifier;

createPopper<ExtendedModifiers>(ref, pop, {
  modifiers: [
    {name: 'offset', options: {offset: [0, '']}}, // type error
    {name: 'custom', options: {customOption: true}}, // no error
  ]
});
```

VSCode screenshot

### DIFF
--- a/src/modifiers/arrow.js
+++ b/src/modifiers/arrow.js
@@ -11,7 +11,12 @@ import expandToHashMap from '../utils/expandToHashMap';
 import { left, right, basePlacements, top, bottom } from '../enums';
 
 // eslint-disable-next-line
-export type Options = {
+export type ArrowModifier = $Shape<Modifier<any>> & {
+  name: 'arrow',
+  options?: $Shape<Options>,
+};
+
+type Options = {
   element: HTMLElement | string,
   padding: Padding,
 };
--- a/src/modifiers/arrow.js
+++ b/src/modifiers/arrow.js
@@ -10,7 +10,8 @@ import mergePaddingObject from '../utils/mergePaddingObject';
 import expandToHashMap from '../utils/expandToHashMap';
 import { left, right, basePlacements, top, bottom } from '../enums';
 
-type Options = {
+// eslint-disable-next-line
+export type Options = {
   element: HTMLElement | string,
   padding: Padding,
 };
--- a/src/types.js
+++ b/src/types.js
@@ -2,12 +2,15 @@
 /* eslint-disable import/no-unused-modules */
 import type { Placement, ModifierPhases } from './enums';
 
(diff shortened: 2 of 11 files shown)


FezVrasta:
Shouldn’t it possible to to `Modifier<{ custom stuff }>` already?

atomiks:
It's typed as `Obj`:

https://github.com/popperjs/popper-core/blob/8a81319d1e7a50d21b92e34caf2328797c97d025/src/types.js#L108-L117

I pushed a fix for that, but the other changes still seem required.

FezVrasta:
I mean, this should work

```
type CustomModifier = Modifier<{customOption: boolean}> & {
  name: 'custom',
};
```

atomiks:
I tried that but it broke the implementation (see edits in prev comment, lol)

src/types.js
```diff
@@ -110,6 +117,54 @@ export type Modifier<Options> = {|
...
+export type StrictModifiers =
+  | OffsetModifier
+  | ApplyStylesModifier
+  | ArrowModifier
+  | ComputeStylesModifier
+  | EventListenersModifier
+  | FlipModifier
+  | PreventOverflowModifier
+  | PopperOffsetsModifier;
+
```
FezVrasta, line 167: I was trying to avoid to have a centralized place where the modifiers types are defined, ideally, each modifier file should export its type.

src/modifiers/arrow.js
```diff
@@ -10,6 +10,12 @@ import mergePaddingObject from '../utils/mergePaddingObject';
 import expandToHashMap from '../utils/expandToHashMap';
 import { left, right, basePlacements, top, bottom } from '../enums';
 
+// eslint-disable-next-line
+export type ArrowModifier = $Shape<Modifier<any>> & {
+  name: 'arrow',
+  options?: $Shape<Options>,
+};
+
```
FezVrasta, line 18: What I'd like to see is the default export work as a replacement for this type. We already export a `Modifier<Options>`, if we could fix the type, and add a `name` argument, we could simplify a lot the types.

```js
Modifier<'arrow', Options>
```

atomiks: Uhm I think I'll leave it up to you from now 😇

PR #1075, opened 2020-04-13
fix: ios visual viewport

### LINKED ISSUE #1074: visualViewport offset make hide and flip modifiers crazy
### CodePen demo

https://6z5bu.csb.app/
https://codesandbox.io/s/loving-feather-6z5bu

1. open xcode ios simulator
2. toggle software keyboard
2. focus input
3. scroll up and down, hide attribute is applied when input is in viewport

Hide attribute should be applied only when input is out of visible viewport

### What went wrong?

visualViewport offsets breaks hide and flip modifiers

atomiks:
fix #1074

Maybe there is a better hack 🤦

### DIFF
--- a/src/dom-utils/getViewportRect.js
+++ b/src/dom-utils/getViewportRect.js
@@ -3,12 +3,23 @@ import getWindow from './getWindow';
 
 export default function getViewportRect(element: Element) {
   const win = getWindow(element);
-  const visualViewport = win.visualViewport || {};
+  const visualViewport = win.visualViewport;
 
-  return {
-    width: visualViewport.width || win.innerWidth,
-    height: visualViewport.height || win.innerHeight,
-    x: visualViewport.offsetLeft || 0,
-    y: visualViewport.offsetTop || 0,
-  };
+  let width = win.innerWidth;
+  let height = win.innerHeight;
+  let x = 0;
+  let y = 0;
+
+  if (visualViewport) {
+    width = visualViewport.width;
+    height = visualViewport.height;
+
+    // iOS has a buggy implementation as expected
+    if (!/iPhone|iPod|iPad/.test(navigator.platform)) {
+      x = visualViewport.offsetLeft;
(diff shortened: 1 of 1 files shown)


TrySound:
Can we assume that with scale === 1 offsets are always 0?

TrySound:
This works well for me on iOS and scaled chrome on mac
```
import getWindow from './getWindow.js';
export default function getViewportRect(element) {
  var win = getWindow(element);
  var visualViewport = win.visualViewport;
  if (visualViewport) {
    return {
      width: visualViewport.width,
      height: visualViewport.height,
      x: visualViewport.scale === 1 ? 0 : visualViewport.offsetLeft,
      y: visualViewport.scale === 1 ? 0 : visualViewport.offsetTop,
    };
  } else {
    return {
      width: win.innerWidth,
      height: win.innerHeight,
      x: 0,
      y: 0,
    };
  }
}
```

atomiks:
It's broken on macOS Safari for me.

TrySound:
I see only horizontal offset problem. Same problem in chrome. Maybe positioning need to be tweaked?

atomiks:
https://bugs.webkit.org/show_bug.cgi?id=170981 It's this I think. Safari and Chrome do the opposite thing

FezVrasta:
I think we should merge this PR as is, we can work on a subsequent PR to improve the situation for other browsers. Do everybody agree?

TrySound:
width/height work pretty well. The problem is only zoomed viewport. It does not work without visualViewport aswell. So browser detection is not necessary.

atomiks:
I personally use pinch zoom a lot in Chrome and it's an accessibility problem if the popper shifts incorrectly because of this. Applying to iOS only is the safest solution

TrySound:
ok

TrySound:
Can this be release or there are other issues?

atomiks:
It should be fine, this is a fairly critical issue so hopefully a release today

PR #1079, opened 2020-04-14
Atomiks fix/modifiers data types

FezVrasta:
Same as https://github.com/popperjs/popper-core/pull/1071, but with TS fixes

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -90,7 +90,39 @@ export type State = {|
   attributes: {|
     [key: string]: { [key: string]: string | boolean },
   |},
-  modifiersData: { [key: string]: any },
+  modifiersData: {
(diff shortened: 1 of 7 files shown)


src/types.js
```diff
@@ -69,6 +69,9 @@ export type StateOffsets = {|
   arrow?: Offsets,
 |};
 
+/*:: type OffsetData = { [Placement]: Offsets }; */
+/*;; type OffsetData = { [key in Placement]?: Offsets } */
```
FezVrasta, line 73: @atomiks from now on we can use this syntax to define TS-specific types. They will be ignored by Flow, and TS ignores the ones within `/*::` comment blocks.

PR #1080, opened 2020-04-16
Type enhancements

lsroman:
Hello. Thanks for the great library.

I am switching to v2 and I have some problems with types from popperJS.

1. My eslint gives an export error and does not find types.
Screenshot 2020-04-16 at 12 53 08
2. I wrapping popper and would like to get options types for default modifiers
Screenshot 2020-04-16 at 12 54 34

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "main:umd": "dist/umd/popper.js",
   "module": "lib/popper.js",
   "unpkg": "dist/umd/popper.min.js",
+  "types": "index.d.ts",
   "author": "Federico Zivolo <federico.zivolo@gmail.com>",
   "license": "MIT",
   "repository": "github:popperjs/popper-core",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "main:umd": "dist/umd/popper.js",
   "module": "lib/popper.js",
   "unpkg": "dist/umd/popper.min.js",
-  "types": "index.d.ts",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -6,6 +6,7 @@
   "main:umd": "dist/umd/popper.js",
   "module": "lib/popper.js",
   "unpkg": "dist/umd/popper.min.js",
+  "types": "index.d.ts",
```
FezVrasta, line 9: We already have a root `index.d.ts`, I don't want to add an ambiguous property to the package.json, since it doesn't make it clear if it's referring to Flow or TypeScript.

lsroman: Hmm... I did not think about Flow(

FezVrasta: Anyways, the imports seem to work correctly here:
https://codesandbox.io/s/angry-fog-pcjfp?file=/src/index.ts

Probably your configuration has something off, TS has like... 1 bilion different options, incompatible with each other.

lsroman: Problem is not in TS, it works. Types cannot find eslint. Now I will make an example.

Field "types"/"typings" assigned to typescript. https://github.com/facebook/flow/issues/5994

FezVrasta: If ESlint doesn't use the root `index.d.ts` that's a bug with the ESLint plugin, they should fix it.

lsroman: Okey, I will create a issue for them

PR #1081, opened 2020-04-16
setOptions uses forceUpdate so initialization is sync

domarmstrong:
This PR changes the initialisation to run sync via forceUpdate rather than async with update. The initial async update causes problems in test suites as the promise is not returned so you have to wait somehow, although there doesn't seem to be a need for this behaviour as far as I can tell.

This is a change in behaviour as setOptions update is not debounced although this seems fine to me as you should not be calling setOptions repeatedly as you might with update. The return type of setOptions is still a promise, but this should be removed in the future.
This also changes the behaviour, even remove the need of onFirstUpdate.

I've added a cancel method to the debounced update function, currently if you had called update and then forceUpdate it would run forceUpdate twice, which isn't needed. It may also be possible to get rid of the isDestroyed flag and call `instance.update.cancel` in destroy instead.

### DIFF
--- a/src/__snapshots__/index.test.js.snap
+++ b/src/__snapshots__/index.test.js.snap
@@ -19,6 +19,21 @@ Object {
     },
     "orderedModifiers": Array [],
     "placement": "bottom",
+    "rects": Object {
+      "popper": Object {
+        "height": 0,
+        "width": 0,
+        "x": 0,
+        "y": 0,
+      },
+      "reference": Object {
+        "height": 0,
+        "width": 0,
+        "x": 0,
+        "y": 0,
+      },
+    },
+    "reset": false,
     "scrollParents": Object {
       "popper": Array [
         [Window],
--- a/src/index.js
+++ b/src/index.js
@@ -155,7 +155,10 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
 
         runModifierEffects();
 
-        return instance.update();
+        instance.forceUpdate();
+        // @deprecated, the promise return type is for api compatibility and
+        // will be removed in the future
+        return Promise.resolve(state);
       },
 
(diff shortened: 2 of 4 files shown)


atomiks:
I used this in v1 and v2:

Jest `setup.js`
```js
jest.useFakeTimers();

window.promise = require('promise'); // uses macrotask timers (`setTimeout`)
```

domarmstrong:
The point is, I don't think people should need to work around it, there doesn't seem to be any benefit to having this async debouncing on setOptions

atomiks:
I am not sure exactly, it's very micro-optimization to let the browser paint maybe (like `useEffect` vs `useLayoutEffect`)? But I think @FezVrasta knows more about it. I originally wanted to make it sync myself:

https://github.com/popperjs/popper-core/pull/952

domarmstrong:
The problem is its a leaky effect, if you are testing higher level components you need to start waiting for the implementation detail of a lower level component. That is a bit ugly where you then have comments explaining why you are doing this work around in several places. If we are talking micro optimisation, is it worth it for the complexity? Maybe it could be an opt in or opt out feature? In the example of react-popper it isn't doing anything with the returned promise just ignoring it, which is the same issue so it would probably prefer to opt out https://github.com/popperjs/react-popper/blob/master/src/usePopper.js#L127. I would suggest making it off by default and an opt in feature.

FezVrasta:
It's a micro optimization, but every little bit matters for libraries like this, where consumers need to call the library methods during scroll events and such.

Test ergonomics have a lower precedence to me, so I'll close this for now. Also because I don't want to break the API even if just subtly.

Thanks for the PR and I'm sorry if you wasted time, it's better do open an issue to discuss this kind of changes before, to avoid this kind of situations.

PR #1082, opened 2020-04-16
feat: add margin warning suppression

jquense:
Not sure this is the best idea, but we are trying cobble together bootstrap v4 with popper without messing with the styles. Accomplishing it with a custom modifier that does check and account for margins when doing element rects https://github.com/react-bootstrap/react-bootstrap/pull/5112

Maybe it's better that we clear the margins manually and do offsets, but I can't think of a way to do that and also respect user overrides of margin.

This seems like it'd be safe to turn off the warning if i know i've handled it?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -127,29 +127,31 @@ export function popperGenerator(generatorOptions: PopperGeneratorArgs = {}) {
             }
           }
 
-          const {
-            marginTop,
-            marginRight,
-            marginBottom,
-            marginLeft,
-          } = getComputedStyle(popper);
+          if (!state.options.suppressMarginWarning) {
+            const {
+              marginTop,
+              marginRight,
+              marginBottom,
+              marginLeft,
+            } = getComputedStyle(popper);
 
-          // We no longer take into account `margins` on the popper, and it can
-          // cause bugs with positioning, so we'll warn the consumer
-          if (
-            [marginTop, marginRight, marginBottom, marginLeft].some(margin =>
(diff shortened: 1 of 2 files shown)


jquense:
Another approach could be to move modifier effects before this warning so we could check the margins, store the offsets and then remove the margin on the element

atomiks:
> Another approach could be to move modifier effects before this warning so we could check the margins, store the offsets and then remove the margin on the element


That's a lot better, adding an option to disable a warning feels really 🤢

jquense:
The counter point to that suggestion tho is that, Popper can't know if I have added margin support so the option is helpful for disabling it when it's wrong. Plus there is no runtime behavior change like moving the effect code might have.

I'm a bit torn on which is the best (for our use case) I'd like to not mutate element styles in the effect if possible, but it's also more correct of a solution. I'm not sure if it will break people tho

FezVrasta:
Would it help if we trigger the warning only if the margins are not equal all around?

jquense:
good question...let me check

jquense:
hmm i don't think it'd help, margin is only on the side relevant to the placement.

jquense:
I have figured out an approach around this warning so maybe nothing is needed after all https://github.com/react-bootstrap/react-bootstrap/pull/5112

FezVrasta:
Great! Thanks for the help

jquense:
thanks for the excellent library ya'll

PR #1084, opened 2020-04-18
fix: documentElement as scrollParent

### LINKED ISSUE #1078: Wrong Popper position in a scrolled content
### CodePen demo

https://codepen.io/alex3ax/pen/qBOONPW

1. Scroll a little bit down
2. Click on the button
3. The tooltip position is wrong

The tooltip should be stuck to the bottom of the button

### What went wrong?

The scrollTop of the ref is always 0 in this example, it looks like the offset calculation can't take scroll into account. 

### Any other comments?

Given such a layout is it possible to configure the Popper behavior correctly without layout changes?

atomiks:
Closes #1078 

🚨We need to make sure this doesn't break Edge (<=2019) because the conditional where I added the new expression was important to it IIRC, we had to fix something


atomiks:
The tests pass for me locally 😢

atomiks:
Uhmm so the module path being misspelled literally didn't error for me anywhere locally (IDE, `yarn test`, ...) but failed in CI? 🤦‍♂️

FezVrasta:
> Uhmm so the module path being misspelled literally didn't error for me anywhere locally (IDE, `yarn test`, ...) but failed in CI? 🤦‍♂️ 

macOS is not case-sensitive, Linux is

FezVrasta:
Edge 18
image

atomiks:
Okay so...

Edge / IE don't report the actual `.scrollTop` on the body, and in this case, it's the only value we can use to determine scroll position (the window and html aren't being scrolled). This DOM layout is so ridiculous anyway 🤦

FezVrasta:
So, it looks like there is no way to detect `document.body` scroll offsets on IE and Edge <= 18 when the `body` is `position: relative; overflow: auto;`.

I'm going to merge this PR to at least fix the other browsers.

PR #1095, opened 2020-04-27
feat(flip): mainAxis and altAxis options

### LINKED ISSUE #1011: Add `mainAxis` and `altAxis` to the flip modifier
## Feature description

The flip modifier checks both axes to determine when to flip in v2. In Popper v1, it only checked the `mainAxis` (the axis along the placement, so `top` = `y`, the opposite of `preventOverflow`). 

## Why should this feature be part of the Popper's core?

There may be cases where Popper 1 behavior is preferred? It also aligns with `preventOverflow` and won't cost many bytes to support (I think). Both will be `true` by default still, just with the ability to configure the behavior.

### LINKED ISSUE #1094: Setting order/hierarchy of modifiers
## Feature description

Provide a way to set a hierarchy or order for modifiers. As far as I can tell looking through the docs, this is not possible and the order of the `modifiers` array is not respected.

This may only pertain to some modifiers as some are independent of others or don't perform conditional changes such as `offset`.

This could be added by respecting the array order or adding `order` prop on `TModifier` such as below.

```ts
createPopper(anchorRef, tooltipRef, {
  modifiers: [
    {
      name: 'preventOverflow',
      order: 1,
    },
    {
      name: 'flip',
      order: 2,
    },
  ],
});
```

It seems this logic would have to go in the core but I'm not sure if this could be implemented in the modifiers themselves.

### Backstory

I am using `popperjs` to display a tooltip for a chart at a given cursor position which I am calculating, which is the blue element in the screenshots (aka the `anchorRef`).

It is working amazingly! However, I came across a case where I would like the `preventOverflow` modifier to take precedence over the `flip` modifier.

These are the options I am passing to `createPopper`
```ts
createPopper(anchorRef, tooltipRef, {
  strategy: 'fixed',
  placement: 'right',
  modifiers: [
    {
      name: 'offset',
      options: {
        offset: [0, 10],
      },
    },
    {
      name: 'preventOverflow',
      options: {
        tether: false,
        boundary: chartContainerRef,
      },
    },
    {
      name: 'flip',
      options: {
        fallbackPlacements: ['right', 'left', 'top', 'bottom'],
        boundary: chartContainerRef,
      },
    },
  ],
});
```

The issue is that I want `bottom` and `top` to be `fallbackPlacements`, which causes `flip` to set the placement to `bottom` even though `preventOverflow` would have positioned the tooltip inside the boundary. So this is what I actually get with the setup above.

Screen Recording 2020-04-27 at 10 45 AM

What I would like to see is this. Where at the top `preventOverflow` pushes the tooltip down to stay inside the boundary. And on the far right, `flip` changes placement to `left` and `preventOverflow` still pushes down at the top. So `preventOverflow` takes priority over `flip`, but in cases where a flip is required, it still performs the flip, such as on the far right.

Screen Recording 2020-04-27 at 10 40 AM

> If there is a way to achieve the above without this feature, I am open to suggestions. Either way, I think this could be a good enhancement.

## Why should this feature be part of the Popper's core?

- Allows greater control and flexibility in using `popperjs` and modifiers
- This feature seems to be a relatively small code change, afaik

atomiks:
Closes #1094
Closes #1011


FezVrasta:
Is the snapshot ok?

tests/functional/__image_snapshots__/flip-test-js-should-not-flip-to-the-top-1-snap.png

atomiks:
Yes, without this change it would not be flush at the right of the container (since it's using the `right` placement + preventOverflowed, not `top`)

PR #1096, opened 2020-04-28
docs: use @popperjs/core downloads

atomiks:
Since it's now at ~1M from 0 when we released it lol


FezVrasta:
There's still quite a difference in numbers, I'd like to keep the other one for a bit more.

atomiks:
At the same time, it's useful to see how it's growing at a glance and I don't think 1M+ downloads/month will turn anyone off from using the lib compared to 20M? 🤔

FezVrasta:
Let's wait a bit more, we can always manually check for the progress of the new package.

PR #1098, opened 2020-04-29
fix: rewrite clippingRect logic

### LINKED ISSUE #1097: Wrong tooltip positions when X axis has overflow (on mobile)
### CodeSandbox demo

gistcdn.githack.com/FezVrasta/11b807834b33667c5d6d5422f0f6258c/raw/34fdc52bf041388bb3f57e4c94690ddfcf684a14/bug.html

(you need to open the dev tools and go in the mobile emulation mode)

1. 
2.
3.

atomiks:
Attempt to close #1097


src/dom-utils/getViewportRect.js
```diff
@@ -1,20 +1,30 @@
...
   // We don't know which browsers have buggy or odd implementations of this, so
   // for now we're only applying it to iOS to fix the keyboard issue.
   // Investigation required
-  if (visualViewport && /iPhone|iPod|iPad/.test(navigator.platform)) {
+  if (visualViewport) {
     width = visualViewport.width;
     height = visualViewport.height;
+
+    // Not Safari (which uses visual viewport by default)
+    if (!/^((?!chrome|android).)*safari/i.test(navigator.userAgent)) {
```
atomiks, line 23: This is terrible; need to figure out a way to feature detect Safari's differences

src/dom-utils/getViewportRect.js
```diff
@@ -13,6 +13,11 @@ export default function getViewportRect(element: Element) {
   let x = 0;
   let y = 0;
 
+  // NB: This isn't supported on iOS <= 12. If the keyboard is open, the popper
+  // can be obscured underneath it.
+  // Also, `html.clientHeight` adds the bottom bar height in Safari iOS, even
+  // if it isn't open, so if this isn't available, the popper will be detected
+  // to overflow the bottom of the screen too early.
```
atomiks, line 20: We could add the `HACK` for iOS <=12's bottom bar like before, but we still have that keyboard issue. iOS 13+ works fine without any hack.

PR #1099, opened 2020-05-01
fix: root rects

atomiks:
Problems I encountered:

1. `documentRect` width was inaccurate (area outside of `<html>` / `<body>` bounds if it was horizontally scrollable) and possibly the height. I'm not sure what effects this has exactly since it seemed to work before with `rootBoundary: "document"`
2. `viewportRect` in mobile emulation mode when pinch zooming broke the horizontal demo in #1097 - it looks like the feature detection doesn't work with when the browser has zoomed the viewport out. I cannot think of a way to fix it besides userAgent detection which can easily break in the future if Safari decides to start using "layout viewport"...

### DIFF
--- a/src/dom-utils/getViewportRect.js
+++ b/src/dom-utils/getViewportRect.js
@@ -25,11 +25,13 @@ export default function getViewportRect(element: Element) {
     // Uses Layout Viewport (like Chrome; Safari does not currently)
     // In Chrome, it returns a value very close to 0 (+/-) but contains rounding
     // errors due to floating point numbers, so we need to check precision.
-    // Safari returns a number <= 0, usually <= 1 when pinch-zoomed
-    if (
-      Math.abs(win.innerWidth / visualViewport.scale - visualViewport.width) <
-      0.001
-    ) {
+    // Safari returns a number <= 0, usually < -1 when pinch-zoomed
+
+    // Feature detection fails in mobile emulation mode in Chrome.
+    // Math.abs(win.innerWidth / visualViewport.scale - visualViewport.width) <
+    // 0.001
+    // Fallback here: "Not Safari" userAgent
+    if (!/^((?!chrome|android).)*safari/i.test(navigator.userAgent)) {
       x = visualViewport.offsetLeft;
       y = visualViewport.offsetTop;
     }
--- a/src/dom-utils/getDocumentRect.js
+++ b/src/dom-utils/getDocumentRect.js
@@ -12,8 +12,18 @@ export default function getDocumentRect(element: HTMLElement): Rect {
   const winScroll = getWindowScroll(element);
   const body = element.ownerDocument.body;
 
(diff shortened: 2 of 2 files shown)


src/dom-utils/getViewportRect.js
```diff
@@ -25,11 +25,13 @@ export default function getViewportRect(element: Element) {
...
     // In Chrome, it returns a value very close to 0 (+/-) but contains rounding
     // errors due to floating point numbers, so we need to check precision.
-    // Safari returns a number <= 0, usually <= 1 when pinch-zoomed
-    if (
-      Math.abs(win.innerWidth / visualViewport.scale - visualViewport.width) <
-      0.001
-    ) {
+    // Safari returns a number <= 0, usually < -1 when pinch-zoomed
+
+    // Feature detection fails in mobile emulation mode in Chrome.
```
FezVrasta, line 30: Does it happen only in the mobile emulation or also on the real devices?

atomiks: It happens when the browser has zoomed out to view overflowing content, which is what mobile devices do. So it would happen on them too.

src/dom-utils/getViewportRect.js
```diff
@@ -25,11 +25,13 @@ export default function getViewportRect(element: Element) {
...
+    // Math.abs(win.innerWidth / visualViewport.scale - visualViewport.width) <
+    // 0.001
+    // Fallback here: "Not Safari" userAgent
+    if (!/^((?!chrome|android).)*safari/i.test(navigator.userAgent)) {
```
FezVrasta, line 34: If we end up using this check we may want to move it outside the function so that we don't run it more than once per session

atomiks: I think it's too impractical to use since it could change in the future, have you had a look at this? Maybe you can find a way to feature detect this that works for all edge cases? There might be something fundamentally wrong with the approach. Not being able to write a test for it doesn't exactly help either.

FezVrasta: Puppetter should allow us to enable the mobile emulation mode, if we want to write a test for this we can. I'll try to add one now.

atomiks: To summarize the issue:

Safari's `getBoundingClientRect()` returns "visual viewport" values by default (including pinch-zoom), but Chrome's returns "layout viewport" values (not including pinch-zoom). This means it already uses visual viewport by default, so the `offsetLeft/Top` values to determine the offset relative to `getBoundingClientRect()` always need to be `0`, but in Chrome, we need to consider them.

So we need to be able to feature detect if `getBoundingClientRect()` is returning visual or layout viewport values, and if it's visual, don't add the offsets.

FezVrasta: I just pushed a commit to add the test for this bug, I haven't added the pinch to zoom test yet since I haven't found how to perform the action 😅

PR #1100, opened 2020-05-04
fix return type for effect function

joshiste:
Without the clarifying parentheses it the return type of
the effect function is "function returning either void or void" 
instead of the intended "void or function returning void"


FezVrasta:
thanks!

PR #1113, opened 2020-05-23
Fixing functional test OOM issue

ianwalter:
By using a top-level container for the job you can pass a Docker option
to increase the shared memory used by the container. I set it to 1
gigabyte since that was suggested but it can be increased if that is
not sufficient for whatever reason.


FezVrasta:
Thanks!!

PR #1119, opened 2020-05-31
add some documentation on getOffsetParent and listScrollParents

mreinstein:
Adding a few comments based on questions and discussion with @FezVrasta


FezVrasta:
Thanks!

Issue #1035, opened 2020-03-23
strategy: fixed is broken if its parent has will-change


layershifter:
### CodeSandbox

https://codesandbox.io/s/misty-resonance-n18pt

### Steps to reproduce the problem

1. Open CodeSandbox
2. A tooltip is positioned inside container

![image](https://user-images.githubusercontent.com/14183168/77322848-a5ffce00-6d14-11ea-805a-1d3dd525e88f.png)

### What is the expected behavior?

The same positioning as without `will-change`.

### What went wrong?

Disabling of `gpuAcceleration` will not work 😭 
Any ideas why it happens? How it can be fixed?

FezVrasta:
Thanks for the report! Most likely the logic we use to detect offset parents is not taking in account the new render context generated by the `will-change: transform` property.

https://github.com/popperjs/popper-core/blob/master/src/dom-utils/getOffsetParent.js

atomiks:
Maybe this code as well, since it only affects the`fixed` strategy (and scrolling is not correct)? 

https://github.com/popperjs/popper-core/blob/be548db65912e1126581a44faa7f758ff004680a/src/dom-utils/getCompositeRect.js#L23-L35

atomiks:
It seems like the `offsetParent` is always `null` for `fixed` elements, but in this case, the `will-change` property (or `perspective` etc. properties as well) makes it an `offsetParent`? Is it a browser bug?

@FezVrasta it reminds me of these tweets you made: https://twitter.com/FezVrasta/status/1197515785289252864

layershifter:
As I understand this is related: https://bugs.chromium.org/p/chromium/issues/detail?id=20574&desc=2

This is very optimistic:
![image](https://user-images.githubusercontent.com/14183168/77404623-4ce67780-6db2-11ea-94ac-b7f2b4efc90d.png)

It looks that `transform` also affects it.


### without
![image](https://user-images.githubusercontent.com/14183168/77404757-7f907000-6db2-11ea-87ee-d754184cf912.png)

### `transform: translateZ(0);`

![image](https://user-images.githubusercontent.com/14183168/77404813-8fa84f80-6db2-11ea-95b1-5d04387c41eb.png)

atomiks:
Since it's a browser bug, I guess the only solution is to move the popper outside of the container element.

layershifter:
[...]

PR #1120, opened 2020-06-04
Get nearest transformed parent

ptu14:
I've been looking at #1035  for a long time and I wonder if in case the absence of an offset parent, we can't search for a parent with transform or will-change. 

This solution fix problem in my project and examples shown in codesanbox in #1035. (Maybe with little hack when I need add 'will-change' to #inner in this example https://codesandbox.io/s/misty-resonance-n18pt?file=/index.html:150-155 , but it's still better solution than broken popper element)

### DIFF
--- a/src/dom-utils/getCompositeRect.js
+++ b/src/dom-utils/getCompositeRect.js
@@ -29,14 +29,14 @@ export default function getCompositeRect(
     ) {
       scroll = getNodeScroll(offsetParent);
     }
+  }
 
-    if (isHTMLElement(offsetParent)) {
-      offsets = getBoundingClientRect(offsetParent);
-      offsets.x += offsetParent.clientLeft;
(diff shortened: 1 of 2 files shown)


FezVrasta:
Thanks for the PR!

The problem I see with this approach is that we could have an offset parent, but not being the right now. For example this could happen if you have a `position: relative` container, with inside a transformed container, with inside the popper.

ptu14:
I'm not quite sure if I know what you mean. Could you prepare a quick example (codesandbox)? I would like to see how it behaves in the current popper2 and how in my approach.

I need to fix this bug couse i migrate my project from popper1 to popper2 and there is regresion.

FezVrasta:
https://codesandbox.io/s/youthful-yalow-nh32p

ptu14:
Thanks, now i get it. I'll try to find a solution

PR #1123, opened 2020-06-06
fix: detect containing block

### LINKED ISSUE #1115: Strategy "fixed" not working fine for popper@2
According to: #1035 I want to continue this topic, don't know why you said that my tone is over the top, but I'm sorry if I offended you.

But to continue I sent you examples, potencial function which made it working better in popper@1 and polite asked you to reopen issue, due to bug in popper@2

atomiks:
Closes #1115


ptu14:
This solution helps with my problem which I tried to solve in #1120 . I tested  this solution in various scenarios and it works very well.

PR #1129, opened 2020-06-09
fix(getContainingBlock): check html in while loop

### LINKED ISSUE #1128: Infinite loop if popper is an unmounted element
### CodeSandbox demo

https://codesandbox.io/s/adoring-meadow-jq7i8

1. Create an element that's not mounted
2. Set it as popper element

Does nothing (before v2.4.1), or maybe a warning can be given.

### What went wrong?

Stuck in infinite loop.

### Any other comments?

I have checked and I think the code causing the problem is in [these lines](https://github.com/popperjs/popper-core/blob/master/src/dom-utils/getOffsetParent.js#L26). When an unmounted element is given to `getParentNode(el)` it returns the root html. Then when html element is given to `isHTMLElement(el)` it will be `false` and cause infinite loop.

I used react-popper, and I think it's not uncommon to have popper element that suddenly gets unmounted.

Doesn't have the problem in v2.4.0.

atomiks:
@AlbertLucianto @rijkvanzanten does this fix the issue?

Closes #1128

### DIFF
--- a/src/dom-utils/getOffsetParent.js
+++ b/src/dom-utils/getOffsetParent.js
@@ -23,7 +23,7 @@ function getTrueOffsetParent(element: Element): ?Element {
 function getContainingBlock(element: Element) {
   let currentNode = getParentNode(element);
 
-  while (getNodeName(currentNode) !== 'body') {
+  while (['html', 'body'].indexOf(getNodeName(currentNode)) < 0) {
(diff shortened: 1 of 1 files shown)


AlbertLucianto:
@atomiks It fixed the issue on my side. Thanks!

PR #1140, opened 2020-06-27
fix: detectOverflow export

### LINKED ISSUE #1130: detectOverflow util not exported on '@popperjs/core'
Screen Shot 2020-06-15 at 3 42 56 PM

1. `yarn add @popper/core`
2. In component file, write `import { detectOverflow } from '@popperjs/core';`
3. Get module not found error.

As describe in [the docs](https://popper.js.org/docs/v2/utils/detect-overflow/), `detectOverflow` should be able to be imported as a util from `@popperjs/core`.
### What went wrong?

I received an error saying there was no exported member `detectOverflow`.

### Any other comments?

I'm using ` "@popperjs/core": "^2.4.2",` I do not see `detectOverflow` as one of the exported members in `./lib`.

atomiks:
@lainermeister does this fix the issue?

Closes #1130

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -17,6 +17,7 @@ import validateModifiers from './utils/validateModifiers';
 import uniqueBy from './utils/uniqueBy';
 import getBasePlacement from './utils/getBasePlacement';
 import mergeByName from './utils/mergeByName';
+import detectOverflow from './utils/detectOverflow';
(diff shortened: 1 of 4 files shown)


lainermeister:
> @lainermeister does this fix the issue?

I'm not able to test as it doesn't seem like 2.4.3 is released yet!

PR #1142, opened 2020-06-27
fix: remove body as a clippingParent

### LINKED ISSUE #1127: Hide is always true beneath the initial height of the screen.
### CodeSandbox demo

https://codesandbox.io/s/naughty-chandrasekhar-tntlb?file=/index.html

hidden-popper

The poppers are the same in the example.

Poppers beneath the initial height of the screen should not always be hidden.

### What went wrong?

Adding `overflow-y: hidden` to the html tag works.  But I noticed this breaks how `window.pageYOffset` works.  So I don't see that as a fix.

### Any other comments?

Changing to earlier versions of popper can fix this issue, like `v2.3.3`

atomiks:
Since the body might not extend the entire scrollable area of the document it shouldn't be included as a clippingParent.

Closes #1127 

@skvale can you confirm?

Also @FezVrasta we should apply Prettier 2 to the entire codebase since saving files adds those changes each time which are unrelated

### DIFF
--- a/src/dom-utils/getClippingRect.js
+++ b/src/dom-utils/getClippingRect.js
@@ -11,6 +11,7 @@ import getComputedStyle from './getComputedStyle';
 import { isElement, isHTMLElement } from './instanceOf';
 import getBoundingClientRect from './getBoundingClientRect';
 import contains from './contains';
+import getNodeName from './getNodeName';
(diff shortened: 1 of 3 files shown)


skvale:
I used this code in the example from #1127 and it fixes the issue.

Thank you

PR #1144, opened 2020-06-28
fix: paths for ts generation

atomiks:


FezVrasta:
Umh wondering if there's some eslint rule to make sure imports don't have trailing slashes. Or maybe we can fix the ts script to handle them

PR #1148, opened 2020-07-03
fix: relative html 2

atomiks:
The original PR did not fix the problems correctly, just if the `html` element had margins.

Remaining non-critical issue: `<html>` position relative + `<body>` with margin = offset incorrectly (by the margin amount).


src/modifiers/computeStyles.js
```diff
@@ -69,22 +69,32 @@ export function mapToStyles({
 
   if (adaptive) {
     let offsetParent = getOffsetParent(popper);
+    let heightProp = 'clientHeight';
```
atomiks, line 72: Changes to this file fixes https://github.com/popperjs/popper-core/issues/1138

src/dom-utils/getOffsetParent.js
```diff
@@ -77,8 +62,9 @@ export default function getOffsetParent(element: Element) {
 
   if (
     offsetParent &&
-    getNodeName(offsetParent) === 'body' &&
-    getComputedStyle(offsetParent).position === 'static'
+    (getNodeName(offsetParent) === 'html' ||
```
atomiks, line 65: Firefox returns `html` as the offsetParent, so this change fixes https://github.com/popperjs/popper-core/issues/1114

src/modifiers/computeStyles.js
```diff
@@ -69,22 +69,32 @@ export function mapToStyles({
...
+      }
     }
 
     // $FlowFixMe: force type refinement, we compare offsetParent with window above, but Flow doesn't detect it
     /*:: offsetParent = (offsetParent: Element); */
 
     if (placement === top) {
       sideY = bottom;
-      y -= offsetParent.clientHeight - popperRect.height;
+      // $FlowFixMe
```
atomiks, line 89: Can't remember how to fix this in Flow

diondiondion:
Hi @atomiks do you have an update on this PR? It seems to be fixing some issues around adaptive positioning like [the one I reported here](https://github.com/popperjs/popper-core/issues/1138#issuecomment-654903800), so it would be great to get this merged unless it's breaking anything else?

FezVrasta:
@atomiks should we merge this?

atomiks:
Yes. It had that one small problem I mentioned but it's not that major

PR #1158, opened 2020-07-30
Use element rect for detecting overflow

jacwright:
Modifiers may alter `state.rects.element` (e.g. adjusting x/y for elements within iframes), and `detectOverflow` should use the state value instead of re-calculating the rect.

fix-example

iframe modifier being used:

```ts
import { ModifierArguments, Modifier, Options } from '@popperjs/core/lib/types';


function frameAdjust({ state, name }: ModifierArguments<Options>) {
  const { x, y } = frameOffsets(state.elements.reference as Element);
  state.rects.reference.x += x;
  state.rects.reference.y += y;
}

export default ({
  name: 'frameAdjust',
  enabled: true,
  phase: 'beforeRead',
  fn: frameAdjust,
  data: {},
} as Modifier<'frameAdjust', Options>);



export function frameOffsets(target: Element) {
  let x = 0, y = 0;
  while (target && target.ownerDocument) {
    const frame = target.ownerDocument
      && target.ownerDocument.defaultView
      && target.ownerDocument.defaultView.frameElement;

    if (frame) {
      const rect = frame.getBoundingClientRect();
      x += rect.left;
      y += rect.top;
    }
    target = frame;
  }
  return { x, y };
}

```


jacwright:
This PR breaks tests. I'm not sure what to do to fix the tests while still providing the underlying fix. I need guidance. Is this something we should even try to do here?

I've got my own version of detectOverflow and preventOverflow that does what this PR does. I am using it in my own project to make this work correctly and everything seems fine. I'm not sure what cases it breaks.

FezVrasta:
The idea is really good, but without green tests I can't proceed with the merge. Tests can get ran locally to observe what's wrong with them. I'm closing this for now since there wasn't any recent activity but a new PR to introduce this would be awesome.

PR #1162, opened 2020-08-26
restructure to make library more tree-shakable without importing built code

### LINKED ISSUE #1161: Using popperjs-lite breaks bundling behavior

My team runs a design system for use in other apps. All of our components have `module` and `main` set up so that when you import our component the consumer's bundler can decide on what code to use.

Recently we switched from the main `createPopper` to the `createPopper` from `@popperjs/core/lib/popper-lite`. But now that this import path goes directly to ESM code, bundlers can't choose the right `popper-lite` to bundle with. This has led to consumers of our component to have to edit their webpack builds to whitelist popperjs as a `node_module` to build.

This behavior carries a lot of friction.

I can use `popper-lite` without directly accessing any particular file.

### Any other comments?

My solution for this is pretty simple, only requires a little bit of restructuring, and is not a breaking change.

1. Move all code from `src/index.js` to `src/createPopper.js`
2. Make `src/index.js` contain only export statements

```js
export type * from './types';
export * from './enums';
export * from './createPopper';
export { createPopper as createPopperLite } from './popper-lite';
```

3. Change all docs to suggest importing from

```js
import { createPopper } from "@popperjs/core/lib/popper-lite";
```

to

```js
import { createPopperLite as createPopper } from "@popperjs/core";
```

Now since consumers of `@popperjs/core` aren't importing a built file bundlers can choose the correct file.

hipstersmoothie:
This PR bring about a few changes to help with tree-shaking when using popper-lite.

Previously:

- Library suggested to import directly from compiled output (popper-lite and modifiers)
  - this breaks how bundlers choose the right code to compile with (ex: cjs vs esm)
  - you also have to configure jest to compile the esm code for your tests

This experience is not the best, we should just let the tools choose the right code!

So instead of importing directly from `lib/` this PR restructures the code so that people can instead import directly from `@popperjs/core` then let bundlers shake the tree.

Main Changes:

- Change `src/index.js` to [only export things](https://github.com/popperjs/popper-core/pull/1162/files#diff-1fdf421c05c1140f6d71444ea2b27638). This allows for the bundlers to choose the right code and for tree shaking to work.
  - This file now exports all the modifiers and createPopperLite
- [Change the `module` field to that index file](https://github.com/popperjs/popper-core/pull/1162/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R7)
- [Export modifiers in `src/popper.js` so the CJS output exports the same code](https://github.com/popperjs/popper-core/pull/1162/files#diff-feb09859c7c2fa2a0a93fcf461e627caR35)

closes #1161


hipstersmoothie:
This should be a patch release and not a minor

FezVrasta:
Thanks for the pr! Are you sure bundlers are still able to properly three shake unused parts?

hipstersmoothie:
So testing this out with the following file before any of my changes

Input:

```js
import { createPopper } from '@popperjs/core';

createPopper()
```

Result:

Screen Shot 2020-08-27 at 12 27 48 PM

Input

```js
import { createPopper } from '@popperjs/core/lib/popper-lite';

createPopper()
```

Result:

Screen Shot 2020-08-27 at 12 28 40 PM

---

Now with my changes

Input:

```js
import { createPopper } from '@popperjs/core';

createPopper()
```

Result:

Screen Shot 2020-08-27 at 12 29 57 PM

Input

```js
import { createPopperLite } from '@popperjs/core';

createPopperLite()
```

Result:

Screen Shot 2020-08-27 at 12 30 27 PM

FezVrasta:
@atomiks could you find some time to quickly review this PR? it looks good to me but one more pair of eyes would be great

FezVrasta:
Merged! I think we also need to update the examples in the [docs](https://github.com/popperjs/website)

PR #1166, opened 2020-09-07
Properly check if we're inside Shadow DOM

atlas-Fox:
> The getRootNode() method of the Node interface returns the context object's root, which optionally includes the shadow root if it is available.
> https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

> The root of an object is itself, if its parent is null, or else it is the root of its parent. The root of a tree is any object participating in that tree whose parent is null.
> https://dom.spec.whatwg.org/#concept-tree-root

When dealing with detached DOM - their `ParentNode` is null, thus `getRootNode()` returns the node itself. 

Popper uses **ducktyping** to determine whether returned object represents `ShadowRoot` or not. It does it by checking for presence of `{probablyShadowRoot}.host` property.
This has a possibility to clash with Location Interface via https://developer.mozilla.org/en-US/docs/Web/API/Location/host 

If specified node happens to be a detached Anchor - it will be (wrongly) recognised by Popper as ShadowDom structure. This will cause Popper to fail when trying to loop trough parent nodes.

This PR removes **ducktyping** and replaces it with proper object comparison.

### DIFF
--- a/src/dom-utils/contains.js
+++ b/src/dom-utils/contains.js
@@ -1,15 +1,15 @@
 // @flow
+import { isShadowRoot } from './instanceOf';
+
 export default function contains(parent: Element, child: Element) {
   const rootNode = child.getRootNode && child.getRootNode();
-  // $FlowFixMe: Node is not aware of host
-  const isShadow = Boolean(rootNode && rootNode.host);
 
   // First, attempt with faster native method
   if (parent.contains(child)) {
     return true;
   }
   // then fallback to custom implementation with Shadow DOM support
-  else if (isShadow) {
+  else if (isShadowRoot(rootNode)) {
     let next = child;
     do {
       if (next && parent.isSameNode(next)) {
--- a/src/dom-utils/contains.test.js
+++ b/src/dom-utils/contains.test.js
@@ -38,3 +38,18 @@ it('returns true that a node contains itself', () => {
   const element = document.createElement('div');
   expect(contains(element, element)).toBeTruthy();
 });
+
+it('returns correct results for detached DOM with anchor as root', () => {
+  const element = document.createElement('a');
+  element.href = '0.0.0.0';
+  const child = document.createElement('span');
+  const sibling = document.createElement('span');
+  const grandChild = document.createElement('span');
+  element.appendChild(child);
+  element.appendChild(sibling);
+  child.appendChild(grandChild);
+
+  expect(contains(element, child)).toBeTruthy();
+  expect(contains(element, grandChild)).toBeTruthy();
+  expect(contains(sibling, grandChild)).toBeFalsy();
+});
--- a/src/dom-utils/instanceOf.js
+++ b/src/dom-utils/instanceOf.js
@@ -17,4 +17,12 @@ function isHTMLElement(node) {
   return node instanceof OwnElement || node instanceof HTMLElement;
 }
 
-export { isElement, isHTMLElement };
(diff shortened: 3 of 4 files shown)


src/dom-utils/contains.js
```diff
@@ -1,7 +1,9 @@
 // @flow
+import {isShadowRoot} from "./instanceOf";
+
 export default function contains(parent: Element, child: Element) {
   // $FlowFixMe: hasOwnProperty doesn't seem to work in tests
-  const isShadow = Boolean(child.getRootNode && child.getRootNode().host);
+  const isShadow = Boolean(child.getRootNode && isShadowRoot(child.getRootNode()));
```
jspangler, line 6: @atlas-Fox please be aware of the PR I submitted (#1174) which does an additional check to ensure the root node from `getRootNode()` is not null. The subsequent calls within `isShadowRoot()` make checks on a possible `null` object and fail.

Lichurec: Hi!
Thanks for the comment… I'd still argue this is a bad approach as it uses ducktyping and the `child.getRootNode().host` **many not be null** even if that element is not with(in) shadow root.

atlas-Fox: Hi!
Thanks for the comment… I'd still argue this is a bad approach as it uses **ducktyping**, the `child.getRootNode()` **many not be null** and the non-null `{probablyShadowRoot}.host` **may exist** even if child does not contain shadow root at all.

The scenario when that happens is described in this PR && showed in the test provided.

PS
Re-commenting as I did not notice I was logged on my private account and this PR is strictly work-related.

Best regards, 
Fox

jspangler: I appreciate the follow up! You are correct, this should never be an issue and that method should never return `null`, but we use popper within salesforce lightning and lightning locker is a layer over the normal web API's and their version of `getRootNode()` returns `null` in certain scenarios causing Popper to break. I understand this is not something popper should handle since it should never be a thing and I am looking into how to alert Salesforce to their issue but this check did remedy our situation.

FezVrasta:
Please could you add `ShadowRoot: ShadowRoot` to the `Window` type in the `types.js` file?

atlas-Fox:
@FezVrasta sure, no problem…

atlas-Fox:
@FezVrasta I rebased onto your current master, updated changes to the code so that they're applied properly and added the `ShadowRoot: ShadowRoot` to the `Window` type in the `types.js`

src/dom-utils/contains.js
```diff
@@ -1,15 +1,15 @@
 // @flow
+import { isShadowRoot } from './instanceOf';
+
 export default function contains(parent: Element, child: Element) {
   const rootNode = child.getRootNode && child.getRootNode();
-  // $FlowFixMe: Node is not aware of host
-  const isShadow = Boolean(rootNode && rootNode.host);
```
atlas-Fox, line 5: this abstraction is unnecessary now, hence we're using imported function.

src/dom-utils/contains.js
```diff
@@ -1,15 +1,15 @@
...
 
   // First, attempt with faster native method
   if (parent.contains(child)) {
     return true;
   }
   // then fallback to custom implementation with Shadow DOM support
-  else if (isShadow) {
+  else if (isShadowRoot(rootNode)) {
```
atlas-Fox, line 12: The intend should be clear here without abstraction to const.

Issue #575, opened 2018-03-06
Making popper.js XHTML-friendly


fuweichin:
popper.js 1.12.9 doesn't support XHTML, I mean W3C-validated XHTML running in XHTML mode, yet.

To support XHTML-mode, I modified popper.js source file, replaced all `.nodeName`  with `.nodeName.toUpperCase()`, and finally my Bootstrap dropdown did work.

Any plans to add XHTML support?

FezVrasta:
That's the first time someone asked about support for it. If enough people vote for this feature we can add it for sure.

Artefact2:
I have the same problem when using Bootstrap dropdowns. Here is a minimal example that fails:

~~~
<?xml version="1.0" encoding="utf-8"?>
<html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
    <head>
        <meta charset="utf-8" />
        <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.min.css" />
        <script src="https://code.jquery.com/jquery-3.2.1.slim.min.js" />
        <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.12.9/umd/popper.min.js" />
        <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js" />
    </head>
    <body>
        <div class="btn-group">
            <button type="button" class="btn">Foo</button>
            <button type="button" class="btn dropdown-toggle dropdown-toggle-split" data-toggle="dropdown" />
            <div class="dropdown-menu">
                <a class="dropdown-item" href="#">Bar</a>
                <a class="dropdown-item" href="#">Baz</a>
            </div>
        </div>
    </body>
</html>
~~~

Here is the error that occurs when trying to open the dropdown:

> Error: regexp too big at popper.min.js:4:602

Prideth:
I have a similar problem with xhtml. It does not support html in for example title attribute. Is there a way to display a tooltip with custom html and not using it with the title attribute?

ylebhan:
I have exactly the same problem with the Bootstrap dropdowns. At least, it should be mentioned in the BS documentation. 
in XHTML, Bootstrap dropdowns work fine within a navbar, but not outside. This is quite confusing. After further readings, I found out in the BS documentation that popper.js in not used In the first case.

atomiks:
Why are you guys using XHTML?

[...]

PR #1172, opened 2020-09-18
Support XHTML nodeNames

cpburnz:
Hello,

This pull request fixes XHTML support for the 1.x branch. I realize v1 is deprecated. I would upgrade to v2 but v1 is a dependency of Bootstrap 4.

The pre-commit hook tests passed with 342 tests completed and 6 tests skipped. According to FezVrasta's last comment in #575 you only want a clean fix. I've addressed all uses of `.nodeName` using a `getNodeName()` utility function similar to the fix in the 2.x branch.


FezVrasta:
Hi, I'm sorry but I'm not accepting contributions for the 1.x branch, my message was from a long time ago before 2.x was released. Only security vulnerabilities patches are accepted.

PR #1174, opened 2020-09-22
Additional check for rootNode

jspangler:
This is an issue we are facing in Salesforce Lightning Locker where `getRootNode()` is returning `undefined` thus resulting in an error checking for `host`.

### DIFF
--- a/src/dom-utils/contains.js
+++ b/src/dom-utils/contains.js
@@ -1,8 +1,8 @@
 // @flow
 export default function contains(parent: Element, child: Element) {
   // $FlowFixMe: hasOwnProperty doesn't seem to work in tests
-  const rootNode = child.getRootNode && child.getRootNode();
-  isShadow = Boolean(rootNode && rootNode.host);
+  const rootNode = child.getRootNode && child.getRootNode(),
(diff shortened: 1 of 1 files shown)


src/dom-utils/contains.js
```diff
@@ -1,7 +1,8 @@
 // @flow
 export default function contains(parent: Element, child: Element) {
-  // $FlowFixMe: hasOwnProperty doesn't seem to work in tests
-  const isShadow = Boolean(child.getRootNode && child.getRootNode().host);
+  const rootNode = child.getRootNode && child.getRootNode(),
+    // $FlowFixMe: Node is not aware of host
+    isShadow = Boolean(rootNode && rootNode.host);
```
atomiks, line 5: Can you put this as a separate `const` statement?

jspangler:
Appreciate the quick review and merge

jspangler:
@FezVrasta do you know when this will go out in a release?

FezVrasta:
I'm publishing 2.5.1

jspangler:
Thank you!

Lichurec:
@jspangler  could you provide any tests for your case scenario?

Issue #1159, opened 2020-08-05
Unable to get property 'toString' of undefined for node objects


gabskoro:
Hi guys, after I updated `@popperjs/core` from `v2.4.2` to `v.2.4.4` I started getting those weird errors on `IE11` and `Edge` about node object being empty in some cases. Unfortunately, can't reproduce it in my app :(

**Here is my Stacktrace**

![screenshot-app bugsnag com-2020 08 05-13_58_13](https://user-images.githubusercontent.com/2737390/89410363-e5f3aa80-d723-11ea-9b15-9da5791a3a8e.png)

Any thoughts?

cb-eli:
Happened to me as well, a fix will be very appreciated.

cb-eli:
Pinned to version 2.5.2 and everything is working.
The problem is in version 2.5.3

gabskoro:
Tnx guys 👍🏻

guillaume86:
I have a very similar error in IE11 (popper v2.6.0) but with a different stack trace:

```
TypeError: Unable to get property 'toString' of undefined or null reference
  at getWindow(./node_modules/@popperjs/core/lib/dom-utils/getWindow.js:5:3)
  at isHTMLElement(./node_modules/@popperjs/core/lib/dom-utils/instanceOf.js:15:3)
  at Anonymous function(./node_modules/@popperjs/core/lib/dom-utils/getOffsetParent.js:33:10)
  at getOffsetParent(./node_modules/@popperjs/core/lib/dom-utils/getOffsetParent.js:57:3)
  at Anonymous function(./node_modules/@popperjs/core/lib/dom-utils/getClippingRect.js:38:7)
  at getClippingRect(./node_modules/@popperjs/core/lib/dom-utils/getClippingRect.js:53:3)
  at detectOverflow(./node_modules/@popperjs/core/lib/utils/detectOverflow.js:34:7)
  at state(./node_modules/@popperjs/core/lib/modifiers/hide.js:32:7)
  at instance(./node_modules/@popperjs/core/lib/createPopper.js:183:13)
  at Anonymous function(./node_modules/@popperjs/core/lib/createPopper.js:196:11)
  at u(/v3/polyfill.min.js:3:44570)
  at tn(/v3/polyfill.min.js:3:46995)
  at resolve(./node_modules/@popperjs/core/lib/createPopper.js:195:9)
  at Anonymous function(./node_modules/@popperjs/core/lib/utils/debounce.js:8:11)
[...]

PR #1185, opened 2020-10-30
Fix: IE11 does not support getRootNode

ktmouk:
IE11 does not support `getRootNode`. 
When `parent.contains(child)` is `false`, IE11 raises `Unable to get property 'toString'` error.　
This PR adds a check that `rootNode` is not `undefined`.

https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode


ptu14:
I checked and it fixes #1159

FezVrasta:
thanks!

PR #1189, opened 2020-11-09
fix #1182: margin not cleaned up properly

### LINKED ISSUE #1182: Margin-styling lingers on element after popper instance is destroyed
### CodeSandbox demo

https://codesandbox.io/s/popper-margin-lingering-myytz

1. Open CodeSandbox link
2. Click toggle-button ("show) in renderer (notice the popup is empty so nothing will be shown)
3. Click toggle-button ("hide") again

My expectation is that the element that was used as popup would have been stripped of all element styling after the Popper instance was destroyed.

### What went wrong?

`margin`-styling lingered on the popup-element

### Any other comments?

This might not seem that important to most, I understand. I happen to work on a project where an element will be shown with Popper when the user's view-size is below a certain threshold. Bigger view-sizes will show a different behavior. The view size can be changed by the user, and this means that both options can occur relatively easily.

So far my own brief research led me to `modifiers/applyStyles.js`. The `effect`-function sets the margin-styling as part of the `initialStyles`, but the cleanup-function it returns will take the style defined in `state.styles.popper` to decide which styles to cleanup. This one lacks the `margin`-style for some reason.

RonaldTreur:
After destroying a Popper instance, an inline `margin: 0`-style remains on the popped-up element. Since there seems to be no clear reason why a margin should be set on a Popper instance, this patch makes sure it is not added to the popped-up element to begin with.

### DIFF
--- a/src/modifiers/applyStyles.js
+++ b/src/modifiers/applyStyles.js
@@ -40,7 +40,6 @@ function effect({ state }: ModifierArguments<{||}>) {
       position: state.options.strategy,
       left: '0',
       top: '0',
-      margin: '0',
     },
     arrow: {
       position: 'absolute',


atomiks:
The margin styling was added to prevent people from adding margins to their poppers because of inherent issues with it. It's not strictly necessary though, so maybe we can just remove it all.

RonaldTreur:
> The margin styling was added to prevent people from adding margins to their poppers because of inherent issues with it. It's not strictly necessary though, so maybe we can just remove it all.

Hi @atomiks, I figured as much :-) I understand throwing a warning if people try to use margins, but I think enforcing they can't apply any is indeed not necessary.

atomiks:
@RonaldTreur in this case we should remove the warnings in the main `index.js` file too

RonaldTreur:
Hi folks (@atomiks), is there anything I can do to help this PR along? Please let me know :-)

FezVrasta:
Sorry for the lack of communication, I think this would be a breaking change. A possible non-breaking solution would be to fix the fact the margin is not unset on destroy though.

FezVrasta:
I'll close this because of inactivity, feel free to send a new one with the required changes though.

PR #1191, opened 2020-11-13
Transpile to browser compatible process.env.NODE_ENV check

### LINKED ISSUE #1188: Test "__DEV__" in a browser compatible way.
When attempting to get browser optimized builds of the library via `@popperjs/core/dist/esm/...` paths the files in question don't seem to have access to types that would be needed to leverage them in a Typescript project. Can the build process be tweaked to include this?

Westbrook:
fixes #1188

Based on https://github.com/popperjs/popper-core/issues/1188#issuecomment-725340952, replace `__DEV__` instances in source with `typeof process !== 'undefined' && process.env.NODE_ENV !== "production"` which is browser compatible, whereas `process.env.NODE_ENV !== "production"` directly is not.

*Benefits*
- allow default entry points to be leveraged in the browser
- allow browser compatible JS to also be associated with TS definitions correctly

*Possibilities*
- this opens the possibility to remove some of the output in `/dist` as consumers no longer need to use an alternative build to leverage PopperJS in browser (this is likely a breaking change, so I've not included this change in this PR)

### DIFF
--- a/.config/babel.config.js
+++ b/.config/babel.config.js
@@ -29,7 +29,17 @@ module.exports = {
             },
           ],
         ]
-      : ['dev-expression']),
+      : [
+          [
+            'inline-replace-variables',
+            {
+              __DEV__: {
+                type: 'node',
+                replacement: `typeof process !== 'undefined' && process.env.NODE_ENV !== "production"`,
+              },
+            },
+          ],
+        ]),
     'annotate-pure-calls',
   ],
   env: {
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "babel-jest": "^25.5.1",
     "babel-plugin-add-import-extension": "^1.3.0",
     "babel-plugin-annotate-pure-calls": "^0.4.0",
-    "babel-plugin-dev-expression": "^0.2.2",
     "babel-plugin-inline-replace-variables": "^1.3.1",
     "babel-plugin-transform-inline-environment-variables": "^0.4.3",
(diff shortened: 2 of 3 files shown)


FezVrasta:
Thanks! I suppose we could deprecate the browser build at this point?

Westbrook:
Hard to say what makes sense as far as the offering here with the builds. It's still pretty common to want fully built dependencies, no matter how they might negatively effect the output of an application, so some users may not be prepared for a full removal. Setting a deprecation warning might be a good idea, especially if you actually saw a breaking version on the horizon. If it's not soon, it might not be worth investing in that yet 🤷

FezVrasta:
Popper has no dependencies so that's not an issue, but I guess it makes sense to keep that dist around since I don't see any other reason to schedule a breaking release

.config/babel.config.js
```diff
@@ -29,7 +29,17 @@ module.exports = {
...
           ],
         ]
-      : ['dev-expression']),
+      : [
+          [
+            'inline-replace-variables',
+            {
+              __DEV__: {
+                type: 'node',
+                replacement: `typeof process !== 'undefined' && process.env.NODE_ENV !== "production"`,
```
TrySound, line 38: This will force webpack 4 to include process polyfill. With rollup or webpack 5 it won't really work in browser because typeof process always will be "undefined". This also break dead code elimination. Falsy branch will not be dropped in production.

Issue #1169, opened 2020-09-09
Popper is not correctly placed


piecyk:
### CodeSandbox demo

https://codesandbox.io/s/elated-sid-sgwqg ( refresh codesandbox embedded browser )  

### Steps to reproduce the problem

No steps required, popper is not correctly placed 

<img width="682" alt="Screenshot 2020-09-09 at 16 23 30" src="https://user-images.githubusercontent.com/82964/92611187-d2a2a480-f2b8-11ea-95bd-c33e2eb5c5dc.png">

### What is the expected behavior?

Popper should be correctly placed 

### What went wrong?

Looks like in flex box container, if element before has some non integer width ( as in this example using content editable `aaas` has width of `27.531` ) popperjs fails to correctly place the popper  

### Any other comments?

Looks like it's connected with #713

davoam:
I found that it happens if `getBoundingClientRect` returns not integer `left` value for the trigger element (e.g. `31.5 px`). In this case popper position is less by 1px from proper one. In my case it happens only when `fixed` strategy is used.

piecyk:
Also one possible workaround is to change display of anchor element to `inline-table` That would make element to render width/height as integers, so we don't get sub-pixels.

piecyk:
Thinking on it, why we need the rounding at all ? Looking on modern browser all would handle it 🤔 Maybe given option to opt out from rounding would be not that bad idea? WDYT?

piecyk:
So something like this https://github.com/popperjs/popper-core/compare/master...piecyk:feat/round-option With `round: false` user can opt out from rounding. 

```
{
  name: 'popperOffsets',
  options: {
    round: false,
  },
},
```

Tested on firefox, safari, edge and looks great with pixel perfect position of popper. On chrome looks like there is a bug setting position with `translate3d/translate` makes it not correctly placed, but using left/top it is. IMHO there should be no difference. Need to check if someone reported a bug. Till then we need to fallback to left/top with disabling gpuAcceleration on `computeStyles`

piecyk:
Example of those inaccuracies in chrome/edge between position `left/top` vs `translate3d/translate` what i was writing above  https://codesandbox.io/s/cool-chandrasekhar-mcbp1

PR #1193, opened 2020-11-17
feat: add `roundOffsets` option to computeStyles modifier

piecyk:
As reported in #1169 rounding logic in some cases makes the popper not correctly aligned with reference element. This PR adds possibility to opt out from it via roundOffsets option.

```js
  modifiers: [
    {
      name: 'computeStyles',
      options: {
        roundOffsets: false,
      },
    },
  ],
});
```

### DIFF
--- a/src/utils/computeOffsets.js
+++ b/src/utils/computeOffsets.js
@@ -68,13 +68,11 @@ export default function computeOffsets({
     switch (variation) {
       case start:
         offsets[mainAxis] =
-          Math.floor(offsets[mainAxis]) -
-          Math.floor(reference[len] / 2 - element[len] / 2);
+          offsets[mainAxis] - (reference[len] / 2 - element[len] / 2);
         break;
       case end:
         offsets[mainAxis] =
-          Math.floor(offsets[mainAxis]) +
-          Math.ceil(reference[len] / 2 - element[len] / 2);
+          offsets[mainAxis] + (reference[len] / 2 - element[len] / 2);
         break;
       default:
     }
--- a/src/modifiers/computeStyles.js
+++ b/src/modifiers/computeStyles.js
@@ -18,6 +18,7 @@ import getBasePlacement from '../utils/getBasePlacement';
 export type Options = {
   gpuAcceleration: boolean,
   adaptive: boolean,
+  roundOffsets: boolean,
 };
 
 const unsetSides = {
@@ -30,7 +31,7 @@ const unsetSides = {
 // Round the offsets to the nearest suitable subpixel based on the DPR.
 // Zooming can change the DPR, but it seems to report a value that will
 // cleanly divide the values into the appropriate subpixels.
-function roundOffsets({ x, y }): Offsets {
+function roundOffsetsByDPR({ x, y }): Offsets {
   const win: Window = window;
   const dpr = win.devicePixelRatio || 1;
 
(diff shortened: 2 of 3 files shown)


FezVrasta:
Can't we fix the rounding algorithm instead? We have quite a lot of tests that cover it already.

piecyk:
> Can't we fix the rounding algorithm instead? We have quite a lot of tests that cover it already.

I think it will be hard to make a fix that will work cross browsers.

Also while testing locally latest chrome/firefox/edge/safari with opting out, didn't break the rendering as mainly that's the reason for rounding.

src/utils/computeOffsets.js
```diff
@@ -68,17 +84,15 @@ export default function computeOffsets({
     switch (variation) {
       case start:
         offsets[mainAxis] =
-          Math.floor(offsets[mainAxis]) -
```
piecyk, line 71: Why we had this rounds functions on start/end variation? Imho they are not needed as  latter we still round the offsets 🤔

src/utils/computeOffsets.js
```diff
@@ -7,23 +7,39 @@ import type {
...
+// cleanly divide the values into the appropriate subpixels.
+function roundOffsets({ x, y }): Offsets {
+  const win: Window = window;
+  const dpr = win.devicePixelRatio || 1;
+
+  return {
+    x: Math.round(x * dpr) / dpr || 0,
+    y: Math.round(y * dpr) / dpr || 0,
+  };
+}
```
piecyk, line 25: I think there is no way to correctly round the offsets using `devicePixelRatio` Maybe we can just go with something simpler, ping @atomiks 

```js
// Round the offsets to the nearest subpixel.
function roundOffsets({ x = 0, y = 0 }): Offsets {
  return {
    x: Math.round(x * 10) / 10 || 0,
    y: Math.round(y * 10) / 10 || 0,
  };
}
```

piecyk:
Don't really see a way to improve rounding algorithm to be more accurate 🤷‍♂️ Maybe till then we can have this opt out option.

src/utils/computeOffsets.js
piecyk: Let's keep the rounding logic in computeStyles.

piecyk:
Changed this to keep the rounding logic in computeStyles modifier, added `roundOffsets` option as boolean to modifier options as in PR description. Was wondering also if not allow to pass custom function, so then user of modifier can tweak the rounding.

src/utils/computeOffsets.js
FezVrasta: @atomiks? 👀

piecyk: Maybe I should add more test 🤔 Also this change didn't break any existing one.

FezVrasta: The PR looks good, I'm just a bit worried about this bit. Let's give @atomiks a couple more days to review it and if he doesn't find time I'll merge it as dog food it to the whole user base :-)

piecyk: Looks like no other comments. IMHO let's merge this 🤞 💪

FezVrasta:
Merged, thanks for the help! Could you also send a PR on the `computeStyles` docs? [here](https://github.com/popperjs/website/edit/master/src/pages/docs/v2/modifiers/compute-styles.mdx)

piecyk:
> Could you also send a PR on the computeStyles docs? 

Yes, will do.

piecyk:
@FezVrasta when are you planing releasing this? I think patch release will be good in this case. Also thanks for this 👏

FezVrasta:
Releasing

PR #1213, opened 2021-01-08
fix: improve computeStyles modifier with customize roundOffsets support

allex:
> For decimal px value may cause transform and keyframe animation shake issues,

This PR will improve computeStyles option `roundOffsets` with customize implement ability.
also fix the builtin `roundOffsetsByDPR` with a real round returns. 

eg. 
```
JSON.stringify(state.styles.popper, null, 2)
"{
  "position": "absolute",
  "top": "auto",
  "right": "auto",
  "bottom": "217.6px", // <<<---
  "left": "440.3px",  // <<<---
  "transform": ""
}"

=> "{
  "position": "absolute",
  "top": "auto",
  "right": "auto",
  "bottom": "218px",  // <<<---
  "left": "440px",  // <<<---
  "transform": ""
}"
```

### DIFF
--- a/src/modifiers/__snapshots__/computeStyles.test.js.snap
+++ b/src/modifiers/__snapshots__/computeStyles.test.js.snap
@@ -52,3 +52,25 @@ Object {
   "transform": "",
 }
 `;
+
+exports[`computes the popper styles 5`] = `
+Object {
+  "bottom": "auto",
+  "left": "auto",
+  "position": "absolute",
+  "right": "-112px",
+  "top": "8px",
+  "transform": "",
+}
+`;
+
+exports[`computes the popper styles 6`] = `
+Object {
+  "bottom": "auto",
+  "left": "auto",
+  "position": "absolute",
+  "right": "-110.3px",
+  "top": "5.83px",
+  "transform": "",
+}
+`;
--- a/src/modifiers/computeStyles.js
+++ b/src/modifiers/computeStyles.js
@@ -15,11 +15,16 @@ import getComputedStyle from '../dom-utils/getComputedStyle';
 import getBasePlacement from '../utils/getBasePlacement';
 
 // eslint-disable-next-line import/no-unused-modules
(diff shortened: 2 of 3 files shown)


FezVrasta:
I think we removed the rounding in [this PR](https://github.com/popperjs/popper-core/pull/1193/files#r525121167), the idea was to leave the rounding to the computeStyles, I'm not sure why this is happening though.

piecyk:
> the idea was to leave the rounding to the computeStyles

Yes exactly. 

>  I'm not sure why this is happening though.

It's happening because the `roundOffsetsByDPR` will not always return an integer. Rounding by default is bad idea. 

This can be addressed by allowing to use custom roundOffsets function from options, for example

```js
  modifiers: [
    {
      name: 'computeStyles',
      options: {
        roundOffsets: ({ x, y }) => ({
          x: Math.round(x),
          y: Math.round(y)
        })
      },
    },
  ],
});
```

allex:
> > the idea was to leave the rounding to the computeStyles
> 
> Yes exactly.
> 
> > I'm not sure why this is happening though.
> 
> It's happening because the `roundOffsetsByDPR` will not always return an integer. Rounding by default is bad idea.
> 
> This can be addressed by allowing to use custom roundOffsets function from options, for example
> 
> ```js
> createPopper(reference, popper, {
>   modifiers: [
>     {
>       name: 'computeStyles',
>       options: {
>         roundOffsets: ({ x, y }) => ({
>           x: Math.round(x),
>           y: Math.round(y)
>         })
>       },
>     },
>   ],
> });
> ```

Still not work by customize `option.roundOffsets`.
It's seems roundOffsets just be a bool for turn on/of the builtin `roundOffsetByDPR`

https://github.com/popperjs/popper-core/blob/24cb3a88d5108d2a42126c8ab3899ee95259ef18/src/modifiers/computeStyles.js#L21

FezVrasta:
Oh my bad I thought it was already implemented, feel free to send a PR to add this feature.

PR #1226, opened 2021-01-28
fix: check for body not based on tagName

gilles-yvetot:
I have been using popper inside my extension so I don't always control the page it is run on.
Some salesforce popup do not have a body but `element.ownerDocument.body` references an element `<frameset>`. 
It was causing an infinite loop trying to loop for all the scroll parents.

This PR is just to ensure we check for this body (whatever its tagName)

### DIFF
--- a/src/dom-utils/listScrollParents.js
+++ b/src/dom-utils/listScrollParents.js
@@ -1,7 +1,6 @@
 // @flow
 import getScrollParent from './getScrollParent';
 import getParentNode from './getParentNode';
-import getNodeName from './getNodeName';
 import getWindow from './getWindow';
 import type { Window, VisualViewport } from '../types';
 import isScrollParent from './isScrollParent';
--- a/src/dom-utils/listScrollParents.js
+++ b/src/dom-utils/listScrollParents.js
@@ -17,7 +17,7 @@ export default function listScrollParents(
   list: Array<Element | Window> = []
 ): Array<Element | Window | VisualViewport> {
   const scrollParent = getScrollParent(element);
-  const isBody = getNodeName(scrollParent) === 'body';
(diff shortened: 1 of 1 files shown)


FezVrasta:
wow I'm always amazed by the amount of quirkness introduced by SalesForce 😬 Thanks for the PR!

XhmikosR:
@FezVrasta it seems this breaks tests for us with `TypeError: Cannot read property 'body' of undefined`: https://github.com/twbs/bootstrap/runs/1930894488

I'll make a new issue tomorrow if the issue persists :)

FezVrasta:
ownerDocument seems very well supported, isn't maybe a jsdom issue?

https://caniuse.com/mdn-api_node_ownerdocument

XhmikosR:
The tests fail in Node.js 10.x and 12.x and [BrowserStack](https://github.com/twbs/bootstrap/runs/1930895646) where we support various browsers. Even Chrome 86 fails. The call stacks all point to popper.

I haven't had a close look yet, but I thought I'd mention the issue.

FezVrasta:
I'm a bit confused because we already read `element.ownerDocument.body` in several other places in the library, so I'm not sure why is it erroring in this specific one 🤔  also our integration tests that run on Chrome are good..

FezVrasta:
@XhmikosR could you try with 2.7.1?

XhmikosR:
Unfortunately, it definitely fails for us :/

Branch: https://github.com/twbs/bootstrap/compare/main-xmr-update-popper

Both Node.js and BrowserStack.

XhmikosR:
> 
> 
> @XhmikosR could you try with 2.7.1?

Confirmed, seems to work!

PR #1228, opened 2021-01-29
feat: apply style if different

gilles-yvetot:
I am using mutation observer in my code and popper is triggering an infinite loop because the attributes get set over and over even if they are identical. I figured a clean fix could be to only apply changes if there is a diff.

Let me know what you think

### DIFF
--- a/src/modifiers/applyStyles.js
+++ b/src/modifiers/applyStyles.js
@@ -18,17 +18,21 @@ function applyStyles({ state }: ModifierArguments<{||}>) {
       return;
     }
 
-    // Flow doesn't support to extend this property, but it's the most
-    // effective way to apply styles to an HTMLElement
-    // $FlowFixMe[cannot-write]
-    Object.assign(element.style, style);
+    Object.entries(style).forEach(([property, value]) => {
+      if (element.style[property] !== value) {
+        element.style[property] = value;
+      }
+    });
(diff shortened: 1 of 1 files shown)


FezVrasta:
Thanks for the PR.

This adds a performance overhead, we need to read and write to the DOM. We'll need to make sure perf is not impacted.

gilles-yvetot:
@FezVrasta understood. Do you have any performance test ?
Also, I'd love to fix my PR but I am not familiar with Flow, any help on what should I modify?
I am taking a look at the functional test at the same time

FezVrasta:
Hi, unfortunately I don't have any performance tests, I usually open the visual tests and run some performance captures before and after the changes.

gilles-yvetot:
@FezVrasta I am not sure how to do that 😕

FezVrasta:
These are the visual tests, they are just html files you can open with any browser:
https://github.com/popperjs/popper-core/tree/master/tests/visual

You can then use the [performance](https://developers.google.com/web/tools/chrome-devtools/evaluate-performance) panel on Chrome to see how it runs.

gilles-yvetot:
@FezVrasta this PR has been opened for a while now and I unfortunately do not have more time to invest in it :( 
Feel free to close it

PR #1244, opened 2021-02-19
fix: will-change create containig block

ptu14:
fix https://github.com/popperjs/popper-core/issues/1229

since only `will-change` value of `transform` or `perspective` I decided to put this in code

https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block

point 4.2

### DIFF
--- a/src/dom-utils/getOffsetParent.js
+++ b/src/dom-utils/getOffsetParent.js
@@ -34,7 +34,7 @@ function getContainingBlock(element: Element) {
     if (
       css.transform !== 'none' ||
       css.perspective !== 'none' ||
-      (css.willChange && css.willChange !== 'auto')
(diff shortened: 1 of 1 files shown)


FezVrasta:
4.3 mentions `filter` too, should it be included?

ptu14:
I thought about it, but `filter` is just a Firefox thing. I don't want to break Chrome, Edge when someone uses `filter`.

FezVrasta:
I pushed a commit on your PR to address more cases, how does it look?

ptu14:
LGTM. I assume there may be other problems with **Safari** (https://github.com/popperjs/popper-core/issues/1156), but I don't have macOS to test it.

I think this fix is needed anyway, regardless of the second issue

PR #1247, opened 2021-02-20
fix: handle scaled parent elements

### LINKED ISSUE #376: Popper position is wrong with transform: scale
Hi,
wenn I set a transform to scale in the style, positioning of the popup element is wrong. Here is a short example in CodePen:
https://codepen.io/FezVrasta/pen/MWbvEjg

I think it use now scaled positions and sizes to calculate the popup position, but else it should use still orginal size, cause the transform value also effect on the popup

FezVrasta:
fixes #376


XhmikosR:
@FezVrasta https://github.com/twbs/bootstrap/compare/main-xmr-test-popper (ignore the lint failure)

Not sure how to test this since including your branch seems to break building.

FezVrasta:
Sorry, nevermind, there are some of our own E2E tests that are failing, we'll need to fix them first.

rmanalan:
Hi. Just checking to see if there are plans to merge this in once tests pass?

FezVrasta:
I don't have time to fix the code so that tests pass. Contributions are welcome

tholmes-rms:
I've written up some logic to only apply the scale in `getBoundingClientRect` when called in `getCompositeRect` and the `offsetParent` is scaled.

I added a few more test cases for some combinations of reference, popper, and parent elements being scaled. I'm not convinced that this isn't going to break anything else or work for all scales scenarios, but all the tests pass (except one functional test in `scrolling-fixed.test.js` that's also failing for me on `master`?).

If you want to give me access, I can create a branch off of this branch and make a PR if you want.

FezVrasta:
Feel free to branch this out, it'd be great to get this fixed

tholmes-rms:
PR here: https://github.com/popperjs/popper-core/pull/1322 with the above-mentioned changes.

Slapbox:
As far as I can tell, our popper's ancestor elements have no scaling applied, but upgrading from `2.9.2` to `2.9.3` caused our popper locations to change (shifted upwards maybe 16px if I had to guess.)

FezVrasta:
> As far as I can tell, our popper's ancestor elements have no scaling applied, but upgrading from `2.9.2` to `2.9.3` caused our popper locations to change (shifted upwards maybe 16px if I had to guess.)

Please open a new issue

prasanga-peak:
The `scaleX` and `scaleY` values will be set to `Infinity` when `element.offsetWidth` is 0, which results in the popper location change. I'll open a new issue for this.

PR #1258, opened 2021-02-26
Make getWindow null/undefined safe again

guillaume86:
Make getWindow null/undefined safe again to fix some IE11 issues.
See #1159 for details.

### DIFF
--- a/src/dom-utils/getWindow.js
+++ b/src/dom-utils/getWindow.js
@@ -3,7 +3,7 @@
 /*:: declare function getWindow(node: Node | Window): Window; */
 
 export default function getWindow(node) {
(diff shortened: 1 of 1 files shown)


src/dom-utils/getWindow.js
```diff
@@ -3,6 +3,10 @@
 /*:: declare function getWindow(node: Node | Window): Window; */
 
 export default function getWindow(node) {
+  if (node === null || node === undefined) {
```
FezVrasta, line 6:
```suggestion
  if (node == null) {
```

guillaume86:
Thx a lot!

PR #1261, opened 2021-03-01
fix: use 1px check for layout rect

atomiks:
https://github.com/popperjs/popper-core/issues/1223 

@mkrds


mkrds:
@atomiks I've tested it locally and it looks like it has fixed the issue :)

mkrds (approved):

PR #1264, opened 2021-03-08
fix: replace includes with indexOf to support IE11 again

FezVrasta:


stefanroex:
@FezVrasta You're fast with fixing this issue 😄. Cool!

I think we've actually found two problems. The change in this PR removes the use of `Array.prototype.includes`, which isn't supported by IE11. So that's a good change. In our project we've already included a polyfill for that, and the problem in our case was [a few lines above this](https://github.com/popperjs/popper-core/blob/b28e9e647ce833b053be823508f9bf65679024ab/src/dom-utils/getOffsetParent.js#L24) which uses `String.prototype.includes`.

FezVrasta:
I released a new version with this fix

PR #1268, opened 2021-03-29
fix: fixing invalid containing block calculation on IE for fixed strategy

wszydlak:
According to CSS 2.1 (https://www.w3.org/TR/CSS21/visudet.html#containing-block-details) which is used by Internet Explorer for containing block calculation (https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/aa768717(v=vs.85)) i prepared a fix which is obtaining correct containing block (same as IE will use for displaying content on page).

This fix will ensure that we use correct containing blocks for IE 9, 10 and 11.

I cannot paste codepen, cause it has some errors on IE.

However You can check new behavior using code below. I also added core-js to polyfill ES6 functions

```html
<!DOCTYPE html>
<title>Transform Visual Test</title>

<style>
  #transform {
    position: fixed;
    transform: translate(200px, 200px);
  }
  
  #tooltip {
    background-color: #663399;
    padding: 20px;
    width: 200px;
  }
</style>

<div id="transform">
  <button type="button" id="button">I'm a button</button>
  <div id="tooltip">I'm a tooltip</div>
</div>

<script src="dist/umd/popper.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/core-js/3.9.1/minified.js"></script>
<script>

  // Pass the button, the tooltip, and some options, and Popper will do the
  // magic positioning for you:
  Popper.createPopper(button, tooltip, {
    placement: "bottom",
    strategy: "fixed"
  });
</script>
```

### DIFF
--- a/src/dom-utils/getOffsetParent.js
+++ b/src/dom-utils/getOffsetParent.js
@@ -22,6 +22,16 @@ function getTrueOffsetParent(element: Element): ?Element {
 // return the containing block
 function getContainingBlock(element: Element) {
   const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') !== -1;
+  const isIE = navigator.userAgent.indexOf('Trident') !== -1;
+
+  if (isIE && isHTMLElement(element)) {
+    // In IE 9, 10 and 11 fixed elements containing block is always established by the viewport
+    const elementCss = getComputedStyle(element);
+    if (elementCss.position === 'fixed') {
+      return null;
+    }
+  }
+
   let currentNode = getParentNode(element);
 
   while (


src/dom-utils/getOffsetParent.js
```diff
@@ -22,6 +22,16 @@ function getTrueOffsetParent(element: Element): ?Element {
 // return the containing block
 function getContainingBlock(element: Element) {
   const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') !== -1;
+  const isIE11 = navigator.userAgent.indexOf('Trident') !== -1;
```
FezVrasta, line 25: Isn't this more like a generic `isIE`?

wszydlak: Trident applies only for IE11, if we want to check older IE versions we need to have more if’s, do we need this?

FezVrasta: No, I thought Trident was in all IE versions, thanks for the info.

wszydlak: So we could resolve this?

wszydlak: I checked again - you were right, Trident is more generic like isIE

wszydlak: According to this i pushed a change

ptu14: I guess @FezVrasta is right with `Trident` and it can be `isIE` variable. I don't have old IE to test this but found something like this: https://stackoverflow.com/a/27069089

src/dom-utils/getOffsetParent.js
```diff
@@ -22,6 +22,16 @@ function getTrueOffsetParent(element: Element): ?Element {
 // return the containing block
 function getContainingBlock(element: Element) {
   const isFirefox = navigator.userAgent.toLowerCase().indexOf('firefox') !== -1;
+  const isIE = navigator.userAgent.indexOf('Trident') !== -1;
+
+  if (isIE && isHTMLElement(element)) {
+    // In IE 11 fixed elements containing block is always established by the viewport
```
FezVrasta, line 28: Is this comment valid for IE9/10 too?

wszydlak: yes, it is, according to mentioned microsoft docs in PR description, i've changed this comment

wszydlak:
@FezVrasta are there any chances to get this checked and confirmed soon to be released as soon as possible?

PR #1280, opened 2021-04-18
fixed not correct position popper on iOS mobile app (ionic)

Sitronik:
Fixed issue https://github.com/popperjs/popper-core/issues/1279

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "@popperjs/core",
-  "version": "2.4.0",
+  "name": "popperjs-core-ionic-scroll-fix",
+  "version": "2.4.1",
   "description": "Tooltip and Popover Positioning Engine",
   "main": "dist/cjs/popper.js",
   "main:umd": "dist/umd/popper.js",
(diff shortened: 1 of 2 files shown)


FezVrasta:
Hi, thanks for the PR.

I'm not sure this is the best way to handle this situation, if we had to add custom logic for every single non-standard environment out there the code would quickly become unmanageable.

A better approach in my opinion would be to understand what's wrong with the existing logic and correct it so that it can automatically identify the correct element to use.

Sitronik:
> Hi, thanks for the PR.
> 
> I'm not sure this is the best way to handle this situation, if we had to add custom logic for every single non-standard environment out there the code would quickly become unmanageable.
> 
> A better approach in my opinion would be to understand what's wrong with the existing logic and correct it so that it can automatically identify the correct element to use.

Hi, I agree this solution is not very abstract.  But what's the best way to write in this situation?  Could this be a problem in the ionic framework?  Since I found that for the Android mobile app, and for all browsers, including iOS and android, it works correctly, but it does not work only on the iOS mobile app

PR #1281, opened 2021-04-21
Update validateModifiers.js

nekud:
Added break; statement to line 48. This prevents running case: 'phase' each time case: 'enabled' is run.


FezVrasta:
Oh! Thanks! I'm surprised it worked correctly so far 😱

nekud:
Can I just report it as an issue and maybe you can fix it. Not sure why the unit test failed.

PR #1302, opened 2021-05-05
fix: use other direction with end variation

Ayc0:
When we use the placement `left`, the coordinates used aren't `top:0; left:0` anymore but `top:0; right:0`.
Same thing with the placement `top`, the coordinates used become `bottom:0; left:0`.

But using `bottom-end` and `top-end` also also use the coordinate `right:0` was end specifies that the popup should be stuck at the end of the the X axis, aka on the right. (and same with `left-end` and `right-end`, they should be stuck at the bottom).

An effective benefits of flipping the axis is that if the popup grows, the sides that are stuck won't move. So if someone uses `left-end`, the popup will be stuck on the right and on the bottom.

### DIFF
--- a/src/modifiers/__snapshots__/computeStyles.test.js.snap
+++ b/src/modifiers/__snapshots__/computeStyles.test.js.snap
@@ -56,6 +56,50 @@ Object {
 exports[`computes the popper styles 5`] = `
 Object {
   "bottom": "auto",
+  "left": "10px",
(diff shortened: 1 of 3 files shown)


Ayc0:
I never created any PR before for popper so I don't what's the usual process (like for the changelog)

FezVrasta:
Thanks for the PR!

PR #1322, opened 2021-06-04
fix: dynamically include scale in bounding client rect

tholmes-rms:
This only factors scale into the bounding client rect if the offset parent is also scaled.

Honestly, this doesn't feel particularly elegant and I don't know this stuff intimately enough to foresee any potential side effects, but the existing tests pass... 🤷


src/dom-utils/getCompositeRect.js
```diff
@@ -8,16 +8,26 @@ import getWindowScrollBarX from './getWindowScrollBarX';
 import getDocumentElement from './getDocumentElement';
 import isScrollParent from './isScrollParent';
 
+function isElementScaled(element) {
+  const rect = element.getBoundingClientRect();
+  const scaleX = rect.width / element.offsetWidth || 1;
+  const scaleY = rect.height / element.offsetHeight || 1;
+
+  return scaleX !== 1 || scaleY !== 1;
+}
```
tholmes-rms, line 17: Not sure if we'd want this pulled out into a util file?

src/dom-utils/getCompositeRect.js
```diff
@@ -8,16 +8,26 @@ import getWindowScrollBarX from './getWindowScrollBarX';
 import getDocumentElement from './getDocumentElement';
 import isScrollParent from './isScrollParent';
 
+function isElementScaled(element) {
+  const rect = element.getBoundingClientRect();
```
atomiks, line 12: There should be a util for this method

tholmes-rms: The util just adds `x` and `y` to the `rect` returned from `element.getBoundingClientRect` (which we don't need) and optionally does the scaling stuff (which we definitely don't want here). 

So, we could but, we don't _need_ to.

tholmes-rms:
@FezVrasta any thoughts on this?

FezVrasta:
Hey sorry, I'm having some busy weeks. I'll try to take a look as soon as possible

tholmes-rms:
Ok, cool. I'm just pinging :trollface:

bufordeeds (approved):

bufordeeds:
Want just poking around. Please disregard.

tholmes-rms:
bump

FezVrasta:
Umh looks like Flow exploded upon merge :-(

tholmes-rms:
😬

PR #1326, opened 2021-06-26
Detect overflow should use state rect

AlexStrNik:
Since modifiers can change state, `detectOverflow` should use state's rect instead of rect returned by `getBoundingClientRect`


FezVrasta:
Thanks for the fix!

PR #1330, opened 2021-07-15
Fix typings for getBoundingClientRect

vutran:
`getBoundingClientRect` should be returning a `DOMRect`, not a `ClientRect`.

From lib.dom.d.ts - https://github.com/microsoft/TypeScript/blob/541e553163bb79fba64fdfd7c8a2c96f201c3b0a/lib/lib.dom.d.ts#L4922

MDN - https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -190,6 +190,6 @@ export type SideObject = {|
 export type Padding = number | $Shape<SideObject>;
 
 export type VirtualElement = {|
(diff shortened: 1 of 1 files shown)


FezVrasta:
Why this change? Both types are accepted.

PR #1342, opened 2021-08-11
Add an exports map

benmccann:
The usage of `process.env.NODE_ENV` is causing issues in svelte-repl and Vite and has been a pain point for other users (https://github.com/popperjs/popper-core/issues/933). `main`, `main:umd`, and `module` do not specify all the various ways to use the library. Right now there's nothing that points at `dist/esm/popper.js`. With an `exports` map we can better specify which version of the distribution should be used. Some tools like Vite want to automatically import one version on the server and another version in the browser. If you make the user manually point and the specific distribution it won't work. Instead we need to specify it in `package.json`

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,19 @@
     "/dist",
     "/lib"
   ],
+  "exports": {
+    ".": {
+      "node": {
+        "import": "./lib/index.js",
+        "default": "./dist/cjs/popper.js"
+      },
+      "browser": {
+        "import": "./dist/esm/popper.js",
+        "default": "./dist/umd/popper.js"
+      }
+    },
+    "./package.json": "./package.json"
+  },
   "sideEffects": false,
   "scripts": {
     "clean": "rimraf lib && rimraf dist && rimraf test/visual/dist",


FezVrasta:
Thanks 🙇

XhmikosR:
This is a breaking change...

nstuyvesant:
As @popperjs/core@2.10.0 is not compatible with SvelteKit 1.0.0-next.162's dev server, going to recommend sveltestrap@5.6.2 and bootstrap@5.1.0 change their package.json dependency on @popperjs/core to `~2.9.3` until this gets sorted out. Currently, they use `^2.9.2` and `^2.9.3` respectively.

FezVrasta:
I wonder what's the correct approach to go about this. Should the package.json map every single file manually?

FezVrasta:
> This is a breaking change...

I reverted this on 2.10.1

nstuyvesant:
Will test it out to see if SvelteKit is back in business.

king612:
Mine is not.  popperjs actually breaks svelte on server startup when using Sveltestrap (depends on popperjs).  

Please add "type": "module" to your package.json.  This fixes the problem.

atomiks:
We can't do that because it's a breaking change (which is why we reverted this). The easiest solution is to use `patch-package` and add that change yourself to `@popperjs/core` package. It's the only reasonable solution

benmccann:
It looks like this is fixed in v3. Any ETA on a v3 release?

PR #1356, opened 2021-09-02
fix: avoid using scrollHeight/scrollWidth when the position is fixed and when the offsetParent is the window

Ayc0:
Fixes https://github.com/popperjs/popper-core/issues/1323

Note: as master is broken at the moment, it isn't up to date with the latest master, but instead is based on top of f4a6550


FezVrasta:
Master should be good now, could you please rebase your branch on top of it?

PR #1358, opened 2021-09-02
fix: use other direction with end variation (#1302)

FezVrasta:
Same as #1302 but with fixed types and tests

cc @Ayc0

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,16 +36,22 @@ jobs:
 
   unit:
     runs-on: ubuntu-latest
+    container:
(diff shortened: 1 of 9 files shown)


Ayc0:
I was looking into how to fix the types and the tests after updating it with the new master
I don't know why the old tests weren't failing :/

FezVrasta:
CI on PRs wasn't working, now it does and it's correctly reporting failing tests

Ayc0:
I re-did the PR here: https://github.com/popperjs/popper-core/pull/1360 if you wanna have a look

PR #1360, opened 2021-09-02
fix: use other direction with end variation

Ayc0:
Reimplements https://github.com/popperjs/popper-core/pull/1302 but with the type fixed

## Description

When we use the placement `left`, the coordinates used aren't `top:0; left:0` anymore but `top:0; right:0`.
Same thing with the placement `top`, the coordinates used become `bottom:0; left:0`.

But using `bottom-end` and `top-end` also also use the coordinate `right:0` was end specifies that the popup should be stuck at the end of the the X axis, aka on the right. (and same with `left-end` and `right-end`, they should be stuck at the bottom).

An effective benefits of flipping the axis is that if the popup grows, the sides that are stuck won't move. So if someone uses `left-end`, the popup will be stuck on the right and on the bottom.

### DIFF
--- a/src/modifiers/computeStyles.test.js
+++ b/src/modifiers/computeStyles.test.js
@@ -7,6 +7,7 @@ it('computes the popper styles', () => {
   expect(
     mapToStyles({
       popper: document.createElement('div'),
+      variation: null,
       placement: 'bottom',
       popperRect: { x: 10, y: 10, width: 100, height: 100 },
       offsets: { x: 10, y: 5 },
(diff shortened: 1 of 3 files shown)


src/modifiers/computeStyles.test.js
```diff
@@ -7,6 +7,7 @@ it('computes the popper styles', () => {
   expect(
     mapToStyles({
       popper: document.createElement('div'),
+      variation: null,
```
Ayc0, line 10: Here I set `null`, but maybe we should set `start`.

What do you think?

FezVrasta: Why do you think it should be `start`?

Ayc0: Nevermind, I was thinking only about `left-start` and `left-end`, but there is also "left-center" aka just `left`

PR #1380, opened 2021-09-27
fix: remove scale logic in getBoundingClientRect

atomiks:
Fixes issue with elements of non-integer sizes (https://github.com/popperjs/popper-core/issues/1351#issuecomment-918015006)

The logic for the `scaleX` / `scaleY` is faulty as explained in the comment, I can't quickly think of a solution but positioning is broken for all elements that aren't an integer of width or height. I think it takes priority over the scaling based on the issues I've been receiving for Tippy


FezVrasta:
Thanks, too bad we need to revert that feature.

PR #1398, opened 2021-11-02
fix(computeStyles): adaptive option with fixed strategy on mobile

### LINKED ISSUE #1136: Fixed strategy works strange with Chrome and Safari navbar on mobiles
### CodeSandbox demo

https://codesandbox.io/s/tender-oskar-s94e6

video_2020-06-23_11-16-33

1. Open sandbox on mobile in Chrome
2. Scroll down when Chrome's navbar is hiding.
3. See that popup's position is changing.

Popup's position shouldn't change.

### What went wrong?

Poper overlaps with reference without any reason

atomiks:
Fixes #1136 

`document.documentElement.clientHeight` returns the height of the viewport including the bottom address bar, but fixed elements do not get positioned using that strategy. Tested on iOS 15 + iPhone 13 Simulator with XCode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "poster": "^0.0.9",
     "prettier": "^2.2.1",
     "pretty-quick": "^3.1.0",
-    "puppeteer": "^7.1.0",
+    "puppeteer": "^10.4.0",
     "replace-in-files-cli": "^1.0.0",
     "rollup": "^2.39.0",
     "rollup-plugin-flow-entry": "^0.3.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -106,7 +106,7 @@
     "poster": "^0.0.9",
     "prettier": "^2.2.1",
     "pretty-quick": "^3.1.0",
-    "puppeteer": "^7.1.0",
+    "puppeteer": "^10.4.0",
```
atomiks, line 109: This makes the installation of the repo work on M1 macs

PR #1401, opened 2021-11-02
fix(preventOverflow): altAxis + tether

### LINKED ISSUE #1365: preventOverflow tether working incorrectly with altAxis
Hi,

I am using Popper via Tippy with the `preventOverflow` modifier with the following options

```
{
    mainAxis: false,
    altAxis: true,
    altBoundary: true,
    boundary: scrollParentRef.current ?? document.body,
    padding: arrowSize / 2,
}
```

The goal is to prevent the link editing dialog to leave the screen

image

While I later found that `tether: false` ensures the desired behavior for me, I think I found a glitch in the `preventOverflow` modifier when `tether` is enabled.

The thing is that at the left side of the screen, the `tether` doesn't seem to work at all for me, even when enabled (the target is about 200px below the fold)

image

While from a certain point it allows the popper element to go deeper and deeper. It goes more towards the correct position, yet not completely correct.

image

The more to the right, the later the `tether` stops working:

image

Based on the documentation for `tether`, I would say that it should always go with the target element.

I tracked this down to the lines where `tetherMin` and `tetherMax` are evaluated, and it seems to me that they are evaluated using the `mainAxis` values:

https://github.com/popperjs/popper-core/blob/094c3115843cdfc207f70437318695a689488a2b/src/modifiers/preventOverflow.js#L150

While a bit later in the code the offset is adjusted for `altAxis` using these same values.

https://github.com/popperjs/popper-core/blob/094c3115843cdfc207f70437318695a689488a2b/src/modifiers/preventOverflow.js#L171

I didn't dig deeper as I am not actually using `tether` anymore, but with my limited knowledge, I would say the boundaries for the `altAxis` should be calculated from the alt axis offset.

I would be glad if someone more knowledgeable looks at this and verified my assumption.

atomiks:
I am sure the logic for the math can be more compressed/elegant because it feels kinda weird, especially regarding `isOriginSide`, but this is what I've come up with to fix the issue.

Fixes #1365 https://github.com/popperjs/popper-core/issues/1242#issuecomment-884651318

@panzarino


atomiks:
Also fixed in v3 https://github.com/popperjs/popper-core/pull/1416

PR #1405, opened 2021-11-06
Allow svg elements to be used as boundaries for `detectOverflow`

Andarist:
I've been trying to use `<svg/>` as a boundary because I'm implementing an svg-based "canvas" which gets panned around etc. Since we have panels on the sides I didn't want for popover to obscure them (that part can just be fixed with the `overflow:hidden` on the container of my `<svg/>`) and while looking into the options of the Popper I've noticed `boundary`. I have thought that this should improve the positioning of my popover and it did, once I've figured out that I need to use my `svg`'s parent element as the `boundary` because this `getClientRectFromMixedType` doesn't support non-HTML elements.

`getBoundingClientRect` is available on the SVG elements too so it makes sense, afaik, to support them here too.

I can look into adding tests once you confirm that there are no downsides to this and that this is something that you want to support.


atomiks:
Looks fine to me considering it's a tiny one line change, assuming it works with some tests of course 👍

FezVrasta:
I'm wondering if Element could be too broad? I'm away from my computer these days so I can't easily check, but I wonder if `HTMLElement | SVGElement` could be a safer choice

Andarist:
@FezVrasta this only depend on the `getBoundingClientRect` and this method is defined on the `Element` interface: https://drafts.csswg.org/cssom-view/#dom-element-getboundingclientrect

So I think this is the correct and simple check to be made here. Let me know what do you think

tests/functional/preventOverflow.test.js
```diff
@@ -131,3 +131,10 @@ it('should be inside scroller container', async () => {
 
   expect(await screenshot(page)).toMatchImageSnapshot();
 });
+
+it('should be able to use svg element as a boundary', async () => {
```
Andarist, line 135: **before**
prevent-overflow-svg-pre

**after**
prevent-overflow-svg-post

Andarist:
Ok, I've pushed out a test for this - let me know if this is enough :)

PR #1412, opened 2021-11-11
Refine `state.placement` type to `ComputedPlacement`

OliverJAsh:
As far as I understand, `state.placement` will never be `auto-*`.


OliverJAsh:
Was this closed because Popper is no longer maintained?

FezVrasta:
The v2 branch entered low maintenance mode and the new version is using TS

PR #1416, opened 2021-11-17
[Concept] Popper v3

atomiks:
Creating this PR now for feedback

TODO:

- [x] `auto` placements
- [x] change all tests to work with new API
- [x] platform-agnostic types(?)
- [x] add tests for new `size` modifier

<details>
  <summary>Read concept below</summary>

# Popper 3

Popper is a low-level, pure, and functional positioning engine for "poppers" — floating elements which are positioned relative to another element (such as tooltips, popovers, dropdowns, and more). It's platform-agnostic, meaning it can run on both web and native.

Popper is built on the concept of gradual adoption, so you add features as you need them. It's super modular and tree-shakeable, so anything you don't use will be removed from your bundles. The core is only 600 bytes when minified and compressed with brotli.

0.62 KB

## Basic positioning

At its most basic, Popper will position your floating element next to a given reference element on one of its four sides: top, right, bottom, left. Two variations to these placements are available, `-start` and `-end`, which will align it to the start or end of the reference element, rather than the center.

Popper’s most useful feature is the **ability to keep the floating element in view for the user as best as possible**, preventing overflow and clipping issues. The most common example is if the reference element (e.g. button) is near the window’s edge and the popover is wider than it. This will cause it to overflow and get cut off if we naively center it below using plain CSS. With CSS, you'd need to manually adjust the positioning each time you add a tooltip or popover, which quickly becomes cumbersome.

## Quick start

```js
import { position } from "@popperjs/dom";

function applyStyles({ x = 0, y = 0 }) {
  Object.assign(tooltip.style, {
    position: "absolute",
    top: `${y}px`,
    left: `${x}px`,
  });
}

applyStyles();
position(button, tooltip).then(applyStyles);
```

This will position the tooltip centered underneath the button. This is the most basic positioning behavior and does not include any of Popper's useful functionality yet.

Above, we import the `position` function from the DOM package to use it on the web. We apply the initial styles, then we call `position()` with the button element, the tooltip element, and the configuration object. It returns a promise to allow for platforms which have async measurement APIs, passing in the positioned `x` and `y` coords.

`position` is pure, as all it does is return `x` and `y` coordinates it calculates for us, which we then use to mutate our floating element's styles, which is entirely up to you to control.

## Modifiers

Popper's features are implemented as "modifiers", or pieces of code which modify the `x` and `y` coordinates in some way the base position, or return data to modify the styles.

The `shift` modifier is one of the most useful, as it prevents overflow (or clipping) issues by **shifting** the popper in view.

```js
import { position, shift } from "@popperjs/dom";

function applyStyles({ x = 0, y = 0 }) {
  Object.assign(tooltip.style, {
    position: "absolute",
    top: `${y}px`,
    left: `${x}px`,
  });
}

applyStyles();
position(button, tooltip, {
  modifiers: [shift()],
}).then(applyStyles);
```

Now, `x` and `y` are "smarter" because they've been modified by `shift`.

### Custom modifiers

Modifiers can be written by you to add custom functionality. Return the new `coords`, and optionally, a `data` object.

```js
const customModifier = (options) => ({
  name: "customModifier",
  fn(modifierArguments) {
    const { coords } = modifierArguments;

    return {
      ...cords,
      data: {
        // custom data
      },
    };
  },
});
```

</details>


src/dom.js
```diff
@@ -0,0 +1,72 @@
+// @flow
```
atomiks, line 1: This file should become its own package, if we are to support `react-native` too

src/modifiers/shift.js
```diff
@@ -0,0 +1,166 @@
...
+
+type Options = {
+  mainAxis: boolean,
+  altAxis: boolean,
+};
+
+export const shift = (options: Options = {}) => ({
+  name: 'shift',
+  async fn(modifierArguments: ModifierArguments) {
+    const { placement, coords } = modifierArguments;
```
atomiks, line 34: `fn` must now be async to access `detectOverflow`. This is due to being platform-agnostic which might only provide async methods to measure something. React Native has a bunch of these async, although getting e.g. the window dimensions is sync.

src/modifiers/shift.js
```diff
@@ -0,0 +1,166 @@
...
+    }) => number)
+  | number;
+
+type LimitShiftOptions = {
+  offset: LimitShiftOffset,
+  mainAxis: boolean,
+  altAxis: boolean,
+};
+
+export const limitShift = (options: LimitShiftOptions = {}) => ({
```
atomiks, line 92: This is `tether`, I feel like this name is more intuitive than it

src/modifiers/shift.js
```diff
@@ -0,0 +1,166 @@
...
+import getVariation from '../utils/getVariation';
+import getFreshSideObject from '../utils/getFreshSideObject';
+import { max as mathMax, min as mathMin } from '../utils/math';
+
+type Options = {
+  mainAxis: boolean,
+  altAxis: boolean,
+};
+
+export const shift = (options: Options = {}) => ({
```
atomiks, line 31: `preventOverflow` renamed to `shift`

src/position.js
```diff
@@ -0,0 +1,128 @@
...
+      diff,
+      data,
+      scheduleReset,
+      rects,
+      platform,
+      options: { placement },
+      elements: { reference, popper },
+    });
+
+    statefulCoords = nextCoords;
```
atomiks, line 109: Modifiers simply return the next coords once they've done their modifications logic. The coords they receive are the **current** ones, which previous modifiers have changed (not mutated though, it's all pure).

src/types.js
```diff
@@ -1,16 +1,40 @@
...
-import type { EventListenersModifier } from './modifiers/eventListeners';
-import type { ComputeStylesModifier } from './modifiers/computeStyles';
-import type { ArrowModifier } from './modifiers/arrow';
-import type { PreventOverflowModifier } from './modifiers/preventOverflow';
-import type { ApplyStylesModifier } from './modifiers/applyStyles';
+import type { Placement, Boundary, RootBoundary } from './enums';
+
+// TODO: platform-agnostic types?
+export type Reference = any;
+export type Popper = any;
```
atomiks, line 7: @FezVrasta

src/utils/computeCoords.js
```diff
@@ -2,57 +2,51 @@
...
-  Offsets,
-  ClientRectObject,
-} from '../types';
+import type { Rect, Coords, ClientRectObject } from '../types';
 import { top, right, bottom, left, start, end, type Placement } from '../enums';
 
-export default function computeOffsets({
+export default function computeCoords({
   reference,
-  element,
```
atomiks, line 15: I kept getting confused by the term `offsets`, so I've changed it to `coords`. imo, more intuitive to represent the `x` and `y` position

tests/visual/basic.html
```diff
@@ -2,33 +2,124 @@
...
+  import { hide } from './dist/modifiers/hide.js';
+  import { platformLite } from './dist/dom.js';
+
+  import listScrollParents from './dist/dom-utils/listScrollParents.js';
+
+  const referenceElement = document.querySelector('#reference');
+  const popperElement = document.querySelector('#popper');
+  const arrowElement = document.querySelector('#arrow');
+
+  async function update() {
```
atomiks, line 83: Demonstrating the new API here

src/modifiers/flip.js
```diff
@@ -115,63 +140,65 @@ function flip({ state, options, name }: ModifierArguments<Options>) {
...
+            (a, b) =>
+              a.overflowChecks
+                .filter((overflow) => overflow > 0)
+                .reduce((acc, overflow) => acc + overflow, 0) -
+              b.overflowChecks
+                .filter((overflow) => overflow > 0)
+                .reduce((acc, overflow) => acc + overflow, 0)
+          )[0].placement;
+
+        scheduleReset({ placement: bestFittingPlacement });
```
atomiks, line 190: This function will schedule a reset for us once the modifier has returned (modifier lifecycle is reset, like `state.reset = true` in v2

src/modifiers/hide.js
```diff
@@ -9,68 +9,51 @@ import type {
...
+function getSideOffsets(overflow: SideObject, rect: Rect): SideObject {
   return {
-    top: overflow.top - rect.height - preventedOffsets.y,
-    right: overflow.right - rect.width + preventedOffsets.x,
-    bottom: overflow.bottom - rect.height + preventedOffsets.y,
-    left: overflow.left - rect.width - preventedOffsets.x,
+    top: overflow.top - rect.height,
+    right: overflow.right - rect.width,
+    bottom: overflow.bottom - rect.height,
+    left: overflow.left - rect.width,
```
atomiks, line 17: Needing to check for data from other modifiers is reduced everywhere. Modifiers are easier and smaller to write.

src/dom.js
```diff
@@ -0,0 +1,72 @@
...
+
+// Considers any scenario in which the popper may be nested within any number of
+// clipping parents.
+export const getClippingClientRect: GetClippingClientRect = (args) =>
+  Promise.resolve(getClippingClientRectUtil(args));
+
+export const platformLite: Platform = {
+  getRects: getRectsLite,
+  getClippingClientRect: getClippingClientRectLite,
+};
```
atomiks, line 67: A concept of a "lite" version of the DOM platform which performs only basic checks, reducing bundle size. I have not yet considered things like adding extra checks for things like `transform` CSS set on `<body />`. We could just support every edge case imaginable in the heavier `platform` object.

tests/visual/basic.html
```diff
@@ -2,33 +2,124 @@
...
+    const { x, y, data } = await position(referenceElement, popperElement, {
+      platform: platformLite,
+      placement: 'top-end',
+      modifiers: [
+        offset([0, 10]),
+        flip({
+          // altAxis: true,
+        }),
+        shift(),
+      ],
```
atomiks, line 93: I'm not sure about needing to auto-order modifiers. Sometimes, you might want a modifier to be placed before *OR* after another, depending on the behavior you want. I noticed this with `flip` before/after `shift` with this new API... some like `offset` should always be run before the others though.

src/modifiers/flip.js
atomiks: This could alternatively be done in the `return`, smth like `return { reset: { placement: ... } }`

tests/visual/modifiers/arrow/hide.html
```diff
@@ -37,55 +37,53 @@
...
+    Object.assign(popperArrow.style, {
+      position: 'absolute',
+      left: `${data.arrow.x}px`,
+      top: `${data.arrow.y}px`,
+      visibility: data.arrow.centerOffset !== 0 ? 'hidden' : '',
+    });
+  }
+
+  Object.assign(popper.style, {
+    position: 'absolute',
```
FezVrasta, line 81: Shouldn't the `position` be provided by Popper depending by the defined strategy?

atomiks: Well the assumption is since they pass the strategy they will know what to use

FezVrasta: But if one wants to dynamically switch between strategies they will need to remember to also update this

src/modifiers/shift.js
FezVrasta: Can we support both async and not? We can `await` for a non-async function anyways, this way people that don't need async capabilities don't need to worry about that.

atomiks: Yeah both are supported

atomiks:
@FezVrasta so I think the v3 overhaul is mostly feature-complete, I need to test it out on React Native to see how the calls work, maybe we could publish an `3.0.0-alpha.0` with an updated README (also, we need to export the `platform` from `dom.js` in the main package for now to use it on the web)

atomiks:
WIP: converted to a monorepo (using Yarn workspaces) with `@popperjs/core`, `@popperjs/dom`, and `@popperjs/react-native` packages. The Rollup config to output typings for Flow & TS needs to be implemented.

FezVrasta:
I can take care of the remaining repo setup, I'll push some changes in a bit.

FezVrasta:
I completed the monorepo migration and fixed unit tests for the core package. I also upgraded yarn to 3.x and enabled the zero-installs system

atomiks:
The yarn cache files should be in `.gitignore`?

package.json
```diff
@@ -9,5 +9,8 @@
     "trailingComma": "es5",
     "singleQuote": true,
     "proseWrap": "always"
+  },
+  "dependencies": {
+    "flow-bin": "^0.165.0"
   }
```
atomiks: It looks like this needs to be added for VSCode's Flow extension to find the correct version.

atomiks: Other option might be to require users to install Flow globally and configure their extension not to use the `npmPackaged` version.

atomiks:
`dist` files naming:

- `@popperjs/core`: `popper-core.js` (UMD) or `popper-core.mjs` (ESM)
- `@popperjs/dom`: `popper-dom.js` (UMD) or `popper-dom.mjs` (ESM)
- `@popperjs/react-native`: `popper-react-native.cjs` (CJS) or `popper-react-native.mjs` (ESM)

FezVrasta:
I think having it as a regular dependency will ease the usage.

I also think it will make sense to bring react-popper into the monorepo and maybe call it @popperjs/react-dom

atomiks:
A couple of issues:

1. `@popperjs/core` needs to be added as a dependency to `@popperjs/dom` for publishing on npm, but this causes issues with `yarn install` on the repo since it's symlinked
2. The functional tests use the compiled `dist/` `@popperjs/core` files, so it needs to be built before running them.

FezVrasta:
I added @popperjs/core to the @popperjs/dom dependencies, it seems to work correctly. I also switched to Yarn PnP to speed everything up and ensure all the dependencies are in the right place.

FezVrasta:
I'm trying to upgrade the visual tests to Playwright but I have issues with it and PnP 😒

atomiks:
I tried to use Playwright with Docker but wasn't successful. Visual mismatches between dev and CI are annoying so Docker solves that, I use it on Tippy repo w/ Puppeteer but it doesn't seem to work with M1 macs :\ Wonder if we could generate the snapshots on CI itself using Docker

FezVrasta:
Yes that's what we do internally, I'll make it work

atomiks:
Wonder if we should have a `@popperjs/utils` package to export all the utils which have a stable API. This becomes very useful since you're applying styles, getting scrollParents etc. yourself where necessary.

FezVrasta:
Maybe it's enough to export them from core? We could consider as stable only the ones exposed by the package.json

atomiks:
(disabled PnP for now to continue working)

nandorojo:
Coming in a bit late to the conversation here and not super familiar with popper internals, so sorry if this is a bit naïve, but on the React Native side, would it be possible to use the synchronous `measure` function from reanimated?

atomiks:
Yes sync to async is always possible, but not the other way around. Just do `Promise.resolve(data)`.

package.json
```diff
@@ -15,11 +15,13 @@
     "extends": "./.config/eslint.config"
   },
   "scripts": {
-    "test:types": "yarn flow",
-    "test:lint": "yarn eslint ."
+    "test:types": "flow",
+    "test:lint": "eslint .",
+    "dev": "concurrently 'yarn workspace @popperjs/core dev' 'yarn workspace @popperjs/dom dev'"
```
atomiks, line 20: When editing `@popperjs/core` source files, Rollup can refresh when doing dev on the `/dom` package.

atomiks:
@FezVrasta if we can alter the GitHub actions files to get the tests passing here + create typings exports, we could release an `alpha.0` of the packages

packages/dom/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "test:functional": "jest tests/functional",
+    "test:unit": "jest src/**/*.test.js",
+    "test": "test:functional && test:unit",
+    "build": "NODE_ENV=build rollup -c"
+  },
+  "author": "",
+  "license": "MIT",
+  "dependencies": {
+    "@popperjs/core": "workspace:^"
+  },
```
atomiks, line 25: @FezVrasta how does this work when publishing the package? Do we need a script to replace this with an actual version before publish?

packages/react-native/src/index.js
```diff
@@ -113,44 +132,37 @@ export const usePopper = ({
...
 
-      computePosition(referenceRef.current, popperRef.current, {
+      computePosition(reference, popper, {
         platform,
         placement,
         modifiers,
       }).then(setData);
     });
-  }, [platform, scrollOffsets]);
+  }, [platform, reference, popper, modifiers, placement]);
```
atomiks: `modifiers` inside the dep array here means the user will always need to `memo` their modifier calls (or place it outside the component

inlining this inside `usePopper`:

```js
modifiers: [shift(), flip()]
```

causes an infinite loop and since it's generating a new object each time you can't compare the objects deeply

atomiks:
Hi, so I am going to close this out because I believe this change in direction is too big of a breaking change for Popper (and any packages which use it as a dependency).

I am instead opting to create a new project built on this foundation I've created, which will be more ambitious but provides similar primitives for positioning, with additional ones for crafting actual UI elements (like Tippy, but lower level). You'll be able to just use the positioning package if needed. 

While leveraging the current mindshare and popularity of an existing project is worthwhile, I have noticed that many authors create a new library or project entirely if the architectural change is too big, so I am opting to go this direction. 

Considering most of the architecture here is done it will be done quite shortly and I'll be updating on it on Twitter: https://twitter.com/atomiksdev

PR #1418, opened 2021-11-23
fix: round after math

### LINKED ISSUE #1300: Yet another subpixel rounding issue [Safari 14.0.3]
Hi, I saw that there are many issues about subpixel rendering and that a lot of impressive work was done to cover all the edge cases, so I'm not sure if this is already a known problem.

On Safari, both on iphone and on mac with hidpi display the popper element gets misaligned by half a pixel when the reference is positioned on a decimal (e.g. 104.5px) x coordinate.

Screenshot 2021-05-04 at 14 36 18

### CodeSandbox demo

https://codesandbox.io/s/wizardly-shadow-j0lui

In this sandbox the problem is reproduced using a minimal popper-lite setup and hardcoding the position of the reference to a decimal value

open the codesandbox on this environment:

Mac OS 11.2.3
Safari 14.0.3
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.3 Safari/605.1.15"
devicePixelRatio = 2

the left borders of the reference and the popper should be perfectly aligned and look like a continuous line

### What went wrong?

The line is broken because the popper is misaligned in relation to the reference
### Any other comments?

It seems to work correctly if the values are not rounded, my understanding is that rounding or not depends on a series of conditions (dpi, user agent, gpu, etc...) and maybe this is a missed corner case where rounding should be disabled?

cc. @sylvesteraswin

### LINKED ISSUE #1410: Popper element is not aligned to trigger, off by 1px
### CodeSandbox demo

https://stackblitz.com/edit/react-nyp2yj?file=src%2FApp.js

Clonable repo: https://github.com/OliverJAsh/popper-subpixel-issue

When we view this test case at odd viewport sizes e.g. 327px wide in this example, the popper element is not aligned to the trigger. It is off by ~1 px.

image

Zoomed in slightly:

image

Westbrook:
Fixes #1300
Fixes #1410
Refs #1400

Rounding was previously happening to _ONE_ of the measured inputs and then other inputs were not, so the output wouldn't actually be rounded. This moves the rounding to _only_ happen at the end of the style calculations so that it ensure the styles sent the the client are rounded.

### DIFF
--- a/src/modifiers/__snapshots__/computeStyles.test.js.snap
+++ b/src/modifiers/__snapshots__/computeStyles.test.js.snap
@@ -102,13 +102,24 @@ Object {
   "bottom": "auto",
   "left": "auto",
   "position": "absolute",
+  "right": "-150.5px",
+  "top": "5px",
+  "transform": "",
+}
+`;
+
+exports[`computes the popper styles 10`] = `
+Object {
+  "bottom": "auto",
+  "left": "auto",
+  "position": "absolute",
   "right": "-112px",
   "top": "8px",
   "transform": "",
 }
 `;
 
-exports[`computes the popper styles 10`] = `
+exports[`computes the popper styles 11`] = `
 Object {
   "bottom": "auto",
   "left": "auto",
--- a/src/modifiers/computeStyles.js
+++ b/src/modifiers/computeStyles.js
@@ -79,12 +79,12 @@ export function mapToStyles({
   roundOffsets: boolean | RoundOffsets,
   isFixed: boolean,
 }) {
-  let { x = 0, y = 0 } =
-    roundOffsets === true
-      ? roundOffsetsByDPR(offsets)
-      : typeof roundOffsets === 'function'
-      ? roundOffsets(offsets)
-      : offsets;
+  let { x = 0, y = 0 } = offsets;
+
+  ({ x, y } =
+    typeof roundOffsets === 'function'
+      ? roundOffsets({ x, y })
+      : { x, y });
 
   const hasX = offsets.hasOwnProperty('x');
   const hasY = offsets.hasOwnProperty('y');
@@ -147,6 +147,11 @@ export function mapToStyles({
     position,
     ...(adaptive && unsetSides),
   };
+
+  ({ x, y } =
+    roundOffsets === true
+      ? roundOffsetsByDPR({ x, y })
+      : { x, y });
 
   if (gpuAcceleration) {
     return {
--- a/src/modifiers/computeStyles.test.js
+++ b/src/modifiers/computeStyles.test.js
@@ -122,6 +122,22 @@ it('computes the popper styles', () => {
     })
   ).toMatchSnapshot();
 
+
+  expect(
+    mapToStyles({
+      popper: document.createElement('div'),
+      variation: 'end',
+      placement: 'bottom',
+      popperRect: { x: 108.34375, y: 8, width: 140.296875, height: 38 },
+      offsets: { x: 10, y: 5 },
+      position: 'absolute',
+      gpuAcceleration: false,
+      adaptive: true,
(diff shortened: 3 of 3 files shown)


atomiks:
Does it also help this https://github.com/popperjs/popper-core/issues/1410?

Westbrook:
It definitely looks a lot better. "Solved" is tricky when _somethings_ can be subpixel and some things can't be. While zooming in continues to throw a big fat wrench, non-zoomed this appears to align things as expect in the case of #1410 as well.

atomiks:
> While zooming in continues to throw a big fat wrench

What about in combo with the DPR PR too https://github.com/popperjs/popper-core/pull/1400

Westbrook:
DPR doesn't seem to change while zooming. I'm not sure there is any way to surface zoom in the client to account for it either. I'll look into that.

Westbrook:
My mistake, `window.devicePixelRatio` does seem to follow the zoom, the issue is when during the 110% zoom where the math still ends up with a non-representable pixel fraction. However, even in that case it looks much better than what was being previously delivered. All other zooms seem to be good to go in the provided demo.

Westbrook:
Any thoughts on this test failure? I’m not sure how it was considered passing before based on the code I’m seeing. Happy to update the baseline with some agreement there. This update helps my project a lot and I’d love for it to come without having to patch the package.

atomiks:
Regarding that test, a difference of 4px seems kinda odd (I'd expect 1px max), do you have a replication of the differences?

FezVrasta:
The test asserts that 100px width + 10.3px offset + 2px should result into 112px after it gets rounded. 108 seems not to be correct, it looks like you are subtracting 2 rather than adding them?

Westbrook:
I stepped through this some more and better understand what is going on, but I'm not sure what to say is the right answer.

The `roundOffsets` method here is:

```
roundOffsets: ({ x, y }) => ({
        x: Math.round(x + 2),
        y: Math.round(y + 2),
}),
```

If you round at the beginning, you'll end up with `-112` because you add 2 to the input (`10.3` to `12`), and if you round at the end you get `-108` because you round the output (`-110.3` to `-108`).

What is the _meaning_ of this test? `// customize roundOffsets impls` doesn't really explain why you are making a custom `roundOffset` to tell what is _supposed_ to be happening here. However, it does imply that my alteration would be a breaking change for some, regardless of it being more or less correct output. Rounding at the beginning and at the end would be less accurate, but would allow for this to both function as before and ensure all number sent to the client are rounded correctly. OR, we could allow for all custom `roundOffset` code to be run _only_ at the beginning, as that likely means the user is getting what they'd prefer here already?

Westbrook:
Or maybe use the `roundOffsets` option at the beginning of the process as available, but not as a replacement for `roundOffsetsByDPR`, which should likely be renamed to `roundPlacementByDPR` and run 100% of the time at the end of `mapToStyles`?

Westbrook:
Split the two parts of rounding outlined in https://github.com/popperjs/popper-core/pull/1418#issuecomment-981094250 to before and after the computations and added a new test (leveraging measurements from https://stackblitz.com/edit/react-nyp2yj?file=src%2FApp.js) that failed to deliver DPR correct computations when using the previous method of rounding (which is the basis for #1410).

Westbrook:
Any insight on the Flow issue would be most helpful. That same command doesn't fail locally and I've never used Flow outside of this project.

Westbrook:
Oh, I think I got it now. My branch was out of date 😅 Hopefully this is the one!

FezVrasta:
Thanks! Merging.

Westbrook:
Can we get a release on this to hold people over until they're able to update to Floating UI?

PR #1427, opened 2021-12-09
fix(dom): transformed body

### LINKED ISSUE #1375: Popper appears in the wrong place if the body element has the transform css property set
### CodeSandbox demo
https://codesandbox.io/s/elastic-sun-r5plx?file=/src/index.css

1. Set the transform css property on the body element.

I'd expect the popper to appear correctly positioned next to the target element.

### What went wrong?

The popper appears offset from the element, both horizontally and vertically, see the CodeSandbox demo above.

### Any other comments?

None

atomiks:
Floating UI's tests are not as extensive as Popper's (yet) since they were rewritten. Also would be good to add tests for both Webkit and Firefox.

Fixes #1375


FezVrasta:
We talked about it in pvt but just to share it with others, I think a Storybook + Playwright would be a great solution for us. 

https://storybook.js.org/addons/storybook-addon-playwright

I can spend some time this weekend to try to implement a POC.

PR #1437, opened 2021-12-11
perf(core): add pure annotation for allPlacements const

webistomin:
I just discovered the library is not fully tree-shakeable for some reason, so I added a `/* #__PURE__ */` comment.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8217,6 +8217,15 @@
         "node": ">=8"
       }
     },
+    "node_modules/babel-plugin-annotate-pure-calls": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-annotate-pure-calls/-/babel-plugin-annotate-pure-calls-0.4.0.tgz",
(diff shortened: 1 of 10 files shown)


atomiks:
I think there's a Babel plugin that does this for us automatically? Which parts weren't tree-shakeable? I know CRA has problems but Parcel/Rollup/etc. work 🧐

webistomin:
I use [agadoo](https://github.com/Rich-Harris/agadoo) for such purposes. As far as I know it uses rollup under the hood. 

I have never used this babel plugin, but I think it should help.

Before:

```
const basePlacements = ['top', 'right', 'bottom', 'left'];
const allPlacements = basePlacements.reduce((acc, basePlacement) => acc.concat(basePlacement, basePlacement + "-start", basePlacement + "-end"), []);

Failed to tree-shake dist\floating-ui.core.esm.js
```

After:

```
Success! dist\floating-ui.core.esm.js is fully tree-shakeable
```

webistomin:
The plugin also adds annotation in the same place as me.😉

 Is it better to use Babel?

atomiks:
Having a plugin do it automatically for us is better, since we might add more code that needs it 👍

webistomin:
Done

PR #1443, opened 2021-12-13
use `===`  replace isSameNode

berber1016:
The ` isSameNode()` method of the Node interface is a legacy, use `===` replace ` isSameNode()` method


atomiks:
Thanks!

PR #1452, opened 2021-12-15
fix: arrow element TS definition

gynekolog:
Arrow element has type HTMLElement | undefined. That cause a TS error with const `arrowRef = useRef(null)`.
If you use useRef without init value (`arrowRef = useRef()`) you got TS error on the target element.

image

image

**Solution**
Change arrow element type to `HTMLElement | null`.

reproduction: https://codesandbox.io/s/floating-ui-react-dom-visual-test-05xpu?file=/src/App.tsx

**Notes**
You can see in the example that there are other errors. It copy-pasted version of the https://github.com/floating-ui/floating-ui/blob/master/packages/react-dom/test/visual/index.js with annotated `arrowRef`. This pull request takes care only of the arrow element, not other errors.

### DIFF
--- a/packages/react-dom/index.test-d.ts
+++ b/packages/react-dom/index.test-d.ts
@@ -3,7 +3,7 @@ import {useFloating, shift, arrow} from '.';
 
 App;
 function App() {
-  const arrowRef = useRef();
+  const arrowRef = useRef(null);
   useFloating();
   const {reference, floating, update} = useFloating({
     placement: 'right',
--- a/packages/react-dom/index.test-d.tsx
+++ b/packages/react-dom/index.test-d.tsx
@@ -3,12 +3,11 @@ import {useFloating, shift, arrow} from '.';
 
 App;
 function App() {
-  const arrowRef = useRef();
+  const arrowRef = useRef(null);
   useFloating();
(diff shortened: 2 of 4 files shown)


packages/react-dom/index.test-d.ts
```diff
@@ -3,7 +3,7 @@ import {useFloating, shift, arrow} from '.';
 
 App;
 function App() {
-  const arrowRef = useRef();
+  const arrowRef = useRef(null);
```
atomiks, line 6: should add the arrow JSX `ref={arrowRef}` to test that as well

gynekolog: @atomiks I want but I'm not sure how can do it. Will you be fine with renaming `index.test-d.ts` to `index.test-d.tsx` and returning the arrow element?:
```tsx
function App() {
  const arrowRef = useRef(null);
  ...
  return (<div ref={arrowRef} />);
}
```

atomiks: Yep that works! as long as `tsc` command typechecks `.tsx`

atomiks (commented):
React Native needs the same changes

atomiks:
Thanks:)

PR #1468, opened 2021-12-24
TypeScript improvements

JuanM04:
- Made some improvements in `core/src/types.ts`
- Re-exported `Platform`, `Placement`, `Strategy` and `Middleware` types (the ones used as arguments of `computePosition`) in every package for easier access

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -1,57 +1,41 @@
+export type Alignment = 'start' | 'end';
 export type BasePlacement = 'top' | 'right' | 'bottom' | 'left';
-export type AlignedPlacement =
-  | 'top-start'
-  | 'top-end'
-  | 'right-start'
-  | 'right-end'
-  | 'bottom-start'
-  | 'bottom-end'
-  | 'left-start'
-  | 'left-end';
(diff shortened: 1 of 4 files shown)


atomiks (commented):
Thanks! What do you think about exporting `Options` from each middleware, is it necessary? Currently you need to do (for instance):

```ts
import type {Options} from '@floating-ui/core/src/middleware/flip';
```

Should as many types as possible be exported if it's part of the public API?

JuanM04:
> What do you think about exporting Options from each middleware, is it necessary?

I think it won't be needed. If you want to eliminate that import, you can create a `FlipFallbackStrategy`. Otherwise for those, in my opinion, few that want a typed `Options`, can use `Parameters<typeof flip>[0]`

> Should as many types as possible be exported if it's part of the public API?

It depends of how the developers uses the library. For example:

```tsx
const Tooltip = ({ placement }: { placement: Placement, strategy: Strategy }) => {
  const { /* ... */ } = useFloating({ placement, strategy })

}

```

atomiks:
> Parameters<typeof flip>[0]

TIL

This works for now, thanks

PR #1469, opened 2021-12-25
fix(react): deepEqual middleware

atomiks:
In order to allow Fast Refresh to work when changing middleware and a good dev experience in general, we need a `deepEqual` function, but it also needs to handle functions as options. Most implementations compare functions by reference; I think we can get away with comparing them by their string values as long as the options passed in to the function are always comparable by returning them on the object (so the closure can be made fresh)

cc: @kubajastrz


packages/react-dom/src/index.ts
```diff
@@ -54,20 +54,28 @@ export function useFloating({
...
-  const latestMiddleware = useLatestRef(middleware);
+  const middlewareRef = useRef(middleware);
+  if (
+    !deepEqual(
+      middlewareRef.current?.map(({options}) => options),
+      middleware?.map(({options}) => options)
+    )
+  ) {
+    middlewareRef.current = middleware;
+  }
```
KubaJastrz, line 65: Don't run side effects in the render function. It can have unexpected results in concurrent mode.

That's what useEffect is for.

KubaJastrz: If you want to compute middleware on the fly in every render, then you can change the ref to a regular variable.

atomiks: I adapted it from https://www.benmvp.com/blog/object-array-dependencies-react-useEffect-hook/#option-4---do-it-yourself but yeah it should probably be done in render

It can't be a regular variable as the value needs to stay consistent between renders if the middleware hasn't changed

atomiks: Writing to a ref inside an effect unfortunately causes an issue where the value can be stale due to the way the refs get assigned I think 🙄 that's the issue I had when writing the `inline` middleware with the current implementation, it has to be written inside the render.

atomiks: It looks like this is "safe" in CM because it doesn't matter what happens when the render() phase is called twice without being cleaned up. https://twitter.com/mjackson/status/1314599887778922497

KubaJastrz: How about using regular state then? Using refs in effect array feels dirty to me 😅

atomiks: state would reduce perf by causing two renders unnecessarily I think

PR #1476, opened 2021-12-28
docs: fetch latest npm versions for CDN links

atomiks:
As Rolling Versions doesn't add a commit to update the versions locally, we now need to fetch it from NPM


FezVrasta:
LGTM

Issue #1493, opened 2022-01-05
[Core] Arrow middleware does not point properly if toggle button is small


megawebmaster:
Hi!

When working on my own Tooltip / Popover I decided to use Floating-UI thanks to it's great API and ease of use. Everything was going smoothly until I found that when I use `arrow` middleware and the button I use to trigger the tooltip is small then the content and the arrow is not properly aligned.

You can see the reproduction here: https://codesandbox.io/s/floating-ui-dom-arrow-positioning-issue-pgf2g

Here's a screenshot for you:
![image](https://user-images.githubusercontent.com/301700/148194294-b8a1e198-6530-4a1e-b154-68495a3cc272.png)

The library, when using `start` or `end` placement, tries to align the container to the start of the trigger which is not what a user wants if they have an arrow (as you can see in my reproduction).

Is this behavior expected? 

I'd love to work on a PR to fix it, but I don't want to spend time if this is intended behavior. I would expect one of these behaviors:

1. Arrow is aligned to the start of the trigger:
   ![image](https://user-images.githubusercontent.com/301700/148194466-6b72e7e9-ce7c-430a-817e-3e602d81542a.png)
2. Arrow is aligned to the middle of the trigger:
   ![image](https://user-images.githubusercontent.com/301700/148194524-d22c8adf-323a-4ca0-b279-749689b0865d.png)

atomiks:
It's caused by the `TIP_EDGE_MARGIN`, you've given it a value of 16, which means it's at a minimum, 16px from the left edge at all times, resulting in the issue. Setting it to `0` fixes it (or a better value to handle the rounded corners).

atomiks:
Something like this seems to work decently:

```js
    offset(({ reference, placement }) => ({
      mainAxis: TIP_WIDTH,
      crossAxis:
        reference.width / 2 < TIP_SIZE
          ? placement.endsWith("start")
            ? -TIP_WIDTH
            : TIP_WIDTH
          : 0
    })),
```

https://codesandbox.io/s/floating-ui-dom-arrow-positioning-issue-forked-hzg0f?file=/src/index.js

You'd check `TIP_HEIGHT` / `reference.height` for `left/right` placements if needed, too

megawebmaster:
Well, your solution "kinda" works - it's still not really good for right/left cases. Anyway I feel a bit weird if I have to add offset for arrow padding - I'd expect the middleware to handle this case, so I can reason better about how I expect the floating item to appear.

[...]

PR #1496, opened 2022-01-05
Adjust floating element position if arrow points too far

megawebmaster:
Hi!

This PR tries to address #1493 when a reference element is narrow causing the arrow pointer to be misplaced.

What happens here:

1. We check if padding is used
2. We check if the floating element is rendering at the edges of reference (where we use `min` or `max` as an offset)
3. If we do: we calculate how much we need to move the whole floating element, so that arrow points at the center of the reference element
4. We add that adjustment to the correct axis.

Let me know what you think!


packages/core/src/middleware/arrow.ts
```diff
@@ -53,14 +55,23 @@ export const arrow = (options: Options): Middleware => ({
...
+    // Make sure that arrow points at the reference
+    const alignmentPadding = baseAlignment === 'start' ? paddingObject[minProp] : paddingObject[maxProp];
+    const alignmentOffset = alignmentPadding > 0 && center !== offset ?
+      center < min
+        ? min - center
+        : max - center
+      : 0;
+
     return {
+      [axis]: coords[axis] - alignmentOffset,
```
atomiks, line 74: This looks pretty neat and I like that it only touches one middleware!

Though, I saw you mentioned

> I'm not sure how to handle the "shift" case though as clicking at the edges of the screen can still render out of the viewport...

Since `arrow` has to go after `shift`, this won't let `shift` do its adjustment based on this change.

atomiks: To solve this, you could use `reset` + `skip` property technique like the other middleware do to make `shift` re-run its logic for this change. You can take a look at how other middleware like `size` do this

Also, tests for this would be great 😉

megawebmaster: Thanks for the review: I wanted to hear your initial thoughts as what else is missing and now I know how to proceed :+1: I'll take a look at `size` to adjust the code and (of course) I'll add tests :muscle:

atomiks: There is another problem though: since `flip` skips itself once it has requested a reset, and is placed before `arrow` too, any flipping which checks `crossAxis` will not work once you've reset the lifecycle inside `arrow` here. 

That means we'd probably need a change inside `computePosition` somewhere to make `reset` only skip the middleware that requested a reset

megawebmaster: @atomiks - I've been digging deeper into how it behaves and I realized there is no API that forces the whole middleware list to run. `reset` allows only for the current middleware to re-run with new values.

I believe that we don't need to handle the `shift` case anymore: we can just add another `shift` after the arrow too. We could add it to the tutorial (and docs), so the user is aware of it. I tested this case and works as one would expect :muscle: 

One other solution is to add an option to `arrow` middleware that will run `shift` as well behind the scene, but I'm not a fan of it as it makes the API more complex and hard to understand.

atomiks: > I've been digging deeper into how it behaves and I realized there is no API that forces the whole middleware list to run.

A `reset` will make the lifecycle restart (so any middleware run prior will re-run their logic _unless_ they skip themsleves by returning `data: {skip: true}`) so `shift` should work by doing this since it doesn't do any skipping. But other middleware like `flip` have this issue.

It's nice that `shift` can be added again, although the duplication is kind of weird and I think it's not necessary with `reset`. It also may not solve the `flip` issue? 

I think there is something missing with the `reset` behavior to make it more adaptable to this type of scenario...

megawebmaster: I might be out of place here, but `computePosition` moves back by 1 (so that current middleware re-runs): https://github.com/floating-ui/floating-ui/blob/master/packages/core/src/computePosition.ts#L106

This is why I couldn't make it work even though I tried really hard. In the end I believe re-running all items could work, but I don't really want to overcomplicate the PR (that will be much more complex with all tests as I needed to add typing for middleware data for tests to not cry out loud :wink:)

Take a look and let me know what you think!

packages/core/src/types.ts
```diff
@@ -92,21 +108,21 @@ export type ComputePosition = (
...
-    };
+    data: TData;
     reset: true | {placement?: Placement; rects?: true | ElementRects};
   }
 >;
 
-export type Middleware = {
+export type Middleware<TData = {
+  [key: string]: any;
+}> = {
```
megawebmaster, line 120: This is the default if you have no data in your middleware. It also keeps backwards compatibility (as this is the same that was before) and adds option to make the type more precise (i.e. so tests will pass).

packages/core/src/middleware/arrow.ts
atomiks: `i = -1` sets the loop index back to the beginning, so it restarts the loop. What you're referring to would be `i = i - 1` I think.

megawebmaster: Oh my, you're right and I must have been tired... I read it `i -= 1` 😆 

Anyway I wasn't able to run it in a way that will make `shift` and `arrow` adjust the position at the same time as `arrow` needs to adjust the position according to the padding, something that `shift` already fixed. I have no idea how to learn whether other middleware applied any changes in the second run. Maybe you have something in mind?

atomiks: Yeah as I mentioned before, we need to do this:

> That means we'd probably need a change inside computePosition somewhere to make reset only skip the middleware that requested a reset

We could potentially make "skip" a core feature and do this behavior, then everything should work correctly.

Btw, your tests should be the functional e2e using visual snapshots, not manual comparisons of the coords

atomiks:
Without considering `shift()`/`flip()` issues, there are still some issues with this implementation (seen here):

https://user-images.githubusercontent.com/22450188/148974507-80086a4b-e1ca-4096-9e99-013c3507e1a8.mov


I'm not sure how worth it is to add this logic when imo the `offset.crossAxis` technique I mentioned in the issue works well, is fully customizable and has no edge case issues to handle.

megawebmaster:
Just for your information: I still have this in mind, I just haven't got enough time to push it forward :+1:

atomiks:
With the latest master the flip/shift thing shouldn't be an issue anymore. Just the issue in the video I posted is remaining I guess 😄

megawebmaster:
@atomiks - I've rebased and checked the edge cases, added functional tests (with snapshots). This PR kinda depends on https://github.com/floating-ui/floating-ui/pull/1657 (required Linux snapshots to make sure things still work as expected).

atomiks:
I haven't forgotten about this, I will try to review it soon

atomiks:
`-start` and `-end` alignments should be respected even though the arrow can't be centered here. It ends up not aligning with the edge of the reference element:

https://user-images.githubusercontent.com/22450188/168476049-95abb82e-fff1-4496-b807-b6de8d78a8ec.mov

megawebmaster:
@atomiks - updated! I didn't know this was a requirement, but makes a lot of sense. All tests seem to pass already, so everything should be good now :muscle:

atomiks:
tbh I can sort of see that behavior sometimes being useful? Maybe as an option? 

I will review this again now 😄

megawebmaster:
I believe that too many options will make library users confused, so I'd stick with simpler solution and add the option if the need arises :+1:

atomiks:
> I believe that too many options will make library users confused, so I'd stick with simpler solution and add the option if the need arises 👍

True! There's one edge case remaining which is the issue I mentioned before about `shift` / `flip`:

`arrow()` has to go after `shift()` so the transformations it does to the `x` and `y` coords means it will overflow the edge despite `shift` being in use. This is the only way the arrow can stay centered to the reference while also respecting the padding, but I think `shift` should be respected instead in this case.

The manual `offset` technique does not suffer from this problem - this one would need a `reset` (prone to infinite loops) to somehow handle this case

Screen Shot 2022-05-17 at 9 33 50 pm

megawebmaster:
@atomiks - here's the thing: I believe that `arrow` should not respect `shift` in this particular case for a few reasons:

1. I added the `arrow` padding, so my floating element won't look weird and I would expect it to override `shift` as I explicitly added it.
2. As middlewares order is meaningful adding `arrow({ padding: 10 })` after `shift()` is (to me) an explicit declaration of importance.
3. This is a super edge case: the button has to be really small (smaller than the arrow) and at the edge of the screen
4. Users can easily "fix" the by using another `shift` middleware if this edge case is important to them.

The offset trick you mentioned makes it harder to calculate the right value and a burden to do it manually for the library user (even though we expose `padding` option in `arrow` middleware).

For me the only option would be to add (or rename) `arrow` options to `requiredPadding` and `preferredPadding` which *could* show the difference and make library user decide. I'm against it though.

atomiks:
@megawebmaster those are good points. I think you've convinced me haha. This looks good 😄

megawebmaster:
@atomiks - done! :smile: I'm really happy I convinced you, that was the point of adding this adjustment :wink:

atomiks:
@megawebmaster tests are failing for some reason

megawebmaster:
I'll take a look at that later this week 👍 It's a mystery to me as well.

atomiks:
Looks like it's caused by removing `place-items: center;` from the container

atomiks:
@megawebmaster you just need to re-run the update visual snapshots action in your fork

PR #1499, opened 2022-01-09
fix(react-native): bundle, async cleanup

atomiks:
- chore: avoid bundling in `core` package by referring to the CJS build in the imports, remove minified bundles
- fix: stale `computePositions` by cleaning up animation frames
- fix: `isMountedRef` boolean 
- docs: use `?? 0` in the example

### DIFF
--- a/packages/react-native/src/index.ts
+++ b/packages/react-native/src/index.ts
@@ -6,7 +6,10 @@ import {
   useCallback,
   RefObject,
 } from 'react';
-import {computePosition, arrow as arrowCore} from '@floating-ui/core';
+import {
+  computePosition,
+  arrow as arrowCore,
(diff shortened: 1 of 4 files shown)


packages/react-native/src/index.ts
```diff
@@ -95,6 +97,16 @@ export const useFloating = ({
     setLatestMiddleware(middleware);
   }
 
+  const animationFrames = useRef<Array<number>>([]);
+
+  const isMountedRef = useRef(true);
+  useLayoutEffect(() => {
+    isMountedRef.current = true;
+    return () => {
```
atomiks, line 105: Not sure if setting this to true again is required in React DOM but it seems to be required in the RN expo environment I'm using... 🤔  (I thought `useRef(true)` is fresh after it gets remounted?)

PR #1502, opened 2022-01-12
feat: expand export maps to support no/low build tooling

### LINKED ISSUE #1497: Export Map expansion?
The use of `process.env.NODE_ENV` causes issues when leveraging the `core` package in no/low build environments. Would you be open to an expansion of the export maps to better support users in this context?

I could see either the addition of the `dev`/`production` listing for the `import` entry so that `production` could point directly to the `.min.js` files that already exclude this. Similarly, I think the `browser`/`node` listings would be similarly.

This does close some debugging options, so it might be useful to include some more build options, particularly a non-minified build without the use of `process.env.NODE_ENV`, but that involves a deeper touch to the overall repo management.

Westbrook:
fixes #1497

Split the `import` fork in the export maps to include `node` and `default` subtrees.

Add `browser` and `default` subtree to initial `default`.

Add a "browser" build at `./dist/floating-ui.core.esm.browser.js` and `./dist/floating-ui.dom.esm.browser.js` the is not minified and does not include the debugging forks. It's possible that this _should_ be the inverse and include the debugging forks _always_, but let me know what you think here.

I've tested this locally in a browser with the `development` condition both on and off and it's working pretty nicely, however I'd definitely suggest a pre-/beta release with this as export maps continue to surprise me when actually consumed across a number of different contexts.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -18,7 +18,11 @@
   "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/floating-ui.core.esm.js",
+      "import": {
+        "development": "./dist/floating-ui.core.esm.development.js",
+        "production": "./dist/floating-ui.core.esm.min.js",
+        "default": "./dist/floating-ui.core.esm.js"
+      },
       "require": "./dist/floating-ui.core.cjs"
     },
     "./package.json": "./package.json"
--- a/packages/core/rollup.config.js
+++ b/packages/core/rollup.config.js
@@ -25,6 +25,13 @@ const bundles = [
   {
     input,
     output: {
+      file: path.join(__dirname, 'dist/floating-ui.core.esm.development.js'),
+      format: 'esm',
+    },
+  },
+  {
+    input,
+    output: {
       name: 'FloatingUICore',
       file: path.join(__dirname, 'dist/floating-ui.core.js'),
       format: 'umd',
@@ -47,6 +54,12 @@ const bundles = [
   },
 ];
 
+const isDevEnv = (file) => file.includes('.development.');
+const isUMD = (file) => file.includes('.core.js');
+const isMinEnv = (file) => file.includes('.min.');
+const isSpecificEnv = (file) => isMinEnv(file) || isDevEnv(file);
+const isDebugAlways = (file) => isDevEnv(file) || isUMD(file) ? 'true' : 'false';
+
 const buildExport = bundles.map(({input, output}) => ({
   input,
   output,
@@ -58,8 +71,8 @@ const buildExport = bundles.map(({input, output}) => ({
       plugins: ['annotate-pure-calls'],
     }),
     replace({
-      __DEV__: output.file.includes('.min.')
-        ? 'false'
+      __DEV__: isSpecificEnv(output.file)
+        ? isDebugAlways(output.file)
         : 'process.env.NODE_ENV !== "production"',
       preventAssignment: true,
     }),
--- a/packages/dom/package.json
+++ b/packages/dom/package.json
@@ -18,7 +18,11 @@
   "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/floating-ui.dom.esm.js",
+      "import": {
+        "development": "./dist/floating-ui.dom.esm.development.js",
+        "production": "./dist/floating-ui.dom.esm.min.js",
+        "default": "./dist/floating-ui.dom.esm.js"
+      },
       "require": "./dist/floating-ui.dom.cjs"
     },
     "./package.json": "./package.json"
--- a/packages/dom/rollup.config.js
+++ b/packages/dom/rollup.config.js
@@ -24,6 +24,13 @@ const bundles = [
   {
     input,
     output: {
+      file: path.join(__dirname, 'dist/floating-ui.dom.esm.development.js'),
+      format: 'esm',
+    },
+  },
+  {
+    input,
+    output: {
       name: 'FloatingUIDOM',
       file: path.join(__dirname, 'dist/floating-ui.dom.js'),
       format: 'umd',
@@ -52,6 +59,12 @@ const bundles = [
   },
 ];
 
+const isDevEnv = (file) => file.includes('.development.');
+const isUMD = (file) => file.includes('.dom.js');
+const isMinEnv = (file) => file.includes('.min.');
+const isSpecificEnv = (file) => isMinEnv(file) || isDevEnv(file);
+const isDebugAlways = (file) => isDevEnv(file) || isUMD(file) ? 'true' : 'false';
+
 const buildExport = bundles.map(({input, output}) => ({
   input,
   output,
@@ -64,8 +77,8 @@ const buildExport = bundles.map(({input, output}) => ({
       plugins: ['annotate-pure-calls'],
     }),
     replace({
-      __DEV__: output.file.includes('.min.')
-        ? 'false'
+      __DEV__: isSpecificEnv(output.file)
+        ? isDebugAlways(output.file)
         : 'process.env.NODE_ENV !== "production"',
       preventAssignment: true,
     }),


Westbrook:
Seems that the test failure here flakes a bit...yay webkit! Should I be worried? I've not changed any functionality and the idea that it _loads_ seems to point at this change working...

atomiks:
Yeah that particular test is flaky for whatever reason, need to investigate it.

In what contexts does this change work, can you give an example? e.g. does it work with Rollup build setups? 

I think we should also add an alias plugin for the min builds of `@floating-ui/dom` that changes `import ... from '@floating-ui/core';` to `import ... from '@floating-ui/core/dist/floating-ui.core.esm.min.js';` (etc the other formats).

Westbrook:
A consuming app that was working with rollup would likely use https://github.com/rollup/plugins/tree/master/packages/node-resolve#exportconditions at some point in their build. This will follow the export map based on the conditions at play when building/running their application.

The majority of resolutions would simply point `.` => `import` => `default` => `default` and load `./dist/floating-ui.{dom|core}.esm.min.js`.

Were the `node-resolve` plugin using `--condition=development` then it would point `.` => `import` => `default` => `development` and load `./dist/floating-ui.{dom|core}.esm.browser.js`, which is why it might be useful to include ALL debugging here, rather than remove it, but depends on the goals you want to target.

Where this file requested via `node` itself, it would resolve `.` => `import` => `node` and load `./dist/floating-ui.{dom|core}.esm.js` with no minifiation and all debugging gates as the `process.env` value would be available.

The `require` fork is unaffected by this PR as you either _require_ tooling or to be in the `node` environment to consume code in this way.

I'm less versed in other tools, however they are all expected to be compliant with https://nodejs.org/api/packages.html#package-entry-points in due course.

atomiks:
> The majority of resolutions

I'd like for the debugging version to still be default in the majority of bundlers which can handle `process`.

So this:

```js
import {computePosition} from '@floating-ui/dom';
```

points to the unminified debugging version in all these popular tools:

- Create React App 5
- vanilla webpack 5
- Parcel 2 (I think this ignores `exports` entirely and just uses `"module"`)
- Vite
- Rollup (they do need the `replace` plugin though)

Does this mean we should instead have a subpath like `@floating-ui/dom/browser` for these no/low build tools to import from?

Westbrook:
I'm gonna try to spend some more time with this, but it's a PITA to build and copy past this into projects to fully test whether they follow this report: https://webpack.js.org/guides/package-exports/#support

However, I've updated the PR here as follows:
- all `import` statements resolve to `./dist/floating-ui.{dom|core}.esm.js` as they had before
- when `--condition=development` is present the import will resolve as `./dist/floating-ui.{dom|core}.esm.development.js,`

This attempts to keep all existing usage (where `--condition=development` isn't being use) the same as it is now, while opening up the "development" builds for no/low build environments.

packages/core/package.json
```diff
@@ -18,7 +18,10 @@
   "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/floating-ui.core.esm.js",
+      "import": {
+        "development": "./dist/floating-ui.core.esm.development.js",
```
atomiks, line 22: `.development.js` or `.browser.js`? Rollup is generating the latter

packages/core/rollup.config.js
```diff
@@ -47,6 +54,15 @@ const bundles = [
...
 ];
 
+// Does this build always enter debug paths.
+const isDevEnv = (file) => file.includes('.development.');
+// Does this build get minified.
+const isMinEnv = (file) => file.includes('.min.');
+// Does this build exhibit a fixes environment: e.g. only "dev" or only "production".
+const isFixedEnv = (file) => isMinEnv(file) || isDevEnv(file);
+// Whether to always enter debug paths or next enter them.
+const fixEnvAs = (file) => isDevEnv(file) ? 'true' : 'false';
```
atomiks, line 64: I'm not sure I understand what it means to "fix" in this context? The comments also seem to have grammar problems?

atomiks (commented):
I realized it's also possible to add "guard" checks like this which works in webpack:

```js
if (typeof process !== 'undefined' && process.env.NODE_ENV !== 'production') {

}
```

This would prevent the need to create a new file. I'm not sure if it works in the other popular bundlers though. I'll need to investigate.

packages/core/package.json
Westbrook: Whoops spent a lot of time on the comments and forgot the actual change there. Update!

packages/core/rollup.config.js
Westbrook: Tweaked this some more. Let me know if it's clear now. 🙏🏼

atomiks:
> This will force webpack 4 to include process polyfill. 

Since CRA is now on webpack 5 along with most other major tools I suppose webpack 4 does not matter.

> With rollup or webpack 5 it won't really work in browser because typeof process always will be "undefined". 

You won't get debugging checks but this seems like an intentional choice for low/no build consumers?

> This also break dead code elimination. Falsy branch will not be dropped in production.

I've checked in webpack 5 and Parcel 2 and both continue to perform DCE with the guard check. I need to check Vite. Rollup won't, unless they also take into account the guard but this could be mentioned in the docs.

packages/dom/package.json
```diff
@@ -18,7 +18,10 @@
   "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/floating-ui.dom.esm.js",
+      "import": {
+        "development": "./dist/floating-ui.dom.esm.development.js",
```
atomiks, line 22: Shouldn't we have a `"production"` key here that points to the minified file?

Westbrook: That's a question for you, I think. Like `development`, `production` is not a "Node.js core condition": https://nodejs.org/api/packages.html#community-conditions-definitions so while tools can opt-into that condition, I'm not sure that any of them do so by default and beyond  https://github.com/rollup/plugins/tree/master/packages/node-resolve#exportconditions I don't have enough knowledge of alternatives to know which ones allow you to choose the condition that would be needed in that case. That being so, I think that a `production` condition might be more onerous for your consumers that helpful.

In my initial attempt I pointed `default` to the minified version, which is somewhat similar in intent as it ensures that all users who haven't chosen to get something that isn't the production version get the production version. This seems to specifically benefit projects that find their unique knowledge of their code base (which arguable includes the use of `process.env`) prepares them to make a minified version in which others might not be able to repeat the extent to which it has been minified. It really begs the question of what is standard use of a library, and with your direction I'd be happy to prepare this however you'd prefer to maintain it over time.

Westbrook:
Question: how does Rolling Versions choose which packages have changed here? I just realized it's only listing "Core" for this PR, but I've made the same changes there as in "Dom" and would expect both to publish if/when you approve this update. Is there something more I can do to support that listing both pages as expected?

atomiks:
> Question: how does Rolling Versions choose which packages have changed here? I just realized it's only listing "Core" for this PR, but I've made the same changes there as in "Dom" and would expect both to publish if/when you approve this update. Is there something more I can do to support that listing both pages as expected?

You need to update both packages, there are separate fields in the UI for each

packages/dom/package.json
atomiks: If they import the `development` condition bundle, how are they minifying it in the end without any build tooling? I'm kinda confused how people are actually using the library without Node.js/bundler being involved

Westbrook: First, without the `min` version in the `exports` map, very few modern tools will allow the `min` version to be imported. That means that ALL consumers of the library _at current_ need to minify it on their own when using node resolution tools to resolve the bare module specifiers. My original proposal of having `min` be the `default` meant that _unless_ a user wanted to start debugging they would get the most optimal build, all the time.

No/low build users come in a lot of flavors, and probably the optimal one isn't fully supported x-browser. In that case they application would use [import maps](https://github.com/WICG/import-maps) to clarify where the bare module specifiers point. JSPM actually has a great tool to see how that would look and start testing it out the some polyfils: https://generator.jspm.io/#0+JhYGBi9goO8FVwT81LLUosyS9SCE7NSVPwTSxgcEjLyU8sycxL1y3N1E/Jz3Uw0DPUMzRgyMkscTACMQFfpTO1PwA This uses the only `developer` or `production` logic gate outlined at https://nodejs.org/api/packages.html#:~:text=Must%20always%20be%20mutually%20exclusive%20with%20%22production%22. and satisfied by include `development` and `default`, but not also `production`. See how this maps the bare module specifiers to some sort of CDN, and while it's great to see you support the unpkg entry in your `package.json`, tools like JSPM and similar will only work with the files exposed by the export map. Until that's fully available x-browser, most no/low build processes still require a build at production time while focusing on reducing tooling at development time. In that way they usually leverage something for resolving bare module specifiers and not much else.

This is the situation I'm looking to support in my work, though being prepared for the longer/larger future of no/low build environments before a 1.0, after which the changes to support would amount to a breaking version, would be great!

Westbrook:
> there are separate fields in the UI for each

Not sure how I missed that. Once it's back up (it's on 503 right now), I'lll look at that close and see it I can find that and get it sorted.

packages/dom/package.json
atomiks: So "mutually exclusive" in the Node docs means you can't have _both_ `production` and `development` keys, only one of either + `default`? 🤔 

I guess this goes back to what I was mentioning before — shouldn't these users import directly from the minified file? We just need an alias plugin in our Rollup build to make this work.

```js
import {computePosition} from '@floating-ui/dom/dist/floating-ui.dom.esm.min.js';
```

 Is it also possible to specify an export that points to this file so it's cleaner to import?

```js
import {computePosition} from '@floating-ui/dom/min';
```

Westbrook: You can have both `development` and `production` but then in what case do you use `default`? I feel that “default” IS “production”, but I can totally see that being subjective to a library and am open to doing what you feel is right there. All, I want to make sure if that there is code that ships in the standard module graph that doesn’t use `process.env` so that this can be used as a browser first piece of code. In that way you could even hang ALL of this off of the `browser` condition so “default” was 100% the same as it was, we’re you to feel the current offering was the best bet for the consumers you are targeting.

atomiks: I wish this problem was easier to solve 😩 

- `"default"` = uses `process.env` checks so debugging is automatic, works in every major bundler
- `"production"` = the minified file
- `"development"` = the debugging checks as plain blocks but without `process.env`

Does that make sense?

Westbrook: Works for me. I’ll fix this up a little later today and am excited to have found a path that works for you, too!

atomiks (commented):
This looks good to me. There's only one thing that hadn't been done yet: the regular UMD development version should have no `process.env` in it (just a block like the `.development` ESM file)

packages/dom/rollup.config.js
```diff
@@ -52,6 +59,17 @@ const bundles = [
...
+// Does this build always enter debug paths.
+const isDevEnv = (file) => file.includes('.development.');
+// Is this build UMD
+const isUMD = (file) => file.includes('.dom.js');
+// Does this build get minified.
+const isMinEnv = (file) => file.includes('.min.');
+// Does this build deliver a specific environment: e.g. only "dev" or only "production".
+const isSpecificEnv = (file) => isMinEnv(file) || isDevEnv(file);
+// Whether to always enter debug paths or next enter them.
+const isDebugAlways = (file) => isDevEnv(file) || isUMD(file) ? 'true' : 'false';
```
atomiks, line 71: I don't think the comments are necessary actually, the functions are very short and self-explanatory

packages/core/rollup.config.js
```diff
@@ -47,6 +54,12 @@ const bundles = [
   },
 ];
 
+const isDevEnv = (file) => file.includes('.development.');
+const isUMD = (file) => file.includes('.dom.js');
```
atomiks, line 58: `.core.js`

Westbrook:
Hopefully everything is good to go, now! 🤞🏼

Westbrook:
Thanks for walking me through all this @atomiks! Do you have plans on when this might hit NPM?

atomiks:
I was going to add some useful JSDoc comments for the exports/types for IntelliSense purposes and do a release with the changes, should be done tomorrow

atomiks:
@Westbrook do you mind adding some documentation about this to the README + website? I'm assuming the Node process needs to run with `--conditions={development|production}`?

Westbrook:
Something along the lines of https://lit.dev/docs/tools/development/#development-and-production-builds?

atomiks:
That looks good. I'm curious about your particular use case too, something that would work for most people when not using a bundler, etc.

Westbrook:
I’ll get into it. I might also want to link to JSPM.io for that but to accurately do so the new version would already need to be released. Once I have it ready it’ll make more sense and you can decide timing and how to prevent extra PRs.

PR #1507, opened 2022-01-17
chore: min path for low/no-build tooling

atomiks:
This should allow the user to do this without any build tool (native ESM in the browser), or using Rollup as the bundler without using `replace`:

```js
import {computePosition} from '@floating-ui/dom/min';
```

@Westbrook, does this solve your case/look sound?

I still dislike this because:

- No debugging messages/checks when using the library. 
- No tree-shaking (if you're importing from a CDN or such). The library is built for DCE. More features can/will be added onto core, meaning the file size without DCE keeps increasing. The alternative is supplying individual bundles for each middleware, but then you have to deal with more HTTP requests + the install size of the library increases....

I would only recommend using this for toy apps/tutorials than anything actually running in production.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -21,7 +21,9 @@
       "import": "./dist/floating-ui.core.esm.js",
       "require": "./dist/floating-ui.core.cjs"
     },
-    "./package.json": "./package.json"
+    "./package.json": "./package.json",
+    "./dist/floating-ui.core.esm.min.js": "./dist/floating-ui.core.esm.min.js",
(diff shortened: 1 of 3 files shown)


Westbrook:
This is likely more trouble than help. While you say “only use this in toy projects”, not only does it beg questions about minifiication and tree shaking for users the sad situation is that some libraries WILL point to this 100% of the time and cause down stream consumers to load this code multiple times. 

All of these things point to why import conditions are preferred. Depending on what a consumer wants at a given time they can ensure locally that the bare module specific points to it in all usage of the library no mater the level of control they have on that consuming location.

PR #1509, opened 2022-01-20
docs: outline usage of condition based export/import maps

Westbrook:


website/pages/docs/getting-started.mdx
```diff
@@ -41,6 +41,27 @@ Floating UI can be loaded via CDN using ESM or UMD format.
...
+  "scopes": {
+    "https://ga.jspm.io/": {
+      "@floating-ui/core": "https://ga.jspm.io/npm:@floating-ui/core@__CORE_VERSION__/dist/floating-ui.core.esm.js"
+    }
+  }
+}
+</script>
+```
+
+JSPM provides a [generator](https://generator.jspm.io/) for leveraging this API with their CDN.
```
Westbrook, line 63: Might want to use a more specific URL here, e.g. https://generator.jspm.io/#4+RhYGBk9AoO8FVwT81LLUosyS9SCE7NSVPwTSxgcEjLyU8sycxL1y3N1E/Jz3Uw0DPUMzQAACoqGNU1AA but it requires a publication to surface all the capabilities we've added with the recent PR.

README.md
```diff
@@ -60,6 +60,27 @@ computePosition(referenceElement, floatingElement, {
 
 [Visit the docs for detailed information](https://floating-ui.com/docs/computePosition).
 
+## Development and production builds
+
+Floating UI is published with default, development, and production builds, using Node's support for [export conditions](https://nodejs.org/api/packages.html#packages_conditional_exports).
```
Westbrook, line 65: This was a lot of a duplication of https://lit.dev/docs/tools/development/#development-and-production-builds. Happy to massage it as you feel appropriate....

atomiks (commented):
This looks well-explained, my only problem is maybe it should be a bit shorter/more terse? The code blocks should also have the language meta (```js) + be formatted with Prettier based on the docs' settings.

I think the "Development and production" builds section can also go inside a `<Collapsible />` component so it doesn't take up too much space and the user can read it only if they want.

Westbrook:
Oh, ```MDX, I never would of thought of that. I was trying ```html 🙈

Westbrook:
_Just realizing this didn't get sent last night..._

Pulled some prose into a list and hid some stuff in a `<details>` element, let me know what you think...

The linter went crazy on when using "```html" in the one example. If you have thoughts on how to escape that, happy to clean it up some more, but working seemed more important the highlighted, here.

Separately, I couldn't get `npm run dev:website` to run a preview locally, so I ended up thrashing Netlify to test the build.

website/pages/docs/getting-started.mdx
```diff
@@ -29,18 +29,95 @@ npm install @floating-ui/dom
 yarn add @floating-ui/dom
 ```
 
+## Development and production builds
+
+Floating UI is published with default, development, and
+production builds, using Node's support for
+[export conditions](https://nodejs.org/api/packages.html#packages_conditional_exports).
+
+- `"default"{:js}`: uses `process.env.NODE_ENV{:js}`, in which
```
Westbrook, line 38: Is this a Next.js tag or a general MD parse thing? Never seen it before, but nice!

atomiks: Plugin I made to add inline highlighting https://github.com/atomiks/rehype-pretty-code. Syntax is from a really old static site generator I found in a post from 2012 🤷

PR #1521, opened 2022-01-27
Document how to solve `act` warnings

### LINKED ISSUE #1520: Appears to cause "act" warnings in React during testing
Hi! Popper v2 had problems with `act` warnings during React testing (floating-ui/react-popper#350), so I wanted to see whether Floating UI has them as well, and it appears it does. I created a minimal [demo branch](https://github.com/silvenon/repro/tree/floating-ui-act-warnings) in my repo where you can reproduce this. Let me know if I'm doing something wrong.

```
Warning: An update to App inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```

silvenon:
Closes #1520.

How's this? Too verbose?

### DIFF
--- a/website/pages/docs/react-dom.mdx
+++ b/website/pages/docs/react-dom.mdx
@@ -282,8 +282,8 @@ This ensures that you're testing the behavior the user would see in the browser.
 To fix this, explicitly call `cleanup(){:js}` at the end of your
 test if you don't care about the positioning:
 
-```js
-import {render, cleanup} from '@testing-library/react';
+```js /cleanup/
+import {render, cleanup, screen} from '@testing-library/react';
 
 test('example', () => {
   render(<Tooltip content="hello" open />);
--- a/website/pages/docs/react-dom.mdx
+++ b/website/pages/docs/react-dom.mdx
@@ -256,3 +256,68 @@ function App() {
   );
 }
 ```
+
+## Testing
+
+During testing, React requires every function that causes the state to change to be called in the [`act`](https://reactjs.org/docs/testing-recipes.html#act) callback. [Testing Library](https://testing-library.com/) already takes care of most of these, but there are still times when you have to do this yourself, for example if your component updates its state 300 ms after mounting, you need to advance timers inside of `act`:
+
+```jsx {9-11}
+import {render, screen, act} from '@testing-library/react';
+import MyComponent from './MyComponent';
+
+describe('MyComponent', () => {
+  it('updates after 300 ms', () => {
+    jest.useFakeTimers();
+    render(<MyComponent />);
+    expect(screen.getByText('same content')).toBeInTheDocument();
+    act(() => {
+      jest.advanceTimersByTime(300);
+    });
+    expect(screen.getByText('different content')).toBeInTheDocument();
+    jest.useRealTimers();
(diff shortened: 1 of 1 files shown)


atomiks:
Yeah I think it's a little verbose, but I can understand that giving some context is useful. 

I just think it should be really short, straight to the point, and relevant for Floating UI itself without anything else distracting. My take (**Prettier formatted the line width which GitHub respects, won't look like this on the site**):

## Testing

Floating UI performs a state update during a Promise microtask,
which will happen **after** your test completes. This will result
in an act warning like this:

```
Warning: An update to App inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```

To solve this, ensure you wait for the microtask queue to be
flushed before finishing the test.

A deliberate way of doing this is like so:

```js
render(<Dropdown open />);
await act(async () => {});
```

Testing Library's
[`waitFor`](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor)
utility is usually more appropriate:

```jsx
const {rerender} = await waitFor(() => render(<MyComponent />));
```

Or:

```jsx
render(<MyComponent />);
await waitFor(() => {
  expect(screen.getByRole('tooltip')).toHaveStyle({
    position: 'absolute',
  });
});
```

This would be needed only for the first assertion after the state
gets updated.

silvenon:
I realized that I'm a little torn about recommending `waitFor`, especially for assertions, because they work without the `waitFor` so it's misleading. The point of `waitFor` is waiting for the function to stop throwing/rejecting, which is not the use case here. I personally prefer the empty `act` call the most because it's most obvious that it's a workaround. What do you think?

atomiks:
Yeah I guess it depends what you're actually testing. Are you testing something related to its position, or is the position irrelevant to what you're testing, so the `act` warning is just an annoyance rather than something to be concerned about?

The latter would make sense to wait for it to use the deliberate flush method. `waitFor` would make sense if you need it to be positioned for the assertion. 

`waitFor(() => render(...))` implicitly waits for the queue to be flushed I'm assuming, not sure what the best recommendation is overall 🤷‍♂️

atomiks:
Hmm I guess the `afterEach()` unmount call done by Testing Library also happens in a microtask or something. Because this also works:

```js
import {cleanup} from '@testing-library/react';

test('Dropdown', () => {
  render(<Dropdown />);

  // make assertions
  
  cleanup(); // `setData()` won't be called now
});
```

silvenon:
Testing Library automatically unmounts in `afterEach`, but the warnings still happen 🤔 

I'll think about this a bit and update my PR.

silvenon:
Ah, I see! I'll think about this a bit and update my PR 😃

atomiks:
Made some changes which I think make sense given what we found out here, merging

PR #1526, opened 2022-02-03
Add Conditional Exports in Popper v2

### LINKED ISSUE #1465: [Popper] JSPM.org Not Providing the /dist/esm URL
I'm using the [Online Import Map Generator](https://generator.jspm.io/) at JSPM.org to see which URL to use for a Popper.js `importmap` reference.  I have "Browser" checked and "Node" unchecked, but it is giving me back:

```
https://ga.jspm.io/npm:@popperjs/core@2.11.0/lib/index.js
```

Here is a minimal example:

```html
<html>
  <head>
    <script type="importmap">{
      "imports": {
        "@popperjs/core": "https://ga.jspm.io/npm:@popperjs/core@2.11.0/lib/index.js"
      }
    }</script>
    <script type="module">
      import { createPopper } from "@popperjs/core";
      createPopper(button, tooltip);
    </script>
  </head>
  <body>
    <button id="button" aria-describedby="tooltip">My button</button>
    <div id="tooltip" role="tooltip">My tooltip</div>
  </body>
</html>
```

This will crash on:

```
Uncaught ReferenceError: process is not defined
createPopper.js:81
```

This is because in the browser there is no `process.env.NODE_ENV` value provided.  According to your [distribution targets documentation](https://github.com/floating-ui/floating-ui/tree/v2.x#distribution-targets) I think the correct "Browser" URL is supposed to be:

```
https://ga.jspm.io/npm:@popperjs/core@2.11.0/dist/esm/index.js
```

Note that I'm really asking about this because the [importmap-rails](https://github.com/rails/importmap-rails) gem makes an API call to JSPM.org to get this URL.  See [this comment for more details](https://github.com/rails/importmap-rails/issues/65#issuecomment-1000389955).

Are you able to provide more info to JSPM.org so that they know about the `/dist/esm` path?

chadrschroeder:
Fixes #1465.

This pull request is similar to https://github.com/floating-ui/floating-ui/pull/1502, but for the `v2.x` branch.

I've left all existing distributions in place and have added 2 new files to use in the `development` and `production` conditional exports:

Exports Entry | File Name | Notes
:--- | :--- | :---
`import.development` | `dist/esm/index.development.js` | New un-minified ESM build with debugging on.
`import.production` | `dist/esm/index.min.js` | New minified ESM build with debugging off.
`import.default` |  `lib/index.js` | Existing ESM build.  Same as `module`.
`require` | `dist/cjs/popper.js` | Existing CJS build.  Same as `main`.

I copied the entire "exports" format from the current version which means I also brought back the update from https://github.com/floating-ui/floating-ui/pull/1462.  I can leave that out if you'd prefer.

Please let me know if I'm doing anything wrong here or if there's a different approach you would recommend.

### DIFF
--- a/.config/rollup.config.js
+++ b/.config/rollup.config.js
@@ -8,14 +8,14 @@ import flowEntry from 'rollup-plugin-flow-entry';
 import compiler from '@ampproject/rollup-plugin-closure-compiler';
 import pkg from '../package.json';
 
-const getFileName = (input) => input.split('/').reverse()[0].split('.')[0];
+const getFileName = (input) => input.split('/')[1].split('.')[0];
 
 const inputs = ['src/popper.js', 'src/popper-lite.js', 'src/popper-base.js', 'src/enums.js'];
 const bundles = [
   { inputs, format: 'umd', dir: 'dist', minify: true, flow: true },
   { inputs, format: 'umd', dir: 'dist' },
   { inputs, format: 'cjs', dir: 'dist', flow: true },
-  { inputs: ['dist/esm/index.js'], format: 'esm', dir: 'dist', minify: true },
+  { inputs: ['lib/index.js'], format: 'esm', dir: 'dist', minify: true },
   { inputs: ['lib/index.js'], format: 'esm', dir: 'dist', development: true },
 ];
 
@@ -28,9 +28,9 @@ const configs = bundles
           replace({
             __DEV__: minify ? 'false' : 'true',
           }),
-        format === 'esm' && development &&
+        format === 'esm' &&
           replace({
-            'process.env.NODE_ENV': '"development"',
+            'process.env.NODE_ENV': development ? '"development"' : '"production"',
           }),
         babel({ babelHelpers: 'bundled' }),
         // The two minifiers together seem to procude a smaller bundle 🤷‍♂️
--- a/.config/rollup.config.js
+++ b/.config/rollup.config.js
@@ -15,8 +15,8 @@ const bundles = [
   { inputs, format: 'umd', dir: 'dist', minify: true, flow: true },
   { inputs, format: 'umd', dir: 'dist' },
   { inputs, format: 'cjs', dir: 'dist', flow: true },
-  { inputs: ['lib/index.js'], format: 'esm', dir: 'dist', minify: true },
-  { inputs: ['lib/index.js'], format: 'esm', dir: 'dist', development: true },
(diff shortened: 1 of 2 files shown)


.config/rollup.config.js
```diff
@@ -8,24 +8,30 @@ import flowEntry from 'rollup-plugin-flow-entry';
 import compiler from '@ampproject/rollup-plugin-closure-compiler';
 import pkg from '../package.json';
 
-const getFileName = (input) => input.split('/')[1].split('.')[0];
+const getFileName = (input) => input.split('/').reverse()[0].split('.')[0];
```
chadrschroeder, line 11: The old version of `getFileName` would take a name like `dist/esm/index.js` and give back `esm`.

.config/rollup.config.js
```diff
@@ -8,24 +8,30 @@ import flowEntry from 'rollup-plugin-flow-entry';
...
       input,
       plugins: [
         format === 'umd' &&
           replace({
             __DEV__: minify ? 'false' : 'true',
           }),
+        format === 'esm' && development &&
+          replace({
+            'process.env.NODE_ENV': '"development"',
+          }),
```
chadrschroeder, line 34: Generating the "development" file is a little weird. Because I didn't want to alter anything about the existing build, [`__DEV__` will still be replaced by `babel-plugin-dev-expression`](https://github.com/floating-ui/floating-ui/blob/v2.x/.config/babel.config.js#L32). Then I replace the `process.env.NODE_ENV` reference after that.

I don't actually need the "development" file for my purposes, so if you want me to leave this out completely, I can do that.  I'm just trying to provide a "development" file similar to the current version.

dominikg:
this was added before https://github.com/floating-ui/floating-ui/pull/1342 and rolled back https://github.com/floating-ui/floating-ui/pull/1362  due to it being a breaking change. (adding an exports map always is)

I think it is worth supporting esm with export maps but it would have to be in a new major release.

atomiks:
Based on Rich's comment here: https://github.com/sveltejs/kit/issues/4504#issuecomment-1135338008 

Would it be enough to create a new `mjs` folder build? `esm` used to have `.mjs` as the extension name, but it changed to `.js` in v2.0.4 for some reason — at this point, it's probably too major of a breaking change to revert that back to `.mjs`, right? I don't think we can add an export map at all though.

But we could add a new folder using that extension. Libraries could then import from that folder instead of `/esm`, which would basically be the same as a new major.

PR #1529, opened 2022-02-06
Fix CJS compatibility when used as a transitive dependency

fshafiee:
### Context

We have this universal React web application where
1. Both CJS and ES modules are used to build the final product (for server-side and client-side respectively).
2. We are using Razzle to build the app for both environments.
3. We maintain our own UI Kit which utilizes `@floating-ui` in couple of its components.
   - It's a separate package (but part of a monorepo).
   - It's going to be open-sourced soon.
   - It targets both ES and CJS modules using conditional exports, much like `@floating-uit`.
4. Web application depends on this UI Kit. As result `@floating-ui` is a transitive dependency for this application.
5. Web application is SSR enabled and uses ReactDOM for this purpose.
6. Our build and tooling stack is up-to-date:
   - webpack 5
   - razzle 4
   - typescript 4 

### Observations

1. Pages that were using the components that had dependency on `@floating-ui` could not be rendered server-side, throwing an error complaining about the imported functions from '@floating-ui/react-dom' not being functions.
   error: TypeError: (0 , react_dom_1.offset) is not a function
2. The issue persisted with dev and production builds alike, so HMR is not causing this.
3. If the users were routed to those pages client-side (where no SRR was involved) everything worked okay.

### Prognosis

(1) and (2) gave me a hint that the issue might be caused by the export structure of `@floating-ui/react-dom` module. So, I started playing around with it.

### Diagnosis

1. ⚠️ Replacing
  https://github.com/floating-ui/floating-ui/blob/b9f683d14a426767de93475125b32701f2f6ad84/packages/react-dom/package.json#L15
   with
   ```json
   "main": "dist/floating-ui.react.dom.js"
   ```
   fixed the issue (using the **umd** build as main entry instead of **cjs**).

2. ❌ Using **typescript** to build the `@floating-ui` packages instead of rollup (which is the same build strategy as our UI Kit) had no effect, which essentially ruled out module interops being an issue.

3. ❌ Importing core functions directly from `@floating-ui/core` caused the same errors.
   -  ⚠️ Changing the `@floating:ui/core` main entry from **cis** to **umd** fixed the problem again.

4. ⚠️ Importing functions directly to into the app didn't cause any errors! 
   - ❌ Making it a peer dependency didn't fix the problem. 

5. ❌ Removing the `"type": "module"` didn't fix the issue.

6. ❌ Changing the extension from `.cjs` to `.js` caused bundling errors requiring the extension to be `.cjs` since `"type": "module"` was set.

7. ✅ Doing (4) and (5) together fixed the issue.

### Conclusion

This is the only package that's causing an error for our UI Kit package, and I know of [one other package that was causing a similar problem](https://github.com/dcastil/tailwind-merge/commit/0bbc258421423d248d23e5082726cac497a45465#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) and they too reversed course, I've reached to the conclusion that `"type": "module"` is not the best strategy for universal packages at the moment and it breaks compatibility in unexpected ways.

Keep in mind that if these functions are imported directly into the web app (Obs. 4), they resolve just okay, so my understanding is that the `@floating-ui` package as it stands, **cannot be a transitive CJS dependency**.
- ✅ `App → CJS @floating-ui`
- ❌ `App → CJS package → CJS @floating-ui`

Looking at it another way, it does makes sense to keep CJS as the default type which legacy tools recognize better, and rely on conditional export for ESM tooling which are more modern. This was also affirmed for me during a research I had recently around popular strategies to publish NPM packages. They either ship with CJS or keep the CJS default as default just to ensure broader compatibility surface with build tools and IDEs.

Right now, these package have been published under my personal NPM account in order to unblock my team. Check them if you want to investigate further:
- [@farzadsh/floating-ui-core](https://www.npmjs.com/package/@farzadsh/floating-ui-core)
- [@farzadsh/floating-ui-dom](https://www.npmjs.com/package/@farzadsh/floating-ui-dom)
- [@farzadsh/floating-ui-react-dom](https://www.npmjs.com/package/@farzadsh/floating-ui-react-dom)


atomiks:
This unfortunately won't work with Nuxt 3... this seems like a problem of your build setup or using an old version of Node?

fshafiee:
> This unfortunately won't work with Nuxt 3... this seems like a problem of your build setup or using an old version of Node?

Hey @atomiks, thank you for responding so quickly.

We are using **Node v16.13.0** for dev and deployment alike.

atomiks:
Nuxt 3 requires the package to be an ES module which is the way forward. Using `.mjs` extension doesn't work with CRA 4 and maybe others. I think we're stuck with the current setup. 

`.cjs` works fine as a transitive dependency in native Node when doing `node index.js` and requiring the CJS file, so maybe it's a problem with Razzle or something? Why wouldn't the current setup work? 

You don't need to create your own packages, just use `patch-package` and patch `.cjs` to `.js` for each package if this somehow can't be fixed

fshafiee:
Unfortunately for us, `patch-package` is not a viable option since we are open sourcing our package and our target audience is react community which is heavily invested in Webpack tooling right now.

Just to make sure how incompatible it is with Nuxt 3, I fired up a barebone project. The dev server worked just okay, but in order to build an deploy it, I had to add the package to `transpile` option

```json
```
 
So pardon me for republishing the package.

fshafiee:
Unfortunately for us, `patch-package` is not a viable option since we are open sourcing the UI Kit and our target audience is React community which is heavily invested in Webpack tooling right now. So pardon me for republishing the package. I'd have to keep them up for now.

For the sake of the argument, I fired up a bare-bone Nuxt 3 project. The dev server worked just okay, but in order to build and deploy it, I had to add the package to `transpile` option

```ts
export default defineNuxtConfig({
  build: {
    transpile: ['@farzadsh/floating-ui-dom']
  }
})
```

This, IMO, is so much simpler than configuring webpack (believe me, I tried, and failed). So, I prefer to keep the webpack experience optimized as Nuxt 3 is still in Beta.

Hopefully React (and JS community as a whole) will catch up soon and we no longer have to be in this mess. 

Feel free to close this MR if you plan to stick with Nuxt 3.

Cheers,

atomiks:
Ideally the vast majority of users are supported, the thing is React already works fine in Create React App, Parcel, Vite, webpack (Next.js), etc... maybe you should be using ESM on the server instead of CJS?

fshafiee:
@atomiks we will definitely move to ESM at some point for our own applications (it breaks more things than it fixes at the moment). The problem is that we don't have control over the 3rd parties that'll be using our package. If we weren't concerned with the 3rd parties, we could've just used the patch-package as you mentioned and be done with it. But right now we are committed to supporting both environments.

Thanks James for the hints and the support. I learned more about the situation in this conversation than I did scrubbing the Internet for last couple of days. ✌️

gabro:
Hi, just dropping here to report the exact same issue, with a very similar setup:

- we have a `design-system` package which exports a `Tooltip` component using Floating UI
- our `app` package depends on the `design-system` package and uses the `Tooltip` component

This causes the same error as described in this issue.

Our `app` is a Create React App without further customizations.

Yes, Floating UI works if _directly_ imported by a CRA-based app, but I would argue that it's fairly common to split a component library into a separate project, which would trigger the issue.

atomiks:
I would rather support the React ecosystem than Nuxt but I'm not sure I understand _why_ this is failing exactly? Is it a problem with webpack or something else?

Nuxt being supported using the "transpile" option is OK as long as it can actually be used, I guess tree-shaking is broken though?

gabro:
I have limited understanding of how the whole thing works, but when I build my design-system package, the output contains this (not minified for readability):

```tsx
// src/Tooltip/Tooltip.tsx
var import_react39 = require("react");
var import_react_dom = require("@floating-ui/react-dom");

// ... the rest of the tile
```

I'm using tsup to build it (which uses esbuild underneath) and the default bundle format is cjs.

Then at runtime, `import_react_dom` gets resolved to the string '/static/media/floating-ui.react-dom.0e771cd1b8e9e57ef577.cjs':

image

(as opposed to being resolved to a JS object or a module)

Definitely something odd going on with webpack there, since I would have expected the `require` to resolve to the cjs bundle.

atomiks:
Can you change the format to ESM? CJS breaks tree-shaking which inflates the size of this lib quite a bit.

gabro:
You mean the format of the design-system package? I can try, give me a second

atomiks:
I've made a new PR here: https://github.com/floating-ui/floating-ui/pull/1552

Can you tell me if this works for your cases? Based on @fshafiee's first dot point, it does?

Issue #1538, opened 2022-02-20
[Umbrella] Roadmap to Floating UI v1


atomiks:
![v1](https://user-images.githubusercontent.com/22450188/158020479-b9f1ff50-34d2-4ed3-b9d4-4fa7ffb211b9.jpg)

# Roadmap

Part of the rebranding to Floating UI was that the package would eventually support "interactions" logic (like Tippy). This has taken a while and I had a lot of false starts but I think it's close to an early release 🥳 . It will start at v0 and it's React-only, as re-inventing `useEffect` for vanilla was a mess. The logic might be able to be shared across frameworks eventually though.

In the meantime, I also want to get the main positioning library to v1. Some have been holding off adopting the package due to it being v0, so the sooner this is fully stable the better. These are the current breaking changes I want to make. For the most part these won't require major refactors for v0 users as the API is quite thin anyways.

If you're making a library and using this package I recommend not exposing any of the underlying API of Floating UI. This will allow you to update easily without affecting your consumers. The main problem with such a low-level library is that it can't update frequently due to churn (Popper is used as a dependency in thousands of packages). I want to get the API right for v1 so the major version will be solid for a long time.

v1 be released by the end of March. Feel free to share any problems with the API you have.

## 1. Work more with logical properties rather than physical ones ☑️

This means the placement string `${physicalSide}-${logicalAlignment}` should support RTL automatically. In addition `offset.crossAxis` should also adapt to the alignment appropriately.

Issues: 

- https://github.com/floating-ui/floating-ui/issues/1530
- https://github.com/floating-ui/floating-ui/issues/1170 

## 2. ~Merge `flip` and `autoPlacement` into one generic `placement`-changing middleware~

Ditched this idea for now

## 3. Rename `getScrollParents` to `getOverflowAncestors` export an official API for updating automatically ☑️

These are the cases you'll want to update:

- Ancestor scrolling
- Ancestor resizing
- Element resizing
- Reference element changing location on the screen (transform scale, animations, etc...) 

## 4. Make `platform` object less dependent on DOM terms ☑️

There are terms like `getDocumentElement()` that are a little too DOM-y and could be made more agnostic. I need to make a Canvas Floating UI POC to dogfood the names. Charting library tooltip stuff is the best example I've seen where Floating UI could be used in a `<canvas />`

## 5. Improve lifecycle resetting/skipping ☑️

[...]

PR #1539, opened 2022-02-21
feat: RTL placement

atomiks:
Solves No. 1 in https://github.com/floating-ui/floating-ui/issues/1538


packages/core/src/utils/getAlignmentSides.ts
```diff
@@ -6,21 +6,22 @@ import {getAlignment} from './getAlignment';
...
 } {
-  const isStart = getAlignment(placement) === 'start';
+  const alignment = getAlignment(placement);
   const mainAxis = getMainAxisFromPlacement(placement);
   const length = getLengthFromAxis(mainAxis);
 
   let mainAlignmentSide: BasePlacement =
     mainAxis === 'x'
-      ? isStart
+      ? alignment === (rtl ? 'end' : 'start')
```
FezVrasta, line 21: I see this is repeated a bit around the code, may be worth having a dedicated utility to convert these?

FezVrasta: It could also make sense to define some enums for `end` and `start` so that they can be used for comparison in a more type-safe way?

atomiks: TS compares the strings literally so there's no need for a separate variable. The code is kinda unique for each case so don't think it's worth splitting

PR #1540, opened 2022-02-21
fix: `shift` middleware should be a part of `MiddlewareData` type

ling1726:
According to the [middleware documentation](https://floating-ui.com/docs/shift#data), the middleware should return the values that the floating element has been shifted by in middleware data. All other middleware data are already a part of the `MiddlewareData` type except for `shift`.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -70,6 +70,7 @@ export type MiddlewareData = {
     skip?: boolean;
   };
   offset?: Coords;
+  shift?: Coords;
   [key: string]: any;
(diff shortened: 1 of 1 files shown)


atomiks:
`limitShift` isn't a middleware, it's a function that returns a value to be used by `shift.limiter`

ling1726:
> `limitShift` isn't a middleware, it's a function that returns a value to be used by `shift.limiter`

Oh my bad, I guess the problem I'm actually solving then is adding `shift` to the middleware data types

PR #1541, opened 2022-02-22
feat(dom): autoUpdate

atomiks:
Solves No. 3 in #1538

⚠️  WIP, needs tests, first I want to get the API ideas right

cc: @diondiondion

Firstly, `getScrollParents` has been renamed to `getScrollableAncestors` as I think it is more accurate. Though, it still considers `overflow: hidden` elements. Maybe, `getOverflowAncestors` instead?

Secondly, this export has 4 options and it looks like this:

```js
import {autoUpdate} from '@floating-ui/dom';

async function update() {
  // computePosition call in here
}

const cleanup = autoUpdate(reference, floating, update, {
  // default options
  ancestorScroll: true,
  ancestorResize: true,
  elementResize: true,
  everyFrame: false,
});
```

## ancestorScroll

Update whenever a scrollable ancestor is scrolled

## ancestorResize

Update whenever a scrollable ancestor is resized (as this is only the native `resize` event, only works on `window` and `VisualViewport`)

## elementResize

Update whenever the reference or floating elements change in size using a `ResizeObserver`

## everyFrame

Watch every frame for reference rect changes and only update when it changes location on the screen (e.g. for animations or general layout changes). This disables all the above listeners since they become unnecessary. I think there may be situations where the floating element needs to be updated even if its `clientRect` values didn't change, but sure of which.

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -17,6 +17,7 @@ export function autoUpdate(
   update: () => void,
   options: Partial<Options> = {}
 ) {
+  let cleanedUp = false;
   const everyFrame = options.everyFrame;
   const ancestorScroll = options.ancestorScroll && !everyFrame;
   const ancestorResize = options.ancestorResize && !everyFrame;
@@ -47,6 +48,10 @@ export function autoUpdate(
   }
 
   function frameLoop() {
+    if (cleanedUp) {
+      return;
+    }
+
     const nextRefRect = reference.getBoundingClientRect();
 
     if (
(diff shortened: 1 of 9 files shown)


FezVrasta (approved):

Slapbox:
@atomiks can you clarify what `ancestorScroll` does more exactly?

I'd expect it to update the position of the element if the ancestor of the `referenceEl` scrolls but I guess I'm misunderstanding the intention because nothing happens when I scroll an ancestor, even if it's the direct ancestor of the `referenceEl`.

PR #1564, opened 2022-03-03
feat(offset): alignmentAxis

atomiks:
cc: @ling1726, @FezVrasta 

`crossAxis` works as `skidding` did in Popper v2, this new `alignmentAxis` option allows alignments to be offset consistently. If the placement is not aligned then nothing happens.

Also, @ling1726 does this inversion make sense to you? A `bottom` placement with `crossAxis` inverts in RTL direction. A `positive` number means the other thing.

https://user-images.githubusercontent.com/22450188/156538569-f7bfe13e-d4d0-4cfb-9b33-104722e08534.mov


ling1726:
> Also, @ling1726 does this inversion make sense to you? A bottom placement with crossAxis inverts in RTL direction. A positive number means the other thing.

I think so it's what we are doing ourselves in Popper v2
https://github.com/microsoft/fluentui/blob/17a19427f5465577427f5c1c4ed61578a063f990/packages/react-positioning/src/utils/positioningHelper.ts#L50-L54

atomiks:
I think `alignmentAxis` makes the most sense as it's consistent with the other two options. Kind of hard to name though. It's set to `null` explicitly by default, by setting it to a number you override `crossAxis` for aligned placements and makes it work logically. Think this makes the most sense.

PR #1574, opened 2022-03-08
Add workflow_dispatch for manual releases

ForbesLindesay:
Unfortunately GitHub doesn't let us specify a branch with a `repository_dispatch` event, so they can only be used to release from the default branch. If you need to release from a different branch, you can use a `workflow_dispatch` instead. This will let you manually trigger the workflow on your desired branch from the GitHub actions UI.

I plan on updating Rolling Versions to use `workflow_dispatch` in the future as well, this will likely be done by allowing custom configuration of the trigger with something like:

```toml
[release_trigger]
  type = "github_workflow_trigger"
  name = "release"
```

This likely will not be ready for at least a month or two though. For now, you can specify both `repository_dispatch` and `workflow_dispatch` in your workflow file, and then use the GitHub UI when you need to release from a custom branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,7 @@ name: Release
 on:
   repository_dispatch:
     types: [rollingversions_publish_approved]
+  workflow_dispatch:
 
 jobs:
   publish:


atomiks:
Thanks!

FezVrasta:
I think workflows only work if they are defined on the master branch, we'll need to move this one on master.

FezVrasta:
@ForbesLindesay we have a Popper Release workflow that should work for this use case but for some reason it's failing. 

https://github.com/floating-ui/floating-ui/runs/5479633104?check_suite_focus=true

PR #1593, opened 2022-03-25
fix: window check for Electron

### LINKED ISSUE #1592: `isWindow` function doesn't work correctly in Electron
When you are working in Electron, the string representation of window returns `[object global]`, not `[object Window]` what you expect. Because of this, the overlay component does not work correctly in Electron
https://github.com/floating-ui/floating-ui/blob/ae2e56e751d247d81b162112cd46238c5290e685/packages/dom/src/utils/window.ts#L1-L3

atomiks:
Closes #1592

### DIFF
--- a/packages/dom/src/utils/getNodeScroll.ts
+++ b/packages/dom/src/utils/getNodeScroll.ts
@@ -1,16 +1,16 @@
 import {NodeScroll} from '../types';
-import {isWindow} from './window';
+import {isElement} from './is';
 
 export function getNodeScroll(element: Element | Window): NodeScroll {
-  if (isWindow(element)) {
+  if (isElement(element)) {
     return {
-      scrollLeft: element.pageXOffset,
-      scrollTop: element.pageYOffset,
+      scrollLeft: element.scrollLeft,
+      scrollTop: element.scrollTop,
     };
   }
 
   return {
-    scrollLeft: element.scrollLeft,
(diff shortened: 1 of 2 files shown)


packages/dom/src/utils/getNodeScroll.ts
```diff
@@ -1,16 +1,16 @@
 import {NodeScroll} from '../types';
-import {isWindow} from './window';
+import {isElement} from './is';
 
 export function getNodeScroll(element: Element | Window): NodeScroll {
-  if (isWindow(element)) {
+  if (isElement(element)) {
```
atomiks, line 5: This change alone makes it work like Popper, but as I mentioned I think there's likely an issue with iframes in Electron if we also don't change `isWindow` check. That means this isn't strictly necessary but it's probably safer

packages/dom/src/utils/window.ts
```diff
@@ -1,5 +1,11 @@
 export function isWindow(value: any): value is Window {
-  return value?.toString() === '[object Window]';
+  return (
+    value &&
+    value.document &&
+    value.location &&
+    value.alert &&
+    value.setInterval
+  );
```
atomiks, line 8: @deface-37 does this check solve the issue?

deface-37: Yes, it does. Thanks

PR #1596, opened 2022-03-30
improve sort compareFunction

ZZITE:
A small optimization

- calculate before sorting, reducing duplicate calculations
- reduce duplicate code

### DIFF
--- a/packages/core/src/middleware/flip.ts
+++ b/packages/core/src/middleware/flip.ts
@@ -129,7 +129,6 @@ export const flip = (
                 .reduce((acc, overflow) => acc + overflow, 0),
             }))
             .sort((a, b) => a.overflowsSum - b.overflowsSum)[0]?.data.placement;
-          console.timeEnd('test loop');
           if (placement) {
             resetPlacement = placement;
           }
--- a/packages/core/src/middleware/flip.ts
+++ b/packages/core/src/middleware/flip.ts
@@ -123,12 +123,13 @@ export const flip = (
         case 'bestFit': {
           const placement = overflowsData
             .map((d) => ({
-              ...d,
+              data: d,
               overflowsSum: d.overflows
                 .filter((overflow) => overflow > 0)
                 .reduce((acc, overflow) => acc + overflow, 0),
             }))
-            .sort((a, b) => a.overflowsSum - b.overflowsSum)[0]?.placement;
+            .sort((a, b) => a.overflowsSum - b.overflowsSum)[0]?.data.placement;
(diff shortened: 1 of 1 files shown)


packages/core/src/middleware/flip.ts
```diff
@@ -122,16 +122,13 @@ export const flip = (
...
-              (a, b) =>
-                a.overflows
-                  .filter((overflow) => overflow > 0)
-                  .reduce((acc, overflow) => acc + overflow, 0) -
-                b.overflows
-                  .filter((overflow) => overflow > 0)
-                  .reduce((acc, overflow) => acc + overflow, 0)
-            )[0]?.placement;
+            .map((d) => ({
+              ...d,
```
atomiks, line 126: Spreading inside a loop is usually a perf hit. Does the performance improve when testing with `console.time()` / `console.timeEnd()`?

ZZITE: You are right. I've reworked the code, please confirm.
I increased the amount of data  a little in the test, and the result shows that it takes less time.

- before 
  WechatIMG350
- after
  WechatIMG351

packages/core/src/middleware/flip.ts
```diff
@@ -122,16 +122,13 @@ export const flip = (
...
-                  .filter((overflow) => overflow > 0)
-                  .reduce((acc, overflow) => acc + overflow, 0)
-            )[0]?.placement;
+            .map((d) => ({
+              data: d,
+              overflowsSum: d.overflows
+                .filter((overflow) => overflow > 0)
+                .reduce((acc, overflow) => acc + overflow, 0),
+            }))
+            .sort((a, b) => a.overflowsSum - b.overflowsSum)[0]?.data.placement;
```
atomiks, line 131: I think you can just use a tiny variable name here because it's local, helps with minification for the consumers. Maybe map it to an array rather than an object?

```js
.map(d => [d, summingResult]).sort(...)
```

ZZITE: Using array does reduce overhead better,i did~

ZZITE: I found Functional tests failed, but I don't know why  >_<

atomiks: Seems like a flaky test and unrelated to it. Will prob pass if you run it again, will need to look into it

ZZITE: passed

atomiks: Thanks for this ❤️‍🔥

PR #1603, opened 2022-04-04
Autoupdate -> fix for ancestor scrolling

denisborovnev:
Hi! 

This PR fixes autoupdate for following case
```
<div id="scroll" style="overflow: auto">
  <div id="container" style="overflow: hidden">
       <div id="reference></div>
  <div/>
</div>
```
In this case the popup position is not updated when user scrolls the div with id "scroll".  
That happens because getOverflowAncestors misses the parent element of the found overflow ancestor:  

`// take parent first time`  
https://github.com/floating-ui/floating-ui/blob/ffdaa13142fcc3ff300c3472c1fc773a8ba88452/packages/dom/src/utils/getOverflowAncestors.ts#L24  

https://github.com/floating-ui/floating-ui/blob/ffdaa13142fcc3ff300c3472c1fc773a8ba88452/packages/dom/src/utils/getOverflowAncestors.ts#L10
`// take parent second time`  
https://github.com/floating-ui/floating-ui/blob/ffdaa13142fcc3ff300c3472c1fc773a8ba88452/packages/dom/src/utils/getNearestOverflowAncestor.ts#L5-L6
____________
Looks like this behavior was broken in this PR https://github.com/floating-ui/floating-ui/pull/1565/files .   
To fix this  just run getOverflowAncestors recursion on the current overflow element 

Also here the example https://codesandbox.io/s/cranky-meadow-pz7m1g?file=/src/App.js of this issue

### DIFF
--- a/packages/dom/src/utils/getOverflowAncestors.ts
+++ b/packages/dom/src/utils/getOverflowAncestors.ts
@@ -1,5 +1,4 @@
 import {getNearestOverflowAncestor} from './getNearestOverflowAncestor';
-import {getParentNode} from './getParentNode';
 import {getWindow} from './window';
 import {isOverflowElement} from './is';
 
@@ -21,5 +20,5 @@ export function getOverflowAncestors(
   return isBody
     ? updatedList
     : // @ts-ignore: isBody tells us target will be an HTMLElement here
-      updatedList.concat(getOverflowAncestors(getParentNode(target)));
+      updatedList.concat(getOverflowAncestors(target));
 }


atomiks:
Thank you for the PR, can you add a test for this?

atomiks:
I've set up DOM unit tests (to avoid Playwright for this test) locally, I'm gonna merge this and add it separately. Thank you :)

denisborovnev:
> I've set up DOM unit tests (to avoid Playwright for this test) locally, I'm gonna merge this and add it separately. Thank you :)

Thank you! By the way I have added Playwright test for this case https://github.com/LogicSoftware/floating-ui/commit/1ba9795b40eca0ace20120d51557ad2df4755829 😊

PR #1607, opened 2022-04-05
Fix "Can't find variable: ResizeObserver" on old browsers

sedubois:


atomiks:
I don't think this makes much sense, instead if you need to support the old browsers:

- Polyfill `ResizeObserver` for those browsers, or
- Disable the `elementResize` option and manually update when their sizes change so all browsers behave consistently

Right now this masks the error and results in inconsistency

sedubois:
Currently when using the floating-ui library as it is on an old browser, it breaks (raises the stated error). The idea was to simply prevent that error from being raised. I was under the impression that doing nothing instead of raising an error may be a better solution than the status quo, as the number of users on such old browsers is getting smaller and smaller. 

Using a polyfill would be putting a cost on the majority of users which I considered not worth it. In line with [graceful degradation](https://www.thoughtco.com/graceful-degradation-in-web-design-3470672), “When older, less feature-rich browsers view the site, it should degrade in a way that is still functional but possibly with fewer features or different display visuals.”

An alternative would have been to output an explanation in the log on library init, explaining that ResizeObserver is not available and must either be polyfilled or the option should be deactivated (either completely or with the typeof proposed here). But I personally preferred the first option.

atomiks:
> Currently when using the floating-ui library as it is on an old browser, it breaks (raises the stated error). The idea was to simply prevent that error from being raised.

If you need to support old browsers but don't want to polyfill, then it makes sense to make it consistent for all:

```js
autoUpdate(ref, flo, update, {elementResize: false});
```

And be mindful of changes to the sizes of each element, requiring a manual update.

sedubois:
Yes, that is what I already do. I was just proposing this PR to improve the default behaviour of the library on old browsers. The current error "Can't find variable: ResizeObserver" is unhelpful.

atomiks:
Maybe we should add a support note to the docs

Nicolerobinn:
So how do we solve this problem？

PR #1613, opened 2022-04-11
fix(react-dom): React 18 support

atomiks:
This PR:

- Updates playgrounds to React 18
- Uses the `flushSync` API in order to ensure the browser doesn't paint before the positioning is complete.


packages/react-dom/src/index.ts
```diff
@@ -74,22 +75,24 @@ export function useFloating({
       strategy,
     }).then((data) => {
       if (isMountedRef.current) {
-        setData(data);
+        flushSync(() => {
+          setData(data);
+        });
```
atomiks, line 80: Fixes https://github.com/floating-ui/react-popper/issues/349, we also need to make this change in `react-popper`

PR #1615, opened 2022-04-11
feat: make reference type in useFloating generic

### LINKED ISSUE #1612: Make useFloating type generic
Currently, the type of useFloating says that `reference` can always be a VirtualElement as well. This will lead to type errors in cases where I would like to use a default method of HTMLElement on the element ref:

```ts
const floating = useFloating(/* options */);
const buttonElement = floating?.refs.reference.current;
if (buttonElement?.contains(event.target as HTMLElement)) {}
// Property 'contains' does not exist on type 'Element | VirtualElement'.
//   Property 'contains' does not exist on type 'VirtualElement'. ts(2339)
```

This can be worked around by type casting `buttonElement`, but that isn't a very optimal solution:

```ts
const buttonElement = floating?.refs.reference.current as HTMLElement | null;
```

It would be great if we could pass a more specific type for the reference element via a generic prop, which should eliminate the need for type casting:

```ts
const floating = useFloating<HTMLElement>(/* options */);
const buttonElement = floating?.refs.reference.current; // This will automatically be HTMLElement | null
if (buttonElement?.contains(event.target as HTMLElement)) {}
```

mihkeleidast:
Resolves #1612.

Not sure if there is a better way to implement this, it needs to be passed through quite a few types to be properly type-safe.

Also not sure if or how to document this, it doesn't seem like the types are documented in the website. Let me know what would be the most correct place to add something and I will.

### DIFF
--- a/packages/dom/test/functional/arrow.test.ts
+++ b/packages/dom/test/functional/arrow.test.ts
@@ -7,6 +7,7 @@ allPlacements.forEach((placement) => {
     page,
   }) => {
     await page.goto('http://localhost:1234/arrow');
+    await page.waitForSelector(`[data-testid="placement-${placement}"]`);
     await click(page, `[data-testid="placement-${placement}"]`);
 
     expect(await page.locator('.container').screenshot()).toMatchSnapshot(
--- a/packages/react-dom-interactions/src/hooks/useFocusTrap.ts
+++ b/packages/react-dom-interactions/src/hooks/useFocusTrap.ts
@@ -6,6 +6,8 @@ import {getDocument} from '../utils/getDocument';
 import {isElement, isHTMLElement} from '../utils/is';
 import {stopEvent} from '../utils/stopEvent';
 import {useLatestRef} from '../utils/useLatestRef';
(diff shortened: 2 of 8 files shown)


atomiks:
> Also not sure if or how to document this, it doesn't seem like the types are documented in the website. Let me know what would be the most correct place to add something and I will.

Under the `Testing` header on `pages/react-dom` you could add a `Types` header and mention this generic and why it exists

mihkeleidast:
Cheers! Added a couple of sentences and an example.

atomiks:
There seems to be a type error when narrowing it in `react-dom-interactions` package when passing the context to each hook

Screen Shot 2022-04-12 at 2 49 49 pm

mihkeleidast:
Thanks for testing! I went through the interaction hooks and made them generic as well, this seems to fix those errors.

atomiks:
@mihkeleidast thank you! Can you add a test file for this? I think there's an `index.d.tsx` or something outside of `src` that has the type tests (in `react-dom` package at least). Writing a bunch of examples and using `@ts-expect-error` is what I've used to ensure the types don't error

PR #1619, opened 2022-04-12
Do not close floating element on blur when focus goes to nested floating elements within a portal

sventschui:
Do not close floating element on blur when focus goes to nested floating elements within a portal


atomiks:
Thanks for the PR @sventschui, do you have a demo/sandbox of the problem so I have some context?

sventschui:
Sorry for not providing any further info, I was in a rush.  The thing is, if you have nested portal'ed floating elements and the inner one contains a focusable element (i.e. input field), then the outer one will be closed, due to loosing focus. The input field is neither within the DOM tree of the reference nor of the floating element (of the outer floating element).

atomiks:
I feel like I should've encountered this  but I'm not sure why I didn't. It makes sense regardless. I'll merge, thanks!~

PR #1621, opened 2022-04-13
fix(dom): Prevent crash in browsers without ShadowRoot support

pleunv:
ShadowRoot support was only implemented as of [Modern Edge](https://caniuse.com/mdn-api_shadowroot), so any Legacy Edge version (including the latest, v18) currently suffers a crash when invoking `isShadowRoot()`. I'm aware Legacy Edge support is probably not something that is actively being chased but @floating-ui seems to work just fine apart from this small hitch. Would be great if you'd consider just popping this fix in.

This is basically the same issue that also occurred and was fixed back in the day in popper by @eps1lon, see #1253.

### DIFF
--- a/packages/dom/src/utils/is.ts
+++ b/packages/dom/src/utils/is.ts
@@ -24,6 +24,11 @@ export function isNode(value: any): value is Node {
 }
 
 export function isShadowRoot(node: Node): node is ShadowRoot {
+  // Legacy Edge doesn't support ShadowRoot.
+  if (typeof ShadowRoot === 'undefined') {
+    return false;
+  }
+
   const OwnElement = getWindow(node).ShadowRoot;
   return node instanceof OwnElement || node instanceof ShadowRoot;
 }
--- a/packages/dom/src/utils/is.ts
+++ b/packages/dom/src/utils/is.ts
@@ -24,7 +24,7 @@ export function isNode(value: any): value is Node {
 }
 
 export function isShadowRoot(node: Node): node is ShadowRoot {
-  // Legacy Edge doesn't support ShadowRoot.
+  // Browsers without `ShadowRoot` support
   if (typeof ShadowRoot === 'undefined') {
     return false;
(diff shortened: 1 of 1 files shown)


FezVrasta (approved):

atomiks:
I know this is small, but given this lib is so new, I removed a lot of these checks intentionally.

Legacy Edge's global usage is tiny, 1/6th of IE11 (which is intentionally not supported either)

Screen Shot 2022-04-14 at 12 29 47 am

pleunv:
> I know this is small, but given this lib is so new, I removed a lot of these checks intentionally.
> 
> Legacy Edge's global usage is tiny, 1/6th of IE11 (which is intentionally not supported either)

Yeah, I figured. We've recently dropped IE11 and moved our baseline up to Edge 16, although we're only actively testing in Edge 18. Granted, that's an enterprise context. Normally I'd just resort to patch-package and call it a day, but I'm mostly worried about the trickle-down effect, given that @floating-ui (just like popper before) leans itself very well towards usage in a variety of libraries, which would then also constrain those to everything-except-legacy-edge.

Unlike `ResizeObserver`, `ShadowRoot` also can't really easily be polyfilled as far as I'm aware. Plus it wouldn't make a lot of sense since @floating-ui is basically only checking it, not really using it. 

Anyway, definitely understand it if you'd rather not include this in the core.

atomiks:
> Unlike ResizeObserver, ShadowRoot also can't really easily be polyfilled as far as I'm aware. Plus it wouldn't make a lot of sense since https://github.com/floating-ui is basically only checking it, not really using it.

That's true, but can't they just do `window.ShadowRoot = window.ShadowRoot || new Function`? If they _are_ using web components they'd be using a proper polyfill (given they need to support old browsers), whereas this is just to support this check. The same can be said for syntax transpilation

atomiks:
Might as well just merge it lol

PR #1631, opened 2022-04-17
feat(useListNavigation): allowEscape prop

atomiks:


website/pages/docs/useListNavigation.mdx
```diff
@@ -171,6 +172,21 @@ useListNavigation(context, {
 > Your virtually-focused list items must have a unique
 > `id{:.keyword}` set on them.
 
+### allowEscape
+
+Determines whether focus can escape the list, such that nothing
+is selected after navigating beyond the boundary of the list. In
+some autocomplete/combobox components, this may be desired.
+
+> `loop{:.objectKey}` must be `true{:js}`
```
atomiks, line 181: I think this is an artificial limitation but looping should probably be enabled anyway for this. Can investigate non-loop support later.

PR #1649, opened 2022-04-24
fix(react-dom-interactions): shadow dom support

### LINKED ISSUE #1648: [React DOM Interactions] FloatingPortal and shadowRoot
In my project, part of the interface works inside shadowRoot.
I'm adding some components using Floating UI and encountered the following behavior:

`FloatingPortal` can't use an existing `div#floating-ui-root` on the page if it's inside a shadowRoot (`document.getElementById("floating-ui-root")` returns `null`).
`FloatingPortal` then creates a `div#floating-ui-root` and adds it to `document.body`.

I would like to be able to use a container inside the shadowRoot.
I understand that using floating components inside a shadowRoot is not a common scenario.
I understand that FloatingPortal can easily be re-implemented in a specific project with all the specifics in mind.

Does it make sense to think about a solution within @floating-ui/react-dom-interactions or is that out of scope?

[sandboxed example](https://codesandbox.io/s/floating-ui-floatingportal-shadowroot-4oqygg?file=/src/index.js)

atomiks:
Closes #1648 

cc: @BATCOH does a `rootId` prop for `<FloatingPortal />` seem sound?

### DIFF
--- a/packages/react-dom-interactions/src/FloatingFocusManager.tsx
+++ b/packages/react-dom-interactions/src/FloatingFocusManager.tsx
@@ -2,6 +2,7 @@ import {hideOthers} from 'aria-hidden';
 import React, {useCallback, useEffect, forwardRef, cloneElement} from 'react';
 import {useFloatingTree} from './FloatingTree';
 import type {FloatingContext, ReferenceType} from './types';
+import {contains} from './utils/contains';
 import {getChildren} from './utils/getChildren';
 import {getDocument} from './utils/getDocument';
 import {isElement, isHTMLElement} from './utils/is';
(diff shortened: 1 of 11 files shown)


BATCOH:
Yep, looks great

atomiks:
@sventschui does the `.contains()` util handle the same thing as your `isEventTargetInside`?

sventschui:
Not really I think.

You pass `event.target` to your custom contains implementation, what does not work out for us.

When there is a shadowRoot between the element the event listener is attached to (i.e. the `document`) and the element emitting the event (i.e. where the user clicked on), then `event.target` will point to the first/top-most element having a shadowRoot between the element the event listener is attached to (document) and the element clicked on. `event.composedPath()` is the only way to get the "real" target of the event, even when it is nested within a shadowRoot.

In our case we have a custom element (with shadowRoot) wrapping our entire application (incl. portal root), what leads to `event.target` of the `onPointerDown` handler inside `useDismiss` always being set to that custom element (instead of the element actually clicked on). This will lead to `useDismiss` thinking that every click is outside the floating element as the custom element is wrapping the floating element.

atomiks:
@sventschui can you add the `<FloatingPortal />` portal changes here into your PR? I'll merge it instead. `.contains()` I guess isn't necessary in this case?

sventschui:
I think the `contains` is still necessary. I fiddled in this CSB: https://codesandbox.io/s/billowing-sunset-o5g4cz?file=/src/index.ts

The first line of inputs is inside the first wrapper custom element, the second line of input is in the second wrapper custom element. If you now imagine that the second custom element is a portaled floating element nested inside the first custom element, then things would go wrong as `is inside wrapper2?` in the console is always false. The `e.relatedTarget` of the `focusout` event emitted on the first custom element will always reference the second custom element (wrapping the portaled floating element). Not sure whether I'm mixing up stuff up I started adding tests for this but did not yet get to the nested portaling use case yet.

PR #1650, opened 2022-04-24
Improve support for shadow dom / custom elements

sventschui:
When using shadow DOM

- events of event listeners attached to the `document`/`window` object will have the first element with a shadow root as their `event.target`. This currently causes issues with a bunch of "contains" checks. To work around this, `event.composedPath()` should be used, as it will give the complete path of the event, including elements contained in an (open) shadow root.
- `document.activeElement` will not traverse into shadow roots, one has to manually traverse into (open) shadow roots to find the actually focused element within a shadow root.

### DIFF
--- a/packages/react-dom-interactions/src/FloatingPortal.tsx
+++ b/packages/react-dom-interactions/src/FloatingPortal.tsx
@@ -21,7 +21,11 @@ export const FloatingPortal = ({
   const portalRef = React.useRef<HTMLElement | null>(null);
 
   useLayoutEffect(() => {
-    const rootNode = root ?? document.getElementById(id);
+    if (root) {
+      return;
+    }
+
+    const rootNode = document.getElementById(id);
 
     if (rootNode) {
       portalRef.current = rootNode;
@@ -30,12 +34,16 @@ export const FloatingPortal = ({
       portalRef.current.id = id;
     }
 
-    if (!root && !document.body.contains(portalRef.current)) {
+    if (!document.body.contains(portalRef.current)) {
       document.body.appendChild(portalRef.current);
     }
 
     setMounted(true);
   }, [id, root]);
+
+  if (root) {
+    return createPortal(children, root);
+  }
 
   if (mounted && portalRef.current) {
     return createPortal(children, portalRef.current);
--- a/packages/react-dom-interactions/src/FloatingFocusManager.tsx
+++ b/packages/react-dom-interactions/src/FloatingFocusManager.tsx
@@ -2,6 +2,7 @@ import {hideOthers} from 'aria-hidden';
 import * as React from 'react';
 import {useFloatingTree} from './FloatingTree';
 import type {FloatingContext, ReferenceType} from './types';
+import {activeElement} from './utils/activeElement';
 import {getChildren} from './utils/getChildren';
 import {getDocument} from './utils/getDocument';
 import {isElement, isHTMLElement} from './utils/is';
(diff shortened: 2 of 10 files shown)


atomiks:
I actually just opened a PR for this: #1649. I wasn't aware of the `.activeElement` thing though so I'll also have to add that

sventschui:
Quoting myself from #1649

> I think the contains is still necessary. I fiddled in this CSB: https://codesandbox.io/s/billowing-sunset-o5g4cz?file=/src/index.ts

> The first line of inputs is inside the first wrapper custom element, the second line of input is in the second wrapper custom element. If you now imagine that the second custom element is a portaled floating element nested inside the first custom element, then things would go wrong as is inside wrapper2? in the console is always false. The e.relatedTarget of the focusout event emitted on the first custom element will always reference the second custom element (wrapping the portaled floating element). Not sure whether I'm mixing up stuff up I started adding tests for this but did not yet get to the nested portaling use case yet.

packages/react-dom-interactions/src/FloatingPortal.tsx
```diff
@@ -11,15 +11,19 @@ const DEFAULT_ID = 'floating-ui-root';
...
+  rootId?: string | null;
 }): React.ReactPortal | null => {
   const [mounted, setMounted] = useState(false);
   const portalRef = useRef<HTMLDivElement | null>(null);
 
   useLayoutEffect(() => {
-    const root = document.getElementById(id) as HTMLDivElement | null;
+    const rootNode = rootId ? document.getElementById(rootId) : null;
+    const root = (rootNode?.querySelector(`#${id}`) ??
+      document.getElementById(id)) as HTMLDivElement | null;
```
BATCOH, line 26: I tested this code today, and it looks like it needs one small addition
```suggestion
    const root = (rootNode?.shadowRoot?.querySelector(`#${id}`) ??
      rootNode?.querySelector(`#${id}`) ??
      document.getElementById(id)) as HTMLDivElement | null;
```

atomiks:
@sventschui that's better I think yeah. I updated the file and docs, it should be good to go?

atomiks:
Actually I guess the `root` shouldn't be in the effect since it's already an `HTMLElement`, we can use it in the render phase directly

sventschui:
Rebased against master, LGTM for me

atomiks:
The `FloatingPortal` changes are gone?

Also `useListNavigation` has one `.activeElement` usage still there

sventschui:
The changes of FloatingPortal are already on master (see https://github.com/floating-ui/floating-ui/blob/master/packages/react-dom-interactions/src/FloatingPortal.tsx#L18)

I'll fixe the  `activeElement` and one `event.target` inside `safePolygon` I just found

atomiks:
Hmm master has the old version. New one shouldn't have it inside the effect, just:

```js
if (root) {
  return createPortal(children, root);
}
```

Is `event.relatedTarget` also a problem? Further, `useFocus` has targets in this form: `dataRef.current.openEvent?.target` - the same problem?

sventschui:
I pushed the other version of FloatingPortal. But note that `root` is still a dependency of the layout effect. If you'd like to get it out of the dependency array, then we'd need to set a `const hasRoot = root != null` in the function body and use that flag as a dependency instead.

Whether `event.relatedTarget` is an issue can not be answered with a clear yes or no, that really depends on a case-by-case basis. The thing is, there is no "composedPath"-like alternative to it. So if it is an issue, then we need to find a different way to support shadow DOM in these cases.

I'll take a look at useFocus, give me a second.

atomiks:
I'm going to delete `useFocusTrap` (replaced by `FloatingFocusManager`) in this next minor version so that's no problem. I guess this is good to go then 👍 if there are any reports of issues new PRs don't hurt

PR #1666, opened 2022-04-30
feat: add "browser" export maps for dom and core packages

Westbrook:
Expand export map approach to better surface browser capable exports.

Do you see this as needing additional documentation? We added some for the non-`browser` exports and it might get confusing to feature the inner working of both sides of the export map, at least in the main site. If there's a good "sub-section" location to include something like this, happy to be pointed in the right direction.

In #1538 we didn't really discuss if there should be any reverting/simplifying of the previous change, so if you have any thoughts there, I'd also be happy to touch on that before this is ready to merge.

Sorry for taking so long to get back to this, but excited to help get things closer to 1.0!


Westbrook:
Also, would you prefer this to be tracked as a PATCH change or a MINOR change? Pre-1.0 it can be hard to decide the right way to apply semver.

atomiks:
Thanks, have you confirmed doing `nodeResolve({ browser: true })` selects the production version of that key with no issues? In practice does selecting `browser: true` have negative effects on any other packages?

Westbrook:
Teams I work with generally use `browser: true` almost 100% of the time with no negative effects on other libraries that I have heard to date.

Copying these changes into a local install of our components project works great with these settings. I would prefer to have `canary` channel release for further testing, but in a pre-1.0 state that might not be as necessary as a release with follow up releases as needed with fixes.

atomiks:
Since we're still v0, we might as well revert the previous change and then instead document that `browser: true` is the better option for these users (plus the optional `development` key to aid development)?

This would make this a MINOR change

Westbrook:
Let me know how this update to the docs suits you. I struggle a bit to fully understand both usage paths in order to represent them fairly here.

atomiks:
Tried to simplify the docs a bit

Westbrook:
LGTM.

atomiks:
@Westbrook I had to change the package's entry points to solve a CJS issue. I'm not sure how this now works with this PR

Westbrook:
I’ll rebase and see what I can see.

atomiks:
Woops I just did. I think the issue is `.mjs` is now unminified/not production file.

atomiks:
^ I think these changes should solve all the issues. I renamed the files:

- `./dist/floating-ui.core.umd.js` `main`
- `./dist/floating-ui.core.umd.min.js` `unpkg`
- `./dist/floating-ui.core.esm.js` `module`
- `./dist/floating-ui.core.mjs` `import`
- `./dist/floating-ui.browser.min.mjs`  import default + `browser`
- `./dist/floating-ui.browser.mjs` import default + `browser` w/ logging

atomiks:
Tested with Rollup and it works with the `browser` conditions. 

I'm going to revert the README change then add it back before release (hopefully soon...)

PR #1669, opened 2022-05-04
fix(FloatingFocusManager): allow `returnFocus` when `preventTabbing` is true

atomiks:


atomiks:
Not desirable with nested list nav. Updated the docs to be clearer: https://github.com/floating-ui/floating-ui/commit/0978c10ac21268941c5bba5c1e47f87f661c1fcf

PR #1672, opened 2022-05-04
feat: add useFloatingPortalNode hook

### LINKED ISSUE #1668: [React DOM Interactions] Expose `useFloatingPortalNode` hook
I'm working through some other dependencies that also do portaling (right now dealing specifically with [react-modal](https://reactcommunity.org/react-modal/)). I'd like to use the same portal container for all the different dependencies, but I can't use the FloatingPortal component itself directly, as those dependencies have their own portaling logic. However, some of them expose props to customize the portaling element.

I'd like to extract most of the FloatingPortal component code that deals with injecting the portal container to a custom hook. This way I could use the hook with other components that accept a reference to a node.

Example code of what I'd like to achieve with react-modal:
```jsx
const App = () => {
  const portalNode = useFloatingPortalNode();

  return <ReactModal parentSelector={() => portalNode}></ReactModal>;
};
```

Does this feature make sense? I can help with a PR, if necessary.

mihkeleidast:
Resolves #1668.

### DIFF
--- a/packages/react-dom-interactions/src/FloatingPortal.tsx
+++ b/packages/react-dom-interactions/src/FloatingPortal.tsx
@@ -52,8 +52,11 @@ export const FloatingPortal = ({
   const portalNode = useFloatingPortalNode({id, enabled: !root});
 
   useLayoutEffect(() => {
+    if (root) {
+      return;
+    }
     setMounted(true);
-  }, []);
+  }, [root]);
 
   if (root) {
     return createPortal(children, root);
--- a/packages/react-dom-interactions/src/FloatingPortal.tsx
+++ b/packages/react-dom-interactions/src/FloatingPortal.tsx
@@ -4,24 +4,17 @@ import useLayoutEffect from 'use-isomorphic-layout-effect';
 
 const DEFAULT_ID = 'floating-ui-root';
 
-/**
- * Portals your floating element outside of the main app node.
- * @see https://floating-ui.com/docs/FloatingPortal
- */
-export const FloatingPortal = ({
-  children,
+export const useFloatingPortalNode = ({
   id = DEFAULT_ID,
-  root = null,
+  enabled = true,
 }: {
-  children?: React.ReactNode;
   id?: string;
-  root?: HTMLElement | null;
-}): React.ReactPortal | null => {
-  const [mounted, setMounted] = React.useState(false);
+  enabled?: boolean;
+} = {}) => {
   const portalRef = React.useRef<HTMLElement | null>(null);
 
   useLayoutEffect(() => {
-    if (root) {
+    if (!enabled) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


packages/react-dom-interactions/src/FloatingPortal.tsx
```diff
@@ -37,16 +30,37 @@ export const FloatingPortal = ({
...
+  id?: string;
+  root?: HTMLElement | null;
+}): React.ReactPortal | null => {
+  const [mounted, setMounted] = React.useState(false);
+  const portalNode = useFloatingPortalNode({id, enabled: !root});
+
+  useLayoutEffect(() => {
     setMounted(true);
-  }, [id, root]);
+  }, []);
```
atomiks, line 56: This will cause another render for `root` prop unnecessarily, should have an early return

website/pages/docs/FloatingPortal.mdx
```diff
@@ -49,3 +49,24 @@ supports shadow roots.
...
+```js
+const App = () => {
+  const portalNode = useFloatingPortalNode({
+    id: 'floating-ui-root',
+    enabled: true,
+  });
+
+  if (portalNode) {
+    return React.createPortal(<div></div>, portalNode);
+  }
```
atomiks, line 68: `ReactDOM.createPortal()`

website/pages/docs/FloatingPortal.mdx
```diff
@@ -49,3 +49,24 @@ supports shadow roots.
...
 ```
+
+## useFloatingPortalNode
+
+Exposes the portal container node for custom use in other
+components. The hook injects the node, if it does not yet exist
+in the DOM.
+
+```js
+const App = () => {
```
atomiks, line 60: `function App() {`

packages/react-dom-interactions/src/FloatingPortal.tsx
```diff
@@ -4,24 +4,17 @@ import useLayoutEffect from 'use-isomorphic-layout-effect';
...
-/**
- * Portals your floating element outside of the main app node.
- * @see https://floating-ui.com/docs/FloatingPortal
- */
-export const FloatingPortal = ({
-  children,
+export const useFloatingPortalNode = ({
   id = DEFAULT_ID,
-  root = null,
+  enabled = true,
```
atomiks, line 9: The `enabled` prop is a little odd for this hook but since you can't conditionally render hooks and prevent the effect from running I guess it makes sense for `<FloatingPortal />` use case

website/pages/docs/FloatingPortal.mdx
```diff
@@ -49,3 +49,24 @@ supports shadow roots.
...
+  const portalNode = useFloatingPortalNode({
+    id: 'floating-ui-root',
+    enabled: true,
+  });
```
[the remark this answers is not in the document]
mihkeleidast: Added some inline comments to these lines and changed them to be custom - this pattern seems to be used elsewhere as well. Is this okay or should I document the hook props in separate blocks?

packages/react-dom-interactions/src/FloatingPortal.tsx
mihkeleidast: Yeah, I'm also not aware of a different way to disable a hook conditionally.

mihkeleidast:
Thanks for the review, added some fixes!

PR #1680, opened 2022-05-06
add .current to useEffect dependencies of React DOM doc

maloguertin:
Without it the hook is never called again even if the refs update.

### DIFF
--- a/website/pages/docs/react-dom.mdx
+++ b/website/pages/docs/react-dom.mdx
@@ -140,7 +140,7 @@ function App() {
       refs.floating.current,
       update
     );
-  }, [refs.reference, refs.floating, update]);
+  }, [refs.reference.current, refs.floating.current, update]);
 
   return (
     <>


FezVrasta:
That example seems off to me, when refs values are changed React doesn't trigger a re-render, so that effect won't be called anyways 🤔

atomiks:
Usually you have an `open` dependency that will cause it to run at the correct time:

```js
useEffect(() => {
  if (open) {
    return autoUpdate(...);
  }
}, [open, refs.reference, refs.floating, update]);

return open && <div ref={floating}>floating</div>
```

But you are right that if `reference` changes (or some other case) the effect fails to run. Specifying `.current` as a dependency is weird though (and is documented for the external reference, but I think that pattern is bad anyway)

FezVrasta:
refs should not be included in the dependencies list, but the eslint rule is not very smart and doesn't understand if a ref is being forwarded from some other hook.

atomiks:
@FezVrasta yeah it's just because of the rule not because it's necessary, in fact it causes additional confusion and would be more obvious without them there :\

atomiks:
@maloguertin I opened a PR for a new callback which would solve this better and makes it simpler: https://github.com/floating-ui/floating-ui/pull/1683 

Let me know what you think.

maloguertin:
> @maloguertin I opened a PR for a new callback which would solve this better and makes it simpler: #1683
> 
> Let me know what you think.

Perfect :100:

atomiks:
This will be released soon (`whileElementsMounted` prop)

PR #1683, opened 2022-05-06
feat(react-dom): `whileElementsMounted` callback

atomiks:
This introduces a new callback to solve issues with `useEffect` + `autoUpdate`

`useEffect` plus regular refs does not allow `autoUpdate` to react to elements changing:

- Specifying `.current` in the dependency array will have the lint rule warn, is not always reliable, and reading refs in render is not recommended.
- Using an `open` dependency works but not if the reference changed.
- It's verbose to use

This solves this by using a callback that gets invoked whenever **both** elements are mounted and ready, AND whenever they change. This allows `autoUpdate` to always be "fresh". This also allows the consumer to still use their own auto-updating logic if necessary (which is why it's not a prop like `autoUpdate: boolean | Options`.

The signature matches `autoUpdate`, so it's simple:

```js
useFloating({
  whileElementsMounted: autoUpdate,
});
```


maloguertin:
Before realising you gave the same signature for the callback as autoUpdate I was about to say you could have a settings like `autoUpdate: true` but your solution is really clean. :+1:

atomiks:
I think this should be called `whileElementsMounted`, because `on*` callbacks seem asymmetric without a cleanup function, aka fire-and-forget...

PR #1685, opened 2022-05-07
fix(react-native): arrows

atomiks:
Three distinct parts were making arrows not work in React Native

cc: @brycnguyen

Screen Shot 2022-05-07 at 11 04 24 am

### DIFF
--- a/packages/core/src/middleware/arrow.ts
+++ b/packages/core/src/middleware/arrow.ts
@@ -56,11 +56,15 @@ export const arrow = (options: Options): Middleware => ({
     const startDiff = coords[axis] - rects.reference[axis];
 
     const arrowOffsetParent = await platform.getOffsetParent?.(element);
-    const clientSize = arrowOffsetParent
+    let clientSize = arrowOffsetParent
       ? axis === 'y'
         ? arrowOffsetParent.clientHeight || 0
         : arrowOffsetParent.clientWidth || 0
       : 0;
+
+    if (clientSize === 0) {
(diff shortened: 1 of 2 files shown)


packages/core/src/middleware/arrow.ts
```diff
@@ -56,12 +56,16 @@ export const arrow = (options: Options): Middleware => ({
...
-    const clientSize = arrowOffsetParent
+    let clientSize = arrowOffsetParent
       ? axis === 'y'
         ? arrowOffsetParent.clientHeight || 0
         : arrowOffsetParent.clientWidth || 0
       : 0;
 
+    if (clientSize === 0) {
+      clientSize = rects.floating[length];
+    }
```
atomiks, line 67: There's no `offsetParent` method in React Native, so we can just assume the offsetParent is the floating element itself.

packages/react-native/src/createPlatform.ts
```diff
@@ -67,9 +67,9 @@ export const createPlatform = ({
       }
     });
   },
-  getDimensions: ({element}) =>
+  getDimensions: (element) =>
     new Promise((resolve) =>
-      element.measure(({width, height}: DimensionsType) =>
+      element.measure((x: number, y: number, width: number, height: number) =>
         resolve({width, height})
       )
```
atomiks, line 74: The signatures of this fn was entirely wrong 😓

brycnguyen:
Verified that it works! Thanks!

PR #1703, opened 2022-05-16
fix: Fix README.md top banner size

sudosubin:
The README.md top banner ratio is broken on narrow devices (or narrow window tab).

| before | after |
| ------ | ----- |
| image | image |


atomiks:
Thanks, somehow didn't ever check it with a narrow window size lol

PR #1715, opened 2022-05-24
fix: save portal node to state instead of ref

mihkeleidast:
I ran into an issue with the current implementation. Essentially, when the portal node ref is passed into another component, it is `null` initially. Since it is saved into a ref, it won't trigger a re-render when it changes from `null` to `HTMLElement`, which causes the component not to render at all or render into a wrong node until it rerenders for some other reason.

An example, maybe it is helpful - in this case, ModalInner will never be rendered:
```
export const Modal: React.FunctionComponent<Props> = (props) => {
  const portalNode = useFloatingPortalNode();

  if (!portalNode) {
    return null;
  }

  // portalNode is required
  return <ModalInner {...props} portalNode={portalNode} />;
};
```

Happy to discuss other ways as well how this could be improved - it was easier to file this simple PR to start :)

### DIFF
--- a/packages/react-dom-interactions/src/FloatingPortal.tsx
+++ b/packages/react-dom-interactions/src/FloatingPortal.tsx
@@ -49,21 +49,13 @@ export const FloatingPortal = ({
   id?: string;
   root?: HTMLElement | null;
 }): React.ReactPortal | null => {
-  const [mounted, setMounted] = React.useState(false);
   const portalNode = useFloatingPortalNode({id, enabled: !root});
-
-  useLayoutEffect(() => {
-    if (root) {
-      return;
-    }
-    setMounted(true);
-  }, [root]);
 
   if (root) {
     return createPortal(children, root);
   }
 
-  if (mounted && portalNode) {
+  if (portalNode) {
     return createPortal(children, portalNode);
   }
 
--- a/packages/react-dom-interactions/src/FloatingPortal.tsx
+++ b/packages/react-dom-interactions/src/FloatingPortal.tsx
@@ -11,7 +11,7 @@ export const useFloatingPortalNode = ({
   id?: string;
   enabled?: boolean;
 } = {}) => {
-  const portalRef = React.useRef<HTMLElement | null>(null);
+  const [portalRef, setPortalRef] = React.useState<HTMLElement | null>(null);
 
   useLayoutEffect(() => {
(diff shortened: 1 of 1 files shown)


atomiks:
I think I noticed that too, `<FloatingPortal />` mitigates the problem by calling `setMounted`.

I was experimenting with a transient node rather than a permanent one which may be better so it doesn't stick around on the DOM permanently, what do you think? 

The issue is the node needs to be ready on first render because the interaction hooks' effects run their logic in `useLayoutEffect` as soon as `open` is true, and can't wait for another render...

```js
open && <FloatingPortal />
```

mihkeleidast:
Would it then inject the node immediately before it is necessary, and remove it after the last child node is removed? How would it know that there aren't any other children rendered into that portal?

atomiks:
Each `FloatingPortal` component creates its own node, so it doesn't need to know about other children

atomiks:
Each portal node would only get appended to the DOM when it's `open`, so elements appended later with the same `z-index` would appear over earlier ones anyways. That makes it work the same as the current singular node, where they get appended to the end of the node

mihkeleidast:
Found the single root perf article - not sure it is relevant anymore though, as it was related to an old Popper version I think, but you can probably evaluate it better than me: https://atfzl.com/don-t-attach-tooltips-to-document-body

atomiks:
Ah I remember that article 👍 I guess it depends if attaching the portal node to the body dynamically based on `open` state, with the tooltip inside of it (not the body), has a perf impact compared to it being always on the body after it's first rendered.

packages/react-dom-interactions/src/FloatingPortal.tsx
```diff
@@ -11,7 +11,7 @@ export const useFloatingPortalNode = ({
   id?: string;
   enabled?: boolean;
 } = {}) => {
-  const portalRef = React.useRef<HTMLElement | null>(null);
+  const [portalRef, setPortalRef] = React.useState<HTMLElement | null>(null);
```
atomiks, line 14: Can we change the name to `portalEl` instead?

atomiks:
Does `FloatingPortal` now no longer need `setMounted`?

mihkeleidast:
Yup, good point, does not seem to need it anymore 👍

PR #1727, opened 2022-05-26
Add missing peer dependencies

mskelton:
Yarn requires peer dependencies to be specified at each level of nesting if you have deeply nested packages with peer deps.  This adds the missing peer deps to react-dom-interactions since it depends on `@floating-ui/react-dom` which has peer deps.

This is what the error currently looks like that this will fix.

image


atomiks:
Idk how that was missing but thanks 😄

PR #1735, opened 2022-05-30
fix: correct return type for useFloatingPortalNode

mihkeleidast:
Without the explicit return type, the built TS definitions say that the return is always HTMLElement, which is not correct:

```ts
export declare const useFloatingPortalNode: ({ id, enabled, }?: {
    id?: string;
    enabled?: boolean;
}) => HTMLElement;
```

After this, the return type also includes `null`, as is correct.

An alternative solution would be to set `"strict": true` in tsconfig, then the null is added by TS - that might be the better solution overall, as it would add the missing nulls to other potential places as well, but it might also create more churn for users who haven't dealt with nulls..

### DIFF
--- a/.config/tsconfig.json
+++ b/.config/tsconfig.json
@@ -12,6 +12,7 @@
     "emitDeclarationOnly": true,
     "jsx": "react-jsx",
     "outDir": "../packages",
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "strict": true
   }
 }


atomiks:
The alternative solution sounds better because there may be other places in need of it (I previously added explicits but wasn't aware of that rule). The user not dealing with null seems like a type bug to me, anyway

mihkeleidast:
Alright, changed the solution then 👍

mihkeleidast:
Not sure if I should change the rolling versions myself or not, but currently it's set to only publish react-dom-interactions. I think this should publish all the packages, as the global tsconfig change can affect the types of all of them?

atomiks:
@mihkeleidast good point, let me update

PR #1739, opened 2022-06-06
Adds requestAnimationFrame to new ResizeObserver

### LINKED ISSUE #1569: What's a stable / unstable reference?
Can't find an explanation of a stable / unstable reference mentioned here:

https://floating-ui.com/docs/react-dom#stable-ref-prop

I'm getting the following error (sometimes, not always): `ResizeObserver loop limit exceeded` while using Floating UI w/ conditional rendered React Portals.

csantos1113:
Fixes #1569

Wrap callback call in requestAnimationFrame to avoid error: "ResizeObserver loop limit exceeded"
ref: https://stackoverflow.com/a/58701523

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -68,7 +68,17 @@ export function autoUpdate(
 
   let observer: ResizeObserver | null = null;
   if (elementResize) {
-    observer = new ResizeObserver(update);
+    observer = new ResizeObserver((props) => {
+      // Wrap callback call in requestAnimationFrame to avoid error: "ResizeObserver loop limit exceeded"
+      // https://stackoverflow.com/a/58701523
+      // REF:
+      // - https://github.com/WICG/resize-observer/issues/38
+      // - https://github.com/w3c/csswg-drafts/issues/5023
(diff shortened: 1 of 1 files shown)


FezVrasta (approved):

atomiks:
Doesn't this make the update run after paint? That makes this solution unusable

atomiks:
[You can see it runs after paint here which results in flickering](https://codesandbox.io/s/elegant-bas-8dqo07?file=/src/index.js), as a result this is not a viable solution, you should ignore/suppress the error in your app instead

csantos1113:
@atomiks thanks for looking into this so quickly.

How do you get around the "ResizeObserver loop limit exceeded" error then? when the element is close to the edge of the window (hence it resizes), the ResizeObserver loop limit exceeded is thrown.

Right now we have DataDog configured, so this is filling out out logger quota

atomiks:
I've never seen that error when it's near the edge of the viewport. Does it happen in incognito mode? Seems like extensions could interfere based on the SO thread?

You could disable the elementResize option and update manually when content changes.

FezVrasta:
With Sentry there's an `ignoreErrors` list that can be used to ignore errors, maybe with DataDog there's something similar.

PR #1752, opened 2022-06-22
fix: prevent unwanted scrolling when dismissing via outside pointerdown

### LINKED ISSUE #1751: [React DOM Interactions - FloatingFocusManager] Unexpected page scroll when clicking from one popover trigger to another.
When using `FloatingFocusManager` and `useDismiss` together:

If an open popover is off-screen, and a trigger is clicked to open a new one, the page will scroll back to the previous element and instantly close it, instead of opening the new popover.

ZplCgrduAq

Steps:
- Click on the first button on the page to open a popover.
- While the popover is still open, scroll down until the first is off the page, then click any other button.

Repro:
https://codesandbox.io/s/floating-ui-react-dom-visual-test-forked-rwogd6

atomiks:
Closes #1751  

cc @threehams

### DIFF
--- a/packages/react-dom-interactions/src/FloatingFocusManager.tsx
+++ b/packages/react-dom-interactions/src/FloatingFocusManager.tsx
@@ -59,7 +59,7 @@ export interface Props<RT extends ReferenceType = ReferenceType> {
  * @see https://floating-ui.com/docs/FloatingFocusManager
  */
 export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
-  context: {refs, nodeId, onOpenChange, dataRef},
+  context: {refs, nodeId, onOpenChange, dataRef, events},
   children,
   order = ['content'],
   endGuard = true,
@@ -230,6 +230,7 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
       return;
     }
 
+    let returnFocusValue = returnFocus;
     const floating = refs.floating.current;
     const previouslyFocusedElement = activeElement(getDocument(floating));
 
@@ -239,12 +240,30 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
       focus(initialFocus.current ?? floating);
     }
 
+    // Dismissing via outside `pointerdown` should always ignore `returnFocus`
+    // to prevent unwanted scrolling. The `esc` key will continue to focus the
+    // reference.
+    function onDismiss() {
+      returnFocusValue = false;
+    }
+
+    events.on('dismiss', onDismiss);
+
     return () => {
-      if (returnFocus && isHTMLElement(previouslyFocusedElement)) {
+      events.off('dismiss', onDismiss);
+
+      if (returnFocusValue && isHTMLElement(previouslyFocusedElement)) {
         focus(previouslyFocusedElement);
       }
     };
-  }, [preventTabbing, getTabbableElements, initialFocus, returnFocus, refs]);
+  }, [
+    preventTabbing,
+    getTabbableElements,
+    initialFocus,
+    returnFocus,
+    refs,
+    events,
+  ]);
 
   const isTypeableCombobox = () =>
     refs.domReference.current?.getAttribute('role') === 'combobox' &&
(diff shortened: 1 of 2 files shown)


threehams:
The original issue is fixed when clicking outside the popover or trigger. However, I'm now seeing the page scroll to the top on the first click on any popover trigger which uses `FloatingFocusManager`. It focuses the `body` element.

This is on a much bigger project than the one on codesandbox (I don't know how to get this branch working on there), but removing the focus manager fixes it. If you can't reproduce, I can create a much smaller project and try it out there tomorrow.

atomiks:
Is `document.activeElement` pointing to the body or the close button? I think it may be focusing the close button before it gets positioned. So it's at the top-left of the screen (`top: 0`, `left: 0)` initially at the time of focus, so it scrolls to the top. 

However, the focus call is wrapped in rAF which gets called after it's positioned so that shouldn't happen (can't repro it either)

threehams:
`document.activeElement` is the close button if it exists. If I remove that, it focuses the floating element, but the same scroll problem happens.

I'll create a minimal repro tomorrow and see if I can make it happen there. Otherwise, I might have to debug in the big project. The [production page](https://www.truecar.com/used-cars-for-sale/listings/location-91403/?dealType=loan) doesn't use `floating-ui` right now, but that'll give you a sense of scale compared to the codesandbox one.

Thanks for jumping on this so quickly!

atomiks:
If you destructure the `x` value out of the context (line 62)

https://github.com/floating-ui/floating-ui/blob/1589c1dec67669c6d43989b58c7203437c7515d5/packages/react-dom-interactions/src/FloatingFocusManager.tsx#L61-L62

and check if it's `null` or not here:

https://github.com/floating-ui/floating-ui/blob/1589c1dec67669c6d43989b58c7203437c7515d5/packages/react-dom-interactions/src/FloatingFocusManager.tsx#L228-L231

```js
if (preventTabbing || x == null) {
  return
}
```

Does the problem occur?

Or if you wrap the `el?.focus()` line at the top of this file in another rAF wrapper?

Seems to be a timing problem for when the position is "ready".

threehams:
I created a minimal Next.js project to take out any codesandbox differences: https://github.com/threehams/floating-ui-popover-scroll
The same bug happens with `create-react-app`.

Exact steps I took to use the version in this PR:
- Clone `floating-ui/floating-ui` locally
- Check out latest `fix/no-focus-reference`
- `npm install`
- `npm build`
- Copy `dist` directory from `packages/react-dom-interactions` into `node_modules/@floating-ui/react-dom-interactions/` and overwrite existing files
- Delete `.next` directory (if it exists) to clear cache

Steps:
- Click the first button on the page (popover pops up)
- Scroll to the bottom of the page, click outside any button. The page doesn't scroll, so that part of the bug is fixed.
- Click the last button on the page. Page scrolls to the top.
- Scroll down and click last button again. Popover opens, and page doesn't scroll.
- Click second to last button. Page scrolls to the top.

This reproduces across a MacBook Pro and Windows desktop:
Chrome (Win + Mac), Firefox (Win + Mac), Edge (Win)
It does **not** happen in Safari (Mac)

threehams:
> If you destructure the `x` value out of the context (line 62)
> 
> https://github.com/floating-ui/floating-ui/blob/1589c1dec67669c6d43989b58c7203437c7515d5/packages/react-dom-interactions/src/FloatingFocusManager.tsx#L61-L62
> 
> and check if it's `null` or not here:
> 
> https://github.com/floating-ui/floating-ui/blob/1589c1dec67669c6d43989b58c7203437c7515d5/packages/react-dom-interactions/src/FloatingFocusManager.tsx#L228-L231
> 
> ```js
> if (preventTabbing || x == null) {
>   return
> }
> ```
> 
> Does the problem occur?

Adding that null check fixes the forced scroll! `x` was null in all the cases where it was happening.

> EDIT: Are you sure you didn't change anything else in the file? I don't see why this problem would only be occurring with the changes I listed. That side effect should be happening regardless of this change

It looks like the problem wasn't caused by any of your changes here - it was already there on release. On release there's always a scroll though, so it wasn't possible to test the second bug without fixing the first.

atomiks:
@threehams thanks, I found the cause 😄 

`autoUpdate` being present runs the update in a microtask (the `ResizeObserver`), but the position is _also_ in its own microtask (nested). This makes it run before `requestAnimationFrame`, causing the scroll to top behavior as it hasn't been positioned yet. 

I think this also caused a different issue somewhere but now I understand why.

I'll merge this fix in a sec and make a new PR for that bug

PR #1755, opened 2022-06-23
fix(autoUpdate): avoid initial update in microtask

atomiks:
Avoids putting the initial update in a microtask as this causes a timing problem with focus calls

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -68,7 +68,14 @@ export function autoUpdate(
 
   let observer: ResizeObserver | null = null;
   if (elementResize) {
-    observer = new ResizeObserver(update);
+    let initialUpdate = true;
+    observer = new ResizeObserver(() => {
+      if (!initialUpdate) {
+        update();
+      }
+
+      initialUpdate = false;
+    });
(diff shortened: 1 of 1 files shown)


threehams:
This change is causing an infinite loop on tooltip/popover open on my (closed-source) work project. I'm trying to create a minimal repro case, but this one seems more specific. So far, I've narrowed out framework (Next.js) and React version.

Any ideas what could cause this? I'll keep trying to track it down regardless.

tychenjiajun:
> This change is causing an infinite loop on tooltip/popover open on my (closed-source) work project. I'm trying to create a minimal repro case, but this one seems more specific. So far, I've narrowed out framework (Next.js) and React version.
> 
> Any ideas what could cause this? I'll keep trying to track it down regardless.

Check my review

atomiks:
Another lib had an infinite loop problem (not from this but an earlier change) because they had to `computePosition` / `autoUpdate` calls running on the same element in parallel, so they were conflicting with each other. Not sure if that's your case here though

atomiks:
Another lib had an infinite loop problem (not from this but an earlier change) because they had two `computePosition` / `autoUpdate` calls running on the same element in parallel, so they were conflicting with each other. Not sure if that's your case here though.

PR #1756, opened 2022-06-23
v1.x , support for IE9

WangZhekun:
support for IE9 in version 1.x


atomiks:
Thanks for the effort, but:

- Popper v1 was deprecated in late 2019/early 2020, we haven't released any new versions since then and will not.
- IE9 in *2022*? That's ridiculous, IE11 was just announced dead 🤣

PR #1761, opened 2022-06-28
fix(react-dom): allow unstable ref props

### LINKED ISSUE #1760: Using a button trigger which sets its own ref in state causes an infinite loop.
## Issue

If the trigger element for a tooltip/popover sets its own ref in internal state, the page locks up in an infinite loop.

## Reproduction

https://codesandbox.io/s/floating-ui-react-dom-infinite-loop-b8s1dh

## Expected

I'm not really sure which code is in the wrong here. React doesn't explicitly ban setting refs in state, and it's necessary for some patterns. That said, it's definitely not common. In this case, it was simple enough for me to take out the `setSelfRef` state, as it was part of an older render props API for the previous tooltip library.

That said, at wider scale, this may come up again, and this was difficult to track down.

atomiks:
closes #1760 @threehams

not sure why i didn't think of this originally 😄 can remove the docs about it too

### DIFF
--- a/packages/react-dom/src/useFloating.ts
+++ b/packages/react-dom/src/useFloating.ts
@@ -19,6 +19,8 @@ export function useFloating<RT extends ReferenceType = ReferenceType>({
 }: UseFloatingProps = {}): UseFloatingReturn<RT> {
   const reference = React.useRef<RT | null>(null);
   const floating = React.useRef<HTMLElement | null>(null);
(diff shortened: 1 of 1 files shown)


atomiks:
hmm seems like there could be a problem with allowing `null` to unset the ref value as the second commit reintroduces the loop

atomiks:
Seems like this realistically can't be fixed for the `whileElementsMounted` prop due to React calling it with null again 😒

Issue #1594, opened 2022-03-25
[DOM] Element position doesn't get computed properly when `iframe` gets scaled


seahindeniz:
I have an element inside an iframe, and I'm trying to compute its position from the parent window. However, when I give `transform: scale(2)` to iframe element, scale doesn't get counted.
https://codesandbox.io/s/floating-ui-dom-template-forked-gz4whh?file=/src/index.js

This is how the tooltip gets rendered:
<img width="402" alt="image" src="https://user-images.githubusercontent.com/5789670/160128684-b39ec960-fb47-470d-adca-37a5613c5f2f.png">

Instead, I was expecting something like this:
<img width="402" alt="image" src="https://user-images.githubusercontent.com/5789670/160128841-f7ad2c15-d93f-4e49-bc7c-d5f5cd1d4e0b.png">

seahindeniz:
@atomiks as a solution, I think if we can pass an external/additional scale value (like passing an offset value) (`{ scaleX: number, scaleY: number }`) somehow which can be passed to the [getBoundingClientRect](https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/utils/getBoundingClientRect.ts#L6) utility function to be included in the calculations, it could solve the problem. But I'm not 100% sure if this is a good approach, of course.

Something like this:
```ts
import type {ClientRectObject, VirtualElement} from '@floating-ui/core';
import {isElement, isHTMLElement, isLayoutViewport} from './is';
import {round} from './math';
import {getWindow} from './window';

type ExternalScaleType = {
  x: number,
  y: number,
}

export function getBoundingClientRect(
  element: Element | VirtualElement,
  includeScale = false,
  isFixedStrategy = false,
  externalScale?: ExternalScaleType
): ClientRectObject {
  const clientRect = element.getBoundingClientRect();

  let scaleX = 1;
  let scaleY = 1;
  const externalScaleX = externalScale?.x || 1;
  const externalScaleY = externalScale?.y || 1;

  if (includeScale && isHTMLElement(element)) {
    scaleX =
      element.offsetWidth > 0
        ? round(clientRect.width) / element.offsetWidth || 1
        : 1;
    scaleY =
      element.offsetHeight > 0
        ? round(clientRect.height) / element.offsetHeight || 1
        : 1;
  }

  const win = isElement(element) ? getWindow(element) : window;
  const addVisualOffsets = !isLayoutViewport() && isFixedStrategy;

  const x =
[...]

PR #1763, opened 2022-06-30
fix(dom): calc position correctly with css transform()

cvlmtg:
This PR should fix the positioning of elements when applying `transform: scale()` to any element or any of its parents.

it's based on https://github.com/floating-ui/floating-ui/issues/1594#issuecomment-1086821366  but instead of passing the scale as a parameter I'm using it always. The scale formula is taken from https://github.com/rfrey-rbx/popper.js/commit/add166ea75be347081990674f346031575b3dac3

before (with `scale(0.5)` applied to the container of the app):

Schermata 2022-06-30 alle 12 41 20

after:

Schermata 2022-06-30 alle 12 41 36

### DIFF
--- a/packages/dom/src/utils/getBoundingClientRect.ts
+++ b/packages/dom/src/utils/getBoundingClientRect.ts
@@ -27,14 +27,13 @@ export function getBoundingClientRect(
   const win = isElement(element) ? getWindow(element) : window;
   const addVisualOffsets = !isLayoutViewport() && isFixedStrategy;
 
-  const x =
-    (clientRect.left +
-      (addVisualOffsets ? win.visualViewport?.offsetLeft ?? 0 : 0)) /
-    scaleX;
-  const y =
-    (clientRect.top +
-      (addVisualOffsets ? win.visualViewport?.offsetTop ?? 0 : 0)) /
-    scaleY;
+  const x0 = clientRect.left +
+    (addVisualOffsets ? win.visualViewport?.offsetLeft ?? 0 : 0);
+  const y0 = clientRect.top +
+    (addVisualOffsets ? win.visualViewport?.offsetTop ?? 0 : 0);
+
+  const x = scaleX < 1 ? x0 * scaleX : x0 / scaleX;
+  const y = scaleY < 1 ? y0 * scaleY : y0 / scaleY;
   const width = clientRect.width / scaleX;
   const height = clientRect.height / scaleY;
 
--- a/packages/dom/src/utils/getClippingRect.ts
+++ b/packages/dom/src/utils/getClippingRect.ts
@@ -24,7 +24,7 @@ function getInnerBoundingClientRect(
 ): ClientRectObject {
   const clientRect = getBoundingClientRect(
     element,
-    false,
+    true,
     strategy === 'fixed'
   );
   const top = clientRect.top + element.clientTop;


atomiks:
Parent scale is already handled via the `includeScale` logic. What part does this fix, iframes? Can you add a test that fails without this change?

cvlmtg:
@atomiks I've studied the code a bit better and used the `includeScale` logic. I had to change a couple of things to get the correct results in my app (I force pushed the new commit, so you can see them by yourself).

However I've some troubles running the tests:

```
 (fix/css-transform) ~/d/floating-ui npm -w packages/dom run dev
Top hits for "packages/dom" "run" "dev"
————————————————————————————————————————————————————————————————————————————————
npm help config                                                    dev:24 run:30
npm help scripts                                                    dev:4 run:33
npm help install                                                    dev:19 run:6
npm help package-json                                               dev:9 run:10
npm help run-script                                                 dev:1 run:18
npm help doctor                                                     dev:1 run:12
npm help ls                                                          dev:6 run:4
npm help audit                                                       dev:4 run:5
npm help folders                                                     dev:1 run:8
npm help prune                                                       dev:1 run:8
————————————————————————————————————————————————————————————————————————————————
(run with -l or --long to see more context)
```

do I have to use a certain npm/node version?

atomiks:
I think you need to be on latest npm with workspaces. Otherwise you can `cd` into `packages/dom` and do `npm run dev` to see the visual tests. The `Transform.tsx` file/route is what you'd want to add onto for the failing test! To verify them just push to this PR and wait for CI to pass. If it fails you can click the Overview to download the snapshot diffs.

cvlmtg:
@atomiks tests pass again, but I still have to add my test, so the PR is not ready yet :) I'll continue tomorrow. Thanks for the help!

cvlmtg:
@atomiks I'm sorry to say this, but the test is done and not working... and if I fix the test my app doesn't work. So it seems the problem is more complex than just looking for `scale < 1` and I have to yet understand what is causing all these problems. It still seems related to a `scale < 1`, but I suppose there are other variables at play tied to the html/css structure

cvlmtg:
I made the test using the current DOM structure and that's the one failing with my patch (and working without). The structure of my app is much more complex and the patch works perfectly or gives some small errors depending on where I apply the `transform:`. As soon as I've found the minimum required tree to replicate the behavior, I'll add another test to the suite and push it. Thanks!

cvlmtg:
@mishaaq no sorry, have been overwhelmed by work recently and i haven't had time yet to write and push the test where the problem appears.

brookjordan:
@cvlmtg any luck here? I’m starting to hit into the same issues.

spadgeaki:
Hello, I would <3 to have this merged too. I don't see any failed tests, is there anything blocking?

atomiks:
Yes, last comment from OP:
> I'm sorry to say this, but the test is done and not working... and if I fix the test my app doesn't work. So it seems the problem is more complex than just looking for scale < 1 and I have to yet understand what is causing all these problems. It still seems related to a scale < 1, but I suppose there are other variables at play tied to the html/css structure

It's not clear what DOM structure/transforms cause positioning issues. Parent transforms are already handled and tested for (and have been for a long time), so it's unclear what the right fix is.

atomiks:
Update:

I found and resolved two distinct issues with transforms that will be out in the next `@floating-ui/dom` release (`1.0.8`) thanks to #2003. You should see improved positioning when dealing with transforms.

There's still the following to be mindful of though:

- If your floating element has a transform, this is not "negated". The floating element is positioned via layout, and a transform you apply in the CSS happens **after** the positioning is done. So if you scale or translate the floating element, the positioning doesn't adjust for it (which is usually a good thing especially with animations). You might be able to write a custom middleware to handle it, especially now with these fixes in preventing other issues.
- Cross-document handling is still not properly supported, so if the floating element is attached to a reference element inside an iframe and the reference is scaled, the position will be off. I tried fixing this months ago but couldn't get the right solution, it seems difficult.

clfeng:
Provides a scene to scale the body to reproduce the bug
https://github.com/floating-ui/floating-ui/issues/2012

atomiks:
I found the cause (99% sure) and solution to this which is now merged, it will be out in `@floating-ui/dom` `v1.0.9`

PR #1796, opened 2022-07-16
Floating UI v1

### LINKED ISSUE #1538: [Umbrella] Roadmap to Floating UI v1

### LINKED ISSUE #1766: `top` get computed to negative value with large tooltips when not setting max-height to dom immediately in `apply` fn
If you dont update the dom immediately in your `apply` function in the `size` middleware the `top` value for the first render of the tooltip get calculated incorrectly  and the tooltip can even be positioned off screen.

Here's a codesandbox showing the bug. When you toggle the tooltip (only the first time) its not positioned at the top of the screen.  I emulated frameworks no updating immediately with a `setTimeout`. Try pressing the rightmost button to disable it and toggle the tooltip. It should now be calculated correctly.

https://codesandbox.io/s/floating-ui-dom-template-forked-yqiiws?file=/src/index.js

Frameworks often do not update the dom immediately. I see that [you can use `flushSync`](https://github.com/floating-ui/floating-ui/issues/1512#issuecomment-1094593531) with react to do it (or just update the dom directly, so bypassing react).

I think this should be noted clearly in the docs of `size` and `apply`.

I am using ember and I have found that just changing [this](https://github.com/floating-ui/floating-ui/blob/master/packages/core/src/middleware/size.ts#L87-L90) from:

```js
    apply?.({
      ...middlewareArguments,
      ...dimensions,
    });
```
to
```js
    await apply?.({
      ...middlewareArguments,
      ...dimensions,
    });
```

allows me to not have to bypass ember and update the dom directly by returning a promise from my apply function that resolves after ember has rendered, like this:

```js
apply({
  availableHeight,
  availableWidth,
  x,
  y,
}) {
  this.maxHeight = `${availableHeight}px`;
  this.maxWidth = `${availableWidth}px`;

  this.left = `${x}px`;
  this.top = `${y}px`;

  return new Promise((resolve) => {
    schedule('afterRender', this, resolve)
  })
```

I guess other frameworks also have ways to wait for a render to have been finished. 

If you want I can make a pr adding the await :)

atomiks:
Closes #1538

This will bump the following packages to v1:

- `core`
- `dom`
- `react-dom`

Following will remain v0 (but will use these packages):

- `react-native`
- `react-dom-interactions`

## Non-breaking

- [Core] `arrow.ts` file just didn't have Prettier applied properly
- [Core] `size()` `apply` has `await` before it (closes #1766)
- [React DOM] allow unstable inline ref callbacks (found a solution to this)
- [React] Add `name` check for middleware comparisons
- [React] Remove `use-isomorphic-layout-effect` from installed deps (as it's bundled in and tiny, anyway)

## Breaking

- [Core] `limitShift()` `offset` option gets passed `MiddlewareArguments` signature to be consistent with other APIs
- [React] Add `.umd` to UMD package files

### DIFF
--- a/packages/core/src/middleware/arrow.ts
+++ b/packages/core/src/middleware/arrow.ts
@@ -79,12 +79,14 @@ export const arrow = (options: Options): Middleware => ({
     const offset = within(min, center, max);
 
     // Make sure that arrow points at the reference
-    const alignmentPadding = alignment === 'start' ? paddingObject[minProp] : paddingObject[maxProp];
-    const shouldAddOffset = alignmentPadding > 0
-      && center !== offset
-      && rects.reference[length] <= rects.floating[length];
-    const alignmentOffset = shouldAddOffset ?
-      center < min
+    const alignmentPadding =
+      alignment === 'start' ? paddingObject[minProp] : paddingObject[maxProp];
+    const shouldAddOffset =
+      alignmentPadding > 0 &&
+      center !== offset &&
+      rects.reference[length] <= rects.floating[length];
+    const alignmentOffset = shouldAddOffset
+      ? center < min
         ? min - center
         : max - center
       : 0;
--- a/packages/core/src/middleware/shift.ts
+++ b/packages/core/src/middleware/shift.ts
@@ -1,10 +1,4 @@
-import type {
-  Middleware,
-  Rect,
-  Placement,
-  MiddlewareArguments,
-  Coords,
-} from '../types';
+import type {Middleware, MiddlewareArguments, Coords} from '../types';
 import {getSide} from '../utils/getSide';
 import {getMainAxisFromPlacement} from '../utils/getMainAxisFromPlacement';
 import {getCrossAxis} from '../utils/getCrossAxis';
@@ -100,7 +94,7 @@ export const shift = (
 });
 
 type LimitShiftOffset =
-  | ((args: {placement: Placement; floating: Rect; reference: Rect}) =>
+  | ((args: MiddlewareArguments) =>
       | number
       | {
           /**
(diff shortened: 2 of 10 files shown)


FezVrasta (approved):
Looks great! Have been using the library in some projects and it really worked well, the API interface ergonomics are really good too.

PR #1799, opened 2022-07-18
docs: switch cdns to jsdelivr

ayuhito:
unpkg is unmaintained and [full of issues](https://github.com/mjackson/unpkg/issues). I switched the documentation over to [jsDelivr](https://www.jsdelivr.com/) since it's actively maintained and is much more reliable due to its [fallback system](https://www.jsdelivr.com/network/infographic) using multiple CDNs.

Likewise, Skypack is also [unmaintained](https://github.com/skypackjs/skypack-cdn/issues), so that was switched over to [esm.run](https://www.jsdelivr.com/esm), which is still part of the jsDelivr network.

Let me know if there are any concerns!

### DIFF
--- a/website/pages/docs/getting-started.mdx
+++ b/website/pages/docs/getting-started.mdx
@@ -83,18 +83,14 @@ Floating UI can be loaded via CDN using ESM or UMD format.
 ### ESM
 
 ```js
-// development
-import {computePosition} from 'https://cdn.skypack.dev/@floating-ui/dom@__DOM_VERSION__';
-
-// production
-import {computePosition} from 'https://cdn.skypack.dev/@floating-ui/dom@__DOM_VERSION__?min';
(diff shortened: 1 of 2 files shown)


FezVrasta:
I never had issues with it to be honest.

atomiks:
The fact that jsDelivr seems better maintained is reason enough to switch I think, thanks

capegreg:
I only used jsdelivr now due to cdnjs issue with cloudflare not fixing Expect-CT header, which Chrome deprecated. 

Though, your distro on jsdelivr is missing source map. 

https://cdn.jsdelivr.net/npm/@floating-ui/core@1.3.1
ENOENT: no such file or directory, open \npm\bootstrap@4.6.2\node_modules\popper.js\dist\esm\popper.js.map

PR #1814, opened 2022-07-26
feat: SafePolygon component

### LINKED ISSUE #1811: [React] Add a SafePolygon component to replace the current safePolygon close handler in useHover
Split out from https://github.com/floating-ui/floating-ui/issues/1807

The current global solution for disabling pointer-events is flaky, some issues raised in that linked ticket:

> We are still running into issues in automated browser tests that don't actually move the cursor off the reference element, rather it tries to just click different elements in sequence. This fails because tools like playwright test before clicking an element if the element is clickable, and this check fails because of the global pointer-events: none;. 

> I also noticed macOS' auto-hiding scrollbars appear when a scrolling element has pointer-events: none put on it in Chrome (doesn't happen in FF/Safari). SafePolygon would have none of these unforseen side effects since it's local

> I pull the latest build with this change and, like you said, scrolling is disabled when the cursor is over the reference. Are there any workarounds to this? We're trying to implement a Slack / Discord style "reaction bar" on hover. This means a user is almost always hovering over a reference item when they want to scroll. I'll play around with it and see what I can find too.

All of these might be fixed by a more local solution.

There could be a <SafePolygon /> component exported by floating-ui that can be rendered inside the Tooltip and positioned correctly. It should render an actual SVG with the correct shape that would cover the safe area. Then there wouldn't be a need for disabling pointer-events globally.

atomiks:
Closes #1811

This exports a new `SafePolygon` component that can be used to block `pointer-events` as the cursor traverses it, a better solution to the current global "holes" strategy due to side effects

Usage:

```js
const {polygonState, safePolygon} = useSafePolygon();

useHover(context, {
  handleClose: safePolygon()
});

return (
  <>
    {/* Avoids containing blocks messing with the position, not always required */}
    <FloatingPortal>
      {open && <SafePolygon state={polygonState} />}
    </FloatingPortal>
    {/* reference / floating stuff here */}
  </>
);
```


atomiks:
@mihkeleidast

packages/react-dom-interactions/src/safePolygon.tsx
```diff
@@ -3,6 +3,7 @@ import type {FloatingContext, FloatingTreeType, ReferenceType} from './types';
 import {getChildren} from './utils/getChildren';
```
atomiks, line 3: This file makes syntax highlighting and TypeScript explode despite being valid 🤔  maybe the JSX component at the bottom, but I'm not sure

mihkeleidast:
A thought on the new component API: as a use, it is not too comfortable to have to use a separate outside state to save the points and pass it on to SafePolygon.

Could it use the context instead? Passing the context to different components seems to be an existing practice:

```tsx
useHover(context, {
  handleClose: safePolygon(),
});

return (
  <>
    {/* Avoids containing blocks messing with the position, not always required */}
    <FloatingPortal>
      {open && <SafePolygon context={context} />}
    </FloatingPortal>
    {/* reference / floating stuff here */}
  </>
);
```

As an alternative, maybe create a hook that saves the state internally:

```tsx
const { SafePolygon, safePolygon } = useSafePolygon();

useHover(context, {
  handleClose: safePolygon(),
});

return (
  <>
    {/* Avoids containing blocks messing with the position, not always required */}
    <FloatingPortal>
      {open && <SafePolygon />}
    </FloatingPortal>
    {/* reference / floating stuff here */}
  </>
);
```

Do I also understand correctly that the SafePolygon is not rendered as a child of the floating element, rather it should be aside it?

packages/react-dom-interactions/src/safePolygon.tsx
```diff
@@ -1,8 +1,11 @@
+import * as React from 'react';
 import type {Side} from '@floating-ui/core';
 import type {FloatingContext, FloatingTreeType, ReferenceType} from './types';
 import {getChildren} from './utils/getChildren';
+import {useLatestRef} from './utils/useLatestRef';
 
 type Point = [number, number];
+type SvgPoints = Array<string | null>;
```
mihkeleidast, line 8: Can you export this type, it is necessary to provide useState this type so TS would not error: `const [points, setPoints] = useState<SvgPoints>([null, null]);`

atomiks:
`useSafePolygon()` is indeed much better. This is breaking in React 17 and 18 when using `flushSync` for some reason due to `onScrollMouseLeave()`, will debug after and try to refactor to a hook

atomiks:
How's this, you still import `SafePolygon` as a component, but now the other one is returned from the hook with `points` state internal within it.

```js
const {polygonState, safePolygon} = useSafePolygon();

useHover(context, {
  handleClose: safePolygon(),
});

return open && <SafePolygon state={polygonState} />;
```

mihkeleidast:
I think this is a good API 👍

atomiks:
Alternative using spreadable props which is slightly more ergonomic imo:

```js
const {polygonProps, safePolygon} = useSafePolygon();

useHover(context, {
  handleClose: safePolygon(),
});

return open && <SafePolygon {...polygonProps} />;
```

atomiks:
Unfortunately this has a problem in Firefox where it won't block pointer-events when the polygon gets generated on the mouseleave for 1 frame ~half the time, doesn't happen in Chrome or Safari 😒 , not sure how to fix it.

mihkeleidast:
How to reproduce that issue? Would investigate a bit, but not sure what exactly to look for.

atomiks:
Add the polygon component to the Menu.tsx component file in the playground. Add another nested submenu right under the "Share..." one. Remove `restMs: 25` from the `safePolygon()` call (non-0 is broken right now)

In Firefox, you can see a 1 frame flicker. This also causes issues seen at the end:

https://user-images.githubusercontent.com/22450188/181524306-82ba4d9f-2e48-45a3-afa6-577fd6bba3bc.mov

maccman:
This one get stalled? :(

atomiks:
https://github.com/floating-ui/floating-ui/pull/2069

Issue #1729, opened 2022-05-26
[Safari/Firefox] Popper is not positioned correctly through shadow-dom when parent is position: relative


WestBrian:
When using `@popperjs/core` the popper is not positioned correctly if it is within the shadow-dom of a `position: relative` parent.

**This issue seems to only be prevalent in the latest Chrome version 102.**

In the picture below the top has a relative parent in the light-dom while the bottom is shadow-dom.

<img width="171" alt="Screen Shot 2022-05-26 at 6 49 21 PM" src="https://user-images.githubusercontent.com/11201644/170592186-12c054fb-7c83-489e-a5c4-d7f56734f4da.png">

[Codesandbox](https://codesandbox.io/s/busy-rumple-hm2oix)

atomiks:
This seems like a bug in Chrome? `tooltip.offsetParent` is incorrectly returning `body`, not its actual `offsetParent` (the relative div)

Dayvisson:
Apparently this is the chromium issue related with this bug : [Issue 1330819: Web component hover is no longer working](https://bugs.chromium.org/p/chromium/issues/detail?id=1330819)

zengguirong:
Same problem in Safari 14.1.2 (16611.3.10.1.6).
Same problem in @floating-ui

atomiks:
This is fixed in Chrome but was always broken in Safari (still is, on 15.4). I would recommend not using this DOM structure due to that. 

Otherwise, you need to patch `.offsetParent` on `HTMLElement.prototype` and traverse the nodes yourself which is another perf hit.

atomiks:
Browsing around... it seems like the spec for `offsetParent` in shadow DOM was changed a while ago but Chrome didn't use that spec until this recent change? Which would explain why Safari and Firefox are currently broken.

https://github.com/PolymerElements/paper-tooltip/pull/156/files

That's a LOT of code just to handle this problem :\

Issue #1815, opened 2022-07-27
[DOM] Wrong placement when an ancestor uses transform and translate when placed in shadow DOM


TomRauchenwaldUnifits:
When floating-ui is used in a webcomponent, and that webcomponent is placed in a container where an ancestor uses a CSS transformation (e.g. transform: translate(..)), the floating element is displaced.

This is a similar issue to #1488, I based the repro on the codepens found in the ticket.
The difference is that the positioning fails if the trigger/floating element is placed in a container, and that div is added to the shadowDom instead of the trigger/floating elements.

OK: https://codepen.io/TomRauchenwald/pen/GRxMMoo
Shadow DOM looks like this:
```html
<button>Trigger</button>
<div class="popper-container"></div>
```

![image](https://user-images.githubusercontent.com/98090196/181191660-574a50a5-84d2-4ec3-b478-b786fb77bfdc.png)

Not ok: https://codepen.io/TomRauchenwald/pen/VwXMzXg
Shadow DOM looks like this:
```html
<div>
<button>Trigger</button>
<div class="popper-container"></div>
</div>
```
![image](https://user-images.githubusercontent.com/98090196/181191754-290229e7-da12-4142-b7b8-950b397be955.png)

The placement is done like this:
```javascript
this.triggerElement.addEventListener("click", () => {
      computePosition(this.triggerElement, this.popperElement, {
        placement: "bottom-end",
        strategy: "fixed"
      }).then(({ x, y }) => {
        Object.assign(this.popperElement.style, {
          display: "block",
          left: `${x}px`,
          top: `${y}px`
        });
      });
    });
```

atomiks:
Previous fix from the other issue for context, maybe something similar needs to be done for this:

https://github.com/floating-ui/floating-ui/blob/22fcda255cbd2ca2464fa63ea8f5b45cef78c768/packages/dom/src/utils/getOffsetParent.ts#L25-L27

atomiks:
1.0.1

PR #1827, opened 2022-08-01
fix: update getOffsetParent#getContainingBlock to avoid shadow roots when traversing ancestors

jcfranco:
#1815 identified the need to avoid shadow roots while traversing ancestors when looking for the containing block. This adds the missing check left out from https://github.com/floating-ui/floating-ui/pull/1535:

> I only modified getOffsetParent to use the host if the starting element's parent is a shadow root, **but this may also be needed for subsequent parent node lookups**.

@atomiks I did not find any existing shadow DOM tests, so please let me know how you'd like me to add coverage for this.


atomiks:
@jcfranco thanks for the PR. I think shadow DOM tests would be great, Popper had these but transition to React/Parcel tests lost them.

jcfranco:
@atomiks I had to slightly sidestep React to set up the custom element scenarios, but this now has shadow DOM tests. May the React gods have mercy on my soul.

atomiks:
Are these images meant to have no floating element visible? 

https://github.com/floating-ui/floating-ui/blob/0d6d83323c658179d94d3c6f4de944d63f4150cf/packages/dom/test/functional/shadow-dom.test.ts-snapshots/direct-host-child-top-fixed-with-transform-darwin.png

jcfranco:
Yikes, they are most definitely not. Apologies.

The issue was that the last click in the functional test did not trigger repositioning in the custom element. This is now fixed and the accompanying screenshots are now up to date.

jcfranco:
FWIW, the following shows the combination of options that this PR fixes:

### before
Screen Shot 2022-08-03 at 12 52 07 AM

### after
Screen Shot 2022-08-03 at 12 52 33 AM

atomiks:
Thank you!

atomiks:
By any chance, do you have an idea of how to fix https://github.com/floating-ui/floating-ui/issues/1729 as well? It seems somewhat important. I think we have to avoid `.offsetParent` entirely for the `getTrueOffsetParent` function.

jcfranco:
I haven’t spent time on that issue, TBH. Would it be reasonable to fix it by using the old `offsetParent` behavior polyfill you found while a better approach is in the works? Seems like it’d be worth the extra bytes for consistent browser behavior.

PR #1835, opened 2022-08-05
fix: use legacy offsetParent logic for consistent browser behavior

### LINKED ISSUE #1729: [Safari/Firefox] Popper is not positioned correctly through shadow-dom when parent is position: relative

jcfranco:
This fixes #1729 based on the PR referenced in [@atomiks's comment](https://github.com/floating-ui/floating-ui/issues/1729#issuecomment-1163894752):

> Browsing around... it seems like the spec for offsetParent in shadow DOM was changed a while ago but Chrome didn't use that spec until this recent change? Which would explain why Safari and Firefox are currently broken.
>
> https://github.com/PolymerElements/paper-tooltip/pull/156/files
>
> That's a LOT of code just to handle this problem :\

It may not be the [desired fix at the moment](https://github.com/floating-ui/floating-ui/pull/1827#issuecomment-1204765505), but it will at least help achieve consistent placement across browsers.

### Additional notes

1. The following shows the combination of options that this PR fixes:

    #### before

    Screen Shot 2022-08-04 at 4 45 10 PM

    #### after

    Screen Shot 2022-08-04 at 4 44 14 PM

2. Before/After file size comparison:

    #### 1.0.0

    * `floating-ui.dom.browser.min.mjs` (7kb) 
    * `floating-ui.dom.browser.mjs` (19kb)
    * `floating-ui.dom.esm.js` (19kb)
    * `floating-ui.dom.mjs` (19kb)
    * `floating-ui.dom.umd.js` (21kb)
    * `floating-ui.dom.umd.min.js` (8kb)

    #### with fix

    * `floating-ui.dom.browser.min.mjs` (7kb)
    * `floating-ui.dom.browser.mjs` (20kb)️️ ⬆️
    * `floating-ui.dom.esm.js` (20kb) ⬆️         
    * `floating-ui.dom.mjs` (20kb) ⬆️
    * `floating-ui.dom.umd.js` (23kb) ⬆️
    * `floating-ui.dom.umd.min.js` (8kb)


atomiks:
I know this is from the PR linked, but the style attr mutations look horrible 😬 and the code increase is huge when many people will never even encounter this issue 😢 

Also how hard does this affect performance of one `computePosition()` call? 

I wonder if an alternative polyfill that's smaller is possible or is this the only way?

jcfranco:
Yeah, I totally understand. I'd have to do some digging to address your concerns. 

Just wanted to make this available as a starting point. At least the tests are ready! 😂

atomiks:
cc @Westbrook I am not sure if you've encountered this, or if you know of a polyfill that can work that is small? 

It's really annoying that they changed this spec without providing a `composedOffsetParent()` method at the same time so now we need these ridiculous hacks to find the actual offsetParent 🤦 .

packages/dom/src/utils/getOffsetParent.ts
```diff
@@ -16,7 +16,54 @@ function getTrueOffsetParent(element: Element): Element | null {
...
+        const oldDisplay = assignedSlot.style.display;
+        assignedSlot.style.display = getComputedStyle(ancestor).display;
+
+        newOffsetParent = assignedSlot.offsetParent;
+
+        assignedSlot.style.display = oldDisplay;
+        if (!hadStyleAttribute) {
+          assignedSlot.removeAttribute('style');
+        }
+      }
```
atomiks, line 53: I'd be fine with this as long as this block can be deleted due to the messy mutations. 

I understand it handles another edge case, but how common is it? Is it feasible to ask users to avoid `display: contents` ancestors?

jcfranco: We actually have [a component](https://developers.arcgis.com/calcite-design-system/components/stepper/) that uses `display: contents` on a slot ancestor to avoid modifying the light DOM when switching between layouts. 😅

Would a flag to opt into this extra check be a fair compromise? With custom elements growing in popularity, it's going to be tricky to prescribe rules of this kind.

atomiks: lol, for all the shadow DOM support, I was considering a flag or similar last year when I first created the lib but it's not DX friendly at all. Everything should just work but the trade-off is extra cost.  Most users should should have layouts that don't end up running the block, only the extra bundle size cost 

I'll do some perf tests to see if this is ok

jcfranco: Gotcha. Thanks for the background. It's a balancing game for sure.

I really appreciate you taking the time to look into the perf aspect of this. LMK if there's anything else I can help with.

Westbrook:
I have not directly encountered this issue, though it does seem important to cover. I'm unsure of a polyfill in this domain, either. I'd be both intrigued and a bit worried about the use of a flag for this. Intrigued in that it's a good escape hatch, but worried because a consumer _might_ not know they are using shadow DOM and _shouldn't_ need to know. Have you opened an issue in the HTML spec for `composedOffsetParent()`? This idea of "new" specs not taking into account shadow DOM or other web components APIs is something we're looking for a way to surface better in the Web Components Community Group, right now.

packages/dom/src/utils/getOffsetParent.ts
atomiks: I tested this briefly and the performance seems fine, although I'm not sure about really deep DOM trees and what not. Probably ok though considering we use while loops in other places that do similar things.

packages/dom/src/utils/getOffsetParent.ts
```diff
@@ -16,7 +16,54 @@ function getTrueOffsetParent(element: Element): Element | null {
...
+}
+
+/**
+ * Polyfills the old offsetParent behavior from before the spec was changed:
+ * https://github.com/w3c/csswg-drafts/issues/159
+ */
+function composedOffsetParent(element: Element): Element | null {
+  if (!isHTMLElement(element)) {
+    return null;
+  }
```
atomiks, line 29: I think this check can be removed if you type the param as `HTMLElement`

packages/dom/src/utils/getOffsetParent.ts
```diff
@@ -16,7 +16,54 @@ function getTrueOffsetParent(element: Element): Element | null {
...
+  let ancestor: Element = element;
+  let foundInsideSlot = false;
+
+  while (ancestor && ancestor !== offsetParent) {
+    const { assignedSlot } = ancestor;
+
+    if (assignedSlot) {
+      let newOffsetParent = assignedSlot.offsetParent;
+
+      if (getComputedStyle(assignedSlot)['display'] === 'contents') {
```
atomiks, line 42:
```suggestion
      if (getComputedStyle(assignedSlot).display === 'contents') {
```

packages/dom/src/utils/getOffsetParent.ts
```diff
@@ -16,7 +16,54 @@ function getTrueOffsetParent(element: Element): Element | null {
...
+        const hadStyleAttribute = assignedSlot.hasAttribute('style');
+        const oldDisplay = assignedSlot.style.display;
+        assignedSlot.style.display = getComputedStyle(ancestor).display;
+
+        newOffsetParent = assignedSlot.offsetParent;
+
+        assignedSlot.style.display = oldDisplay;
+        if (!hadStyleAttribute) {
+          assignedSlot.removeAttribute('style');
+        }
```
atomiks, line 52: Is this just so it doesn't leave a `style` attr in the Elements panel? Realistically does this matter?

jcfranco: I think it matters if the `assignedSlot`'s computed `display` changes from `contents` to something else between positioning calls. Otherwise, the `assignedSlot`'s inline style would remain, be out of sync, and potentially affect layout.

packages/dom/src/utils/getOffsetParent.ts
jcfranco: Awesome, thanks for testing!

jcfranco:
@atomiks Thanks for the merge! 🎉

Do you think `@floating-ui/dom` could be updated soon?

atomiks:
in a sec

PR #1847, opened 2022-08-09
[dom] Do not treat elements with display: inline or display: contents as overflow ancestors

### LINKED ISSUE #1831: [DOM] elements with overflow hidden but display inline or contents should not be an overflow ancestor
If I'm not mistaken, elements that have an overflow hidden (or auto, etc.) should not be taken into account when determining overflow ancestors as the following CSB shows browsers ignore the overflow property and the calculation of the `referenceHidden` flag of the `hide` middleware is wrong as the elements have a `clientHeight`/`clientWidth` of 0 (the middleware says the reference is always hidden, which is not true).

Note I originally detected this with custom elements (defaulting to `display: inline` and thus always being an overflow ancestor if a floating element is used in a custom element) but it seems it also applies to all other elements with `display: inline` (or `contents`).
 
https://codesandbox.io/s/floating-ui-custom-element-overflow-hslr9n?file=/src/index.tsx

sventschui:
Fixes #1831

### DIFF
--- a/packages/dom/src/utils/is.ts
+++ b/packages/dom/src/utils/is.ts
@@ -36,8 +36,11 @@ export function isShadowRoot(node: Node): node is ShadowRoot {
 
 export function isOverflowElement(element: HTMLElement): boolean {
   // Firefox wants us to check `-x` and `-y` variations as well
-  const {overflow, overflowX, overflowY} = getComputedStyle(element);
-  return /auto|scroll|overlay|hidden/.test(overflow + overflowY + overflowX);
+  const {overflow, overflowX, overflowY, display} = getComputedStyle(element);
+  return (
+    /auto|scroll|overlay|hidden/.test(overflow + overflowY + overflowX) &&
+    !['inline', 'contents'].includes(display)
+  );
 }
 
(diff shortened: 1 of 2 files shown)


sventschui:
@atomiks any chance of getting this merged?

atomiks:
It looks fine but i'm wondering why it's urgent? Can't you just remove `overflow: hidden` on these elements since it doesn't even do anything?

sventschui:
I can try. The thing is this is added by some general CSS rules that are partially undone. I.e. we apply this to all `:host` elements but in this case said element gets added a `display: inline`. I'll try to work around this then

atomiks:
I forgot to merge this sorry :(

sallerli1:
hello, I am developer of ui component lib called idux, we  sincerely thank you for providing floating-ui.

I am wandering when will @floating-ui/dom@1.0.3 be released because one of our component relies on the issue: https://github.com/floating-ui/floating-ui/issues/1831 being fixed

Thanks!

atomiks:
@sallerli1 I plan to do a release this week (was meant to do it on the weekend)

PR #1853, opened 2022-08-10
feat(safePolygon): add `blockPointerEvents` option

atomiks:
Since #1811 is stalled, and maybe doesn't have a good fix, this option is much cleaner for disabling the `pointer-events` blocking behavior without needing the useLayoutEffect workaround.


dbousamra:
This helped me immensely. I've been trying to implement a nested hover, but need to be able to scroll the parent. Adding `blockPointerEvents: false` works

PR #1854, opened 2022-08-10
(WIP) Recording functional tests with Replay

jasonLaster:
### 🚧  DO NOT MERGE 🚧 

Replay.io is the first time travel debugger for the web. This means you can record your playwright tests and debug them after the fact exactly as they ran before. By the way, you can also record other bugs with the Replay Browser as well ([docs](https://www.notion.so/replayio/Recording-Browser-Tests-Beta-5bf7d91b65cd46deab1867b07bd12bdf)). 

### Status

We're just beginning to roll out Test Suite support. The hardest part is getting projects setup, so happy to help with the integration piece.


jasonLaster:
going to close for now and come back later

PR #1861, opened 2022-08-20
feat(useClick): add ignoreKeyboard option

huijiewei:


atomiks:
When is this needed?

huijiewei:
> When is this needed?

When write a searchable and disable closeOnSelect option Select Component.
because need focus on search input, so need keep keyboard event on reference dom.

source example: https://github.com/huijiewei/agile-ui/blob/main/packages/react/src/select/Select.tsx

atomiks:
If the reference is an input (combobox), shouldn't `useFocus` (w/ `keyboardOnly` disabled) be used instead of `useClick`?

huijiewei:
Select component with searchable need click open option menu, not need focus open menu. can try React Select: https://codesandbox.io/s/codesandboxer-example-forked-g7ei5v ,it's events very awesome. now i need same action with closeMenuOnSelect.

sorry my english.

huijiewei:
The selector component needs to be clicked on to close it. Clicking on it with the closeOnSelect property on is still valid, but keyboard actions should be ignored.

For example, the behavior of the React-Select component is the effect I want。
https://codesandbox.io/s/codesandboxer-example-forked-g7ei5v?file=/example.tsx:1443-1460

atomiks:
Does disabling `toggle` option work?

huijiewei:
If toggle is disabled, the original behavior on the reference is disabled

atomiks:
@huijiewei can you make a demo of this on CodeSandbox (minimal repro) of what you want? I'm lost 

Regardless, I think this option should be `addKeyboardHandlers` or something (which is set to `true` by default) because this `ignoreKeyboard` behavior won't work for `<button>` elements.

huijiewei:
You are right, I use the option name ignoreKeyboard just to correspond to ignoreMouse, addKeyboardHandlers or other option names are more suitable, later I will submit a minimum example, the code is more

huijiewei:
mini demo :https://codesandbox.io/s/hardcore-khayyam-11lv3h?file=/src/App.tsx

atomiks:
So to clarify:

- `useFocus` isn't suitable because you don't want the floating element to open right away when tabbing into the input?
- `toggle: false` in `useClick` isn't suitable because you don't want pressing `Enter` while on the input to open the floating element?

As far as I can tell those are the only differences with your custom click solution

huijiewei:
1. yes, i don't need focus open.
2. if toggle set false, the mouse click can't toggle open and close.

i want to third(closeOnSelect is off and use ignoreKeyboard) demo.

atomiks:
Okay I see. It's fine to merge if this option name is changed to `keyboardHandlers: boolean`

huijiewei:
Thanks.

PR #1866, opened 2022-08-22
fix(useClick) ensure correct handling of events for wrapped buttons

jmurzy:
The current implementation of `useClick` bails on keyDown/Up handlers on button targets to avoid redundant handling of float state because pressing ENTER/SPACE on buttons trigger a click event. However, checking if the reference element is a button causes it to incorrectly bail on the wrong element type when an event is triggered by an element that is not the reference element. 

Instead, one possible solution is to check if the target that triggered the event is a button regardless of the type of the wrapping reference.

[Here](https://codesandbox.io/s/serene-noether-bqk39m?file=/src/App.tsx)'s a reproduction of the issue.

### DIFF
--- a/packages/react-dom-interactions/src/hooks/useClick.ts
+++ b/packages/react-dom-interactions/src/hooks/useClick.ts
@@ -2,8 +2,8 @@ import type {ElementProps, FloatingContext, ReferenceType} from '../types';
 import * as React from 'react';
 import {isTypeableElement} from '../utils/isTypeableElement';
 
-function isButtonTarget(event) {
-  return event.target && event.target.tagName === 'BUTTON';
+function isButtonTarget(event: React.KeyboardEvent<Element>) {
+  return event.target && event.target instanceof HTMLElement && event.target.tagName === 'BUTTON';
 }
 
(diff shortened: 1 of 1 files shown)


atomiks:
This DOM layout has issues with screen readers because the reference props are attached to the non-focusable span element.

Why is the wrapper span necessary around the button inside of the component's `children` and not outside of it?

jmurzy:
@atomiks that is just an example to demonstrate the issue.

atomiks:
I see but is what is the real world layout you have? If the button is not receiving the events/reference props, it would be incompatible with screen readers. Why is it not possible to attach the props to the button directly instead of a wrapper?

jmurzy:
One use case is a tooltip on a button with an accessory in which the button as well as the accessory can also have its own individual popovers. I understand I can pass event listeners around and have the individual elements trigger events for the parent tooltip. However, this isn't possible in UI components where the consumer has limited access to its internals to spread over the underlying DOM elements.

Given floating-ui aims to stay as a lower level API, I don't think we should prescribe a layout for it to handle bubbled events correctly.

PR #1868, opened 2022-08-24
fix(limitShift): always apply offset regardless of the axis and the side

### LINKED ISSUE #1859: `limitShift()`: take `middlewareData.offset` into account when limiting the `mainAxis`
While working on https://github.com/WordPress/gutenberg/pull/42950, I reliazed a discrepance in the `limitShift` logic (that is probably at the cause of [this issue](https://github.com/WordPress/gutenberg/pull/42950#issuecomment-1209674636) that I haven't been able to fix):

When limiting the `crossAxis`, any potential `offset` from previous offset middleware is taken into account when computing the `min` and `max` limits for the `crossAxis`:

https://github.com/floating-ui/floating-ui/blob/15d1f49b966f9cf472141749f6ebdb4c75e62d5d/packages/core/src/middleware/shift.ts#L197-L206

However, the offset from the `offset` middleware is NOT taken into account when computing the `min` and `max` limits for the `mainAxis`:

https://github.com/floating-ui/floating-ui/blob/15d1f49b966f9cf472141749f6ebdb4c75e62d5d/packages/core/src/middleware/shift.ts#L178-L185

This means that the `limitShift` logic can't work properly on the `mainAxis` if there's an offset from the `offset` middleware on that axis — is there a reason why the offset is not taken into account?

Can the code be changed to include such offset? e.g.

```js
const limitMin =
  rects.reference[mainAxis] -
  rects.floating[len] +
  middlewareData.offset?.[mainAxis] ?? 0 +
  computedOffset.mainAxis;
const limitMax =
  rects.reference[mainAxis] +
  rects.reference[len] +
  middlewareData.offset?.[mainAxis] ?? 0 -
  computedOffset.mainAxis;
```

ciampo:
Fixes #1859 

This PR changes the `limitShift` function from the `shift` middleware so that it always applies the `middlewareData.offset` when computing mix/max limits for both the main and the cross axis, regardless of the side (origin or not).

See https://github.com/floating-ui/floating-ui/issues/1859 and https://github.com/WordPress/gutenberg/pull/42950 for extra context

### DIFF
--- a/packages/core/src/middleware/shift.ts
+++ b/packages/core/src/middleware/shift.ts
@@ -199,12 +199,12 @@ export const limitShift = (
       const limitMin =
         rects.reference[crossAxis] -
         rects.floating[len] +
-        (isOriginSide ? middlewareData.offset?.[crossAxis] ?? 0 : 0) +
+        (middlewareData.offset?.[crossAxis] ?? 0) +
         (isOriginSide ? 0 : computedOffset.crossAxis);
       const limitMax =
         rects.reference[crossAxis] +
         rects.reference[len] +
-        (isOriginSide ? 0 : middlewareData.offset?.[crossAxis] ?? 0) -
+        (middlewareData.offset?.[crossAxis] ?? 0) -
         (isOriginSide ? computedOffset.crossAxis : 0);
 
       if (crossAxisCoord < limitMin) {
--- a/packages/core/src/middleware/shift.ts
+++ b/packages/core/src/middleware/shift.ts
@@ -178,11 +178,13 @@ export const limitShift = (
       const limitMin =
         rects.reference[mainAxis] -
         rects.floating[len] +
-        computedOffset.mainAxis;
+        computedOffset.mainAxis +
+        (middlewareData.offset?.[mainAxis] ?? 0);
       const limitMax =
         rects.reference[mainAxis] +
         rects.reference[len] -
-        computedOffset.mainAxis;
+        computedOffset.mainAxis +
+        (middlewareData.offset?.[mainAxis] ?? 0);
 
       if (mainAxisCoord < limitMin) {
         mainAxisCoord = limitMin;


packages/core/src/middleware/shift.ts
```diff
@@ -178,11 +178,13 @@ export const limitShift = (
       const limitMin =
         rects.reference[mainAxis] -
         rects.floating[len] +
-        computedOffset.mainAxis;
+        computedOffset.mainAxis +
+        (middlewareData.offset?.[mainAxis] ?? 0);
```
ciampo, line 182: Not sure if this offset should be applied, or if it should be applied conditionally (ie. only if the placement includes `top` and `bottom`) ?

packages/core/src/middleware/shift.ts
```diff
@@ -197,12 +199,12 @@ export const limitShift = (
       const limitMin =
         rects.reference[crossAxis] -
         rects.floating[len] +
-        (isOriginSide ? middlewareData.offset?.[crossAxis] ?? 0 : 0) +
+        (middlewareData.offset?.[crossAxis] ?? 0) +
         (isOriginSide ? 0 : computedOffset.crossAxis);
       const limitMax =
         rects.reference[crossAxis] +
         rects.reference[len] +
-        (isOriginSide ? 0 : middlewareData.offset?.[crossAxis] ?? 0) -
```
ciampo, line 205: I don't really understand the current approach on `master` where `middlewareData.offset` is added depending on `isOriginSide`

atomiks:
The current logic ensures that the offset does not impact the goal to stop shifting once the opposite edges of the reference and floating elements are aligned. With this change, the shifting stops based on the offset, not whether their edges are aligned.

ciampo:
Thank you for the explanation.

I'll explain my use-case a bit more in depth:

In the Gutenberg editor, there are situations in which the reference element is inside an iframe, while the floating element is rendered outside of the iframe.

The iframe has an offset with respect to the document's viewport, which `floating-ui` doesn't seem to apply correctly to the floating element — and therefore, [we've been adding the iframe offset ourselves via the `offset` middleware](https://github.com/WordPress/gutenberg/blob/2978ee2d3ff2d883565ba3fa8aa71a5a88863fe3/packages/components/src/popover/index.js#L207-L235).

This causes problems with the `limitShift` function, since it applies the offset from the `offset` middleware arbitrarily.
We also tried to use the `offset` property of the `limitShift` function, but that also doesn't work (since the `limitShift` function adds/subtracts the offset for the min/max values).

We're now experimenting with adding the iframe offset via a separate custom middleware, and also using a custom version of the `limitShift` function which always adds that iframe offset to the min/max values.

We're still testing that solution, but it seems to work for us — although I'm wondering if there's a better way around it. Do you have any thoughts?

Probably the ideal case scenario would be that `floating-ui` adds better support for when the reference is in a different `ownerDocument` than the floating?

atomiks:
I think better handling of cross-document elements is likely required to get it working properly. 

Interestingly the future CSS anchor positioning proposal says it will forbid this:

> Security
> [...] elements cannot be anchored to each other across document boundaries.

ciampo:
> I think better handling of cross-document elements is likely required to get it working properly.

I think so too — is that on the roadmap for `@flaoting-ui` ?

It would be quite useful to the Gutenberg project too, so in that case it may be interesting to see if we can help move this feature forward.

> Interestingly the future CSS anchor positioning proposal says it will forbid this:

That is something I wasn't aware of — it looks interesting, though! (and good point about document boundaries). I guess it would be great if there was an explicit way to remove those boundaries if both frames specially state it in some ways (maybe via CSP or something?)

> Alternatively, a flag option in the limitShift function to ignore offset (or something else) might work

That would be the quickest short-term solution — an additional flag that allows to pass an arbitrary offset value that _always_ gets _added_ to min/max limits. Would that be possible?

atomiks:
I probably won't work on it myself soon, but you're welcome to create a PR to better support cross-document elements for the `dom` package

ciampo:
I'm going to close this PR and the related issue, and instead I opened two new issues:

- https://github.com/floating-ui/floating-ui/issues/1869 (as a long-term solution)
- https://github.com/floating-ui/floating-ui/issues/1870 (as a more short-term solution)

It would be great if we could continue the conversation in those two issues.

PR #1881, opened 2022-09-10
feat: configure events

### LINKED ISSUE #1879: [React DOM Interactions] useDismiss hook behavior
I have a Dropdown component, which in the mobile version (iPhone SE 375x667), when opened, instead of Popup, shows options from the bottom of the screen

https://codesandbox.io/s/magical-wilbur-bd9pqk?file=/src/Dropdown.tsx

The problem is that if these are nested Dropdowns, then closing the upper layers becomes problematic, because when clicking outside the floating element, it can immediately open another Dropdown and I can’t basically close it without a separate button or clicking on a non-interactive area

This behavior is only present on touch screens, is it possible to configure this to be the same as on the desktop?

atomiks:
This allows the user to configure the events for `useClick` and `useDismiss` and avoids the `pointerDown` terminology

Closes #1879 

## useDismiss

- `outsidePointerDown` => `outsidePress`
- `outsidePressEvent: string`
- `referencePointerDown` => `referencePress`
- `referencePressEvent: string`

`pointerdown` (default)

When you want to dismiss on both mouse/touch **eagerly**

`mousedown`

When you want to dismiss eagerly with mouse but lazily with touch

`click`

When you want to dismiss on both mouse/touch **lazily**

## useClick

After discovering it's poor UX to use `pointerdown` instead of `mousedown` on touch devices this was changed. So now the option doesn't match what it is doing.

`pointerDown` => `event: 'click' | 'mousedown'`

### DIFF
--- a/packages/react-dom-interactions/src/hooks/useDismiss.ts
+++ b/packages/react-dom-interactions/src/hooks/useDismiss.ts
@@ -8,6 +8,18 @@ import {getTarget} from '../utils/getTarget';
 import {isElement} from '../utils/is';
 import {isEventTargetWithin} from '../utils/isEventTargetWithin';
 
+const bubbleHandlerKeys = {
+  pointerdown: 'onPointerDown',
+  mousedown: 'onMouseDown',
+  mouseup: 'onMouseUp',
+};
+
+const captureHandlerKeys = {
+  pointerdown: 'onPointerDownCapture',
+  mousedown: 'onMouseDownCapture',
+  mouseup: 'onMouseUpCapture',
+};
+
 export interface Props {
   enabled?: boolean;
   escapeKey?: boolean;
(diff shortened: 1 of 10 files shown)


packages/react-dom-interactions/src/hooks/useDismiss.ts
```diff
@@ -174,17 +179,29 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
...
+    pointerdown: 'onPointerDown',
+    mousedown: 'onMouseDown',
+    mouseup: 'onMouseUp',
+  }[referencePressEvent];
+
+  const floatingPropKey = {
+    pointerdown: 'onPointerDownCapture',
+    mousedown: 'onMouseDownCapture',
+    mouseup: 'onMouseUpCapture',
+  }[outsidePressEvent];
```
FezVrasta, line 192: Maybe these could be defined on the root scope to avoid to re-allocate them on each call?

PR #1884, opened 2022-09-13
Add better cross-document support

### LINKED ISSUE #1869: [dom] Add better support for cross-document elements
Currently, the library doesn't handle very well scenarios in which the reference and the floating elements are rendered in two different documents — for example, when the reference is inside an iframe, and the floating element is rendered outside of the iframe.

For more context, see conversation in #1868

ciampo:
Closes #1869

## What

This PR adds better support for cases where the reference and the floating elements are in two different documents — for example, if the reference is inside an `iframe`, while the floating element is rendered in the top-level document.

TODO:

 - [ ] make sure current approach doesn't introduce regressions (especially around clipping rect detection)
 - [ ] support scenarios:
   - [ ] reference inside iframe, floating in top document
   - [ ] reference in top document, floating inside iframe
   - [ ] reference and floating inside same iframe
   - [ ] nested iframes?
 - [ ] Add event listeners for the extra documents:
   - [ ] scroll
   - [ ] resize
   - [ ] mutation observer?
   - [ ] make sure it works when using the `animationFrame` autoupdate strategy
 - [ ] add tests
   - [ ] for each scenario highlighted above
   - [ ] for each placement
   - [ ] including arrow position
   - [ ] including shifting (with and without limiting)
   - [ ] including flipping
 - [ ] add documentation

## How

In particular, this PR:

- adds a new page to the Visual Playground
- adds a new optional prop on the `VirtualElement` type — `ownerDocument`. This prop is necessary to implement checks for cross-document support.
- improves the logic of the `getRectRelativeToOffsetParent` function in the `dom` package, taking into account (when needed) the offset of the reference's parent `iframe`

### DIFF
--- a/packages/dom/test/visual/spec/CrossDocument.tsx
+++ b/packages/dom/test/visual/spec/CrossDocument.tsx
@@ -1,0 +1,150 @@
+import { useState, useLayoutEffect, useMemo, HTMLProps } from 'react';
+import { createPortal } from 'react-dom'
+
+import type { Placement } from '@floating-ui/core';
+import {useFloating} from '@floating-ui/react-dom';
+
+import { Controls } from '../utils/Controls';
+import { defineElements } from '../utils/shadowDOM';
+import { allPlacements } from '../utils/allPlacements';
+import {useScroll} from '../utils/useScroll';
+
+defineElements();
+
+export const IFrame = ({
+  children,
+  ...props
+}: HTMLProps<HTMLIFrameElement>) => {
+  const [contentRef, setContentRef] = useState<HTMLIFrameElement | null>(null)
+
+  const iFrameHead = contentRef?.contentWindow?.document?.head;
+  const iFrameRoot = contentRef?.contentWindow?.document?.body;
+
+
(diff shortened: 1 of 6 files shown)


packages/dom/test/visual/spec/CrossDocument.tsx
```diff
@@ -0,0 +1,150 @@
...
+    update,
+    placement,
+  ]);
+
+  useLayoutEffect(() => {
+    // Hack to force a re-render that also creates new values for
+    // `refs` and/or `update`, so that `useScroll` updates internally
+    // and scrolls the reference element into view.
+    setTimeout(() => setPlacement('bottom-start'));
+  }, []);
```
ciampo, line 68: This fix doesn't really seem to work for now. I'll need to find a better way to force `floating-ui` to update

atomiks:
I've made a PR here that seems to solve it — I haven't bothered with nested iframes but does that have any actual use case?

https://github.com/floating-ui/floating-ui/pull/2043

PR #1894, opened 2022-09-19
perf: revert offsetParent polyfill

atomiks:
Kinda felt this was going to happen, this is causing poor performance as noted in here: https://github.com/shoelace-style/shoelace/discussions/902#discussioncomment-3607642

Going to revert this for now, unfortunately there's no fast native `composedOffsetParent()` to use and this polyfill is slow.

Users will have to:

- Simply avoid using the DOM layout that causes issues, or;
- This PR exports `platform` from the `dom` package, so they can spread in the other methods and only rewrite `getOffsetParent` to include this polyfill themselves without needing `patch-package`. If this is necessary for React DOM it can also be added

### DIFF
--- a/packages/dom/index.test-d.ts
+++ b/packages/dom/index.test-d.ts
@@ -10,6 +10,7 @@ import {
   arrow,
   detectOverflow,
   Middleware,
+  platform,
 } from '.';
 
 // @ts-expect-error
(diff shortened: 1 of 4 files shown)


jcfranco:
We started noticing performance hits from https://github.com/floating-ui/floating-ui/pull/1835/ as well. 😞

> ... the default platform object is now exported so you can spread in all the other methods then write your own single getOffsetParent hook that has the polyfill. You might want to add your own custom flags to turn it on/off when it's considered acceptable in a given scenario perf-wise.

Thanks for exposing this, seems like a good middle-ground.

WRT to the lack of `composedOffsetParent()`, I'll create an issue as suggested in  https://github.com/floating-ui/floating-ui/pull/1835#issuecomment-1207216945 to hopefully get some traction on it.

PR #1899, opened 2022-09-23
[v2] use `ComputedPlacement` type in returned `State`

tjenkinson:
I don't think it's possible for `placement` to be `auto` in the returned state, so this updates the types to make that the case. This will make the types slightly nicer in a project we have that uses this

If this is accepted I'll open up a followup PR to update https://github.com/floating-ui/react-popper/blob/beac280d61082852c4efc302be902911ce2d424c/src/usePopper.js#L35


atomiks:
Hi 👋 , thank you for the PR, but old unmerged PRs for Popper are now being closed.

PR #1925, opened 2022-10-20
fix: virtual contextElement ancestors overflow

kelreel:
When using a [Virtual Elements](https://floating-ui.com/docs/virtual-elements) and `contextElement` property, event listeners are not added to its ansectors.

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -1,7 +1,7 @@
 import type {ReferenceElement, FloatingElement} from './types';
 import {getBoundingClientRect} from './utils/getBoundingClientRect';
 import {getOverflowAncestors} from './utils/getOverflowAncestors';
-import {isElement, isVirtualElement} from './utils/is';
+import {isElement} from './utils/is';
 
 export interface Options {
   /**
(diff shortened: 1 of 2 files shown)


atomiks:
The same `isElement`-first check would be simpler:

From Popper:
https://github.com/floating-ui/floating-ui/blob/e764f2cc893e5a9cc466af888f7eb7efa3b197f6/src/createPopper.js#L91-L96

I forget when this is useful, do you have an example?

kelreel:
An additional type guard is not needed, the code from the example is enough

atomiks:
The `ResizeObserver` should also get attached to the `contextElement`, I think:

https://github.com/floating-ui/floating-ui/blob/fbc617ae00b6d016bc0be169272e26ca720f7d2b/packages/dom/src/autoUpdate.ts#L82

kelreel:
`isVirtualElement` would allow you  check the link type more clearly.
But now several tests are falling with a difference of a few pixels.

image

atomiks:
The tests run on CI so you don't need to look at them locally. They passed in the prev commit and are still running

packages/dom/src/autoUpdate.ts
```diff
@@ -76,6 +80,13 @@ export function autoUpdate(
       initialUpdate = false;
     });
     isElement(reference) && !animationFrame && observer.observe(reference);
+    if (
+      !isElement(reference) &&
+      isElement(reference.contextElement) &&
```
atomiks, line 85: I think it can just be `reference.contextElement &&`

PR #1934, opened 2022-10-30
update: border radius when in mobile layout

thetechie7:
The mobile layout for the cards had a border-radius large when in mobile view, which makes it looks odd. 

Screenshot 2022-10-30 at 11 58 42 AM

### DIFF
--- a/website/pages/index.js
+++ b/website/pages/index.js
@@ -608,7 +608,7 @@ function HomePage() {
             . Tooltips, popovers, dropdowns, menus, and more.
           </p>
         </div>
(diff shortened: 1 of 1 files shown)


atomiks:
I can't see any overflow or padding issues on the current site

Screenshot 2022-10-31 at 10 05 40 am

thetechie7:
I meant I think it needs some padding as it looks better with it contained so I added it.

atomiks:
The reason it doesn't have padding is to make it as wide as possible, especially for 320px wide screens. I think instead the `border-radius` should become 0 once it takes up 100% width

thetechie7:
alright let me do that...

atomiks:
Looks good, ty

PR #1936, opened 2022-10-31
feat: accept focusVisible in FloatingFocusManager

brookjordan:
### Motivation

A lot of websites are relying on `:focus-visible` to tidy up the UI for sighted users. By default, programatically setting the focus after a click will not trigger the focus visible. In situations where a modal is being used for a dialog where the expected action is to hit enter to continue, having the focused action button visibly highlighted to remove ambiguity on what the enter button will do is useful.

#### Provisos

- I want to apologise as I’ve built this PR blind. When I build the project I keep getting errors about using `useState` outside of a component, even if I build direct from master. If anyone could help out and pick up where I’ve started I’d be super grateful!
- So far only Firefox supports this syntax, hence the `@ts-expect-error`. I’d love to have projects like this help to get these features noticed to raise the priority of implementation

### DIFF
--- a/packages/react-dom-interactions/src/FloatingFocusManager.tsx
+++ b/packages/react-dom-interactions/src/FloatingFocusManager.tsx
@@ -12,11 +12,21 @@ import {isTypeableElement, TYPEABLE_SELECTOR} from './utils/isTypeableElement';
 import {stopEvent} from './utils/stopEvent';
 import {useLatestRef} from './utils/useLatestRef';
 
-function focus(el: HTMLElement | undefined, preventScroll = false) {
+function focus(
+  el: HTMLElement | undefined,
+  {
+    preventScroll = false,
+    focusVisible,
+  }: {preventScroll?: boolean; focusVisible?: boolean} = {}
+) {
   // `mousedown` clicks occur before `focus`, so the button will steal the
   // focus unless we wait a frame.
   requestAnimationFrame(() => {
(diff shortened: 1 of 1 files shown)


atomiks:
> In situations where a modal is being used for a dialog where the expected action is to hit enter to continue, having the focused action button visibly highlighted to remove ambiguity on what the enter button will do is useful.

Does sound useful. I see you've added it to every `focus()` call, but is this property only relevant for `initialFocus`? Would it be wanted for `returnFocus` in some situations? 

I also see you've added it to `onKeyDown` ones, but they should already trigger `:focus-visible`, so shouldn't be necessary for them.

brookjordan:
That totally makes sense, thanks. I can imagine that some places would either not be need it (due to keyboard interaction) or wouldn’t be desirable (not sure where yet, but maybe). I’ll try to find a little more time to make this PR more thoughtful.

I think I honestly got frustrated that I was in a bit of a rush and couldn’t work out how to get the packages running in a local project to test it out quickly, so ended up doing it everywhere.

atomiks:
This is pretty out of date now so I'm going to close this, but I still want to add it. 
 
We should wait until the option is available in Chrome at least first though because it's not viable in practice in its current state (i.e., it would be better to manually polyfill it). [MDN still says it's only available in Firefox](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#browser_compatibility).

PR #1956, opened 2022-11-13
fix: clipping detection

### LINKED ISSUE #1867: [DOM] `hide` middleware `referenceHidden` incorrect when reference ancestor is `position: absolute`
For the following HTML structure, hide middleware return `true`, indicate that reference element been hided.
Event though its parent is  `overflow:hidden` and computed style `height: 0`,  actually  `position: absolute` make it related with outer relatived ancestor.

```
    <div style="position: relative;">
      <div style="overflow: hidden; border: 4px solid green;">
        <div style="position: absolute;">
          <div id="reference"></div>
        </div>
        <div id="floating">Floating element</div>
      </div>
    </div>
```

sandbox: https://codesandbox.io/s/mutable-browser-euniq7?file=/src/index.js:318-369

image

atomiks:
Fix #1867 

Interestingly with this change the entire `contains` util can be removed...


packages/dom/src/utils/getClippingRect.ts
```diff
@@ -93,35 +59,39 @@ function getClientRectFromClippingAncestor(
...
-    clippingAncestors
-  );
+function getClippingElementAncestors(element: Element): Array<Element> {
+  let result = getOverflowAncestors(element).filter((el) =>
+    isElement(el)
+  ) as Array<Element>;
+  let currentNode: Node | null = element;
+  let hasEscapableParent = false;
+
+  while (isElement(currentNode) && !isLastTraversableNode(currentNode)) {
```
gavinxgu, line 69: This logic is not quite correct, we should follow the [CSS Containing Block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block) specification.

I think this logic will break in this case https://codesandbox.io/s/nervous-newton-ouzf5e.

PR #1959, opened 2022-11-15
fix: clipping detection (#1956)

gavinxgu:
Refactor `getClippingElementAncestors` according to https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block


atomiks:
Wow thanks @gavinxgu! There is a utility called `isContainingBlock` which you could use instead. It's missing `backdrop-filter` though

PR #1961, opened 2022-11-15
fix(getClippingRect): ignore `body` for clipping elements

atomiks:
I just noticed this on the MacSelect demo - `overflow: hidden` (or any `overflow`) on the `<body>` should likely be ignored in the majority of cases. Maybe there is an edge case where it shouldn't, or a better technique here? 

The previous code filtered out the body.

Without this change, the clipping rect would match the `<body>`, not the entire window (which is possible for the floating element to overflow to)

cc: @gavinxgu 

Screenshot 2022-11-16 at 4 08 01 am

### DIFF
--- a/packages/dom/src/utils/getClippingRect.ts
+++ b/packages/dom/src/utils/getClippingRect.ts
@@ -15,6 +15,7 @@ import {isElement, isLastTraversableNode, isContainingBlock} from './is';
 import {getBoundingClientRect} from './getBoundingClientRect';
 import {max, min} from './math';
 import {getParentNode} from './getParentNode';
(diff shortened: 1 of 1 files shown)


atomiks:
@gavinxgu's change detected that it's related to containing blocks creating the "clipping" effect. So adding something like `will-change: transform` to the `<body>` will do that (or a containing block that's a child of it) in the screenshot. 

This change seems to work fine when making the `<body>` a containing block itself (or the `root`) although the scroll arrows have the wrong position for some reason — that is a different issue though.

Screenshot 2022-11-16 at 4 27 41 am

PR #1967, opened 2022-11-21
 remove redundant floating dom validity check

Yiniau:
the isHTMLElement call to offsetParent may be redundant?


FezVrasta (approved):

atomiks:
Well as you can see the type checking check failed, TypeScript 4.1 is not good in this case, I think it's fixed in later versions though

PR #1968, opened 2022-11-21
Fix error with user-agent brand not being an array

arishoham:
We have thousands of errors a day caused by `navigator.userAgentData.brands` not being an array. This PR aims to fix that.

### DIFF
--- a/src/utils/userAgent.js
+++ b/src/utils/userAgent.js
@@ -10,7 +10,7 @@ interface NavigatorUAData {
 export default function getUAString(): string {
   const uaData = (navigator: Navigator).userAgentData;
 
-  if (uaData?.brands) {
+  if (uaData?.brands && Array.isArray(uaData.brands)) {
     return uaData.brands
       .map((item) => `${item.brand}/${item.version}`)
       .join(' ');


atomiks:
Where does this error surface?

huntharo:
> Where does this error surface?

It throws an exception when calling .map on something that is not an array.

Could you clarify if you were looking for something more specific?

atomiks:
I'm fine with merging this, I'm just trying to understand how this is happening and how many people it affects.

I understand that it's erroring, but that should be impossible (in theory). What environment causes `navigator.userAgentData.brands` to exist, but `brands` to not be an array AND some kind of truthy value, especially since it's a new API - are you capturing this in Sentry or something? What value is `brands` in the env where it's erroring? The API docs says it is an array (although technically experimental).

arishoham:
We are using New Relic, our logs don't tell us much about what's going on, but here's a screenshot of one of the error instances. 
image

atomiks:
Going to just merge, since this check isn't critical anyway, it only avoids a message in DevTools

We'll try to release this soon 🕐  , in the meantime you can always use [patch-package](https://www.npmjs.com/package/patch-package) to fix dependencies while you wait for a new release or PR to get in for strange cases like this

PR #1976, opened 2022-11-25
Fix link color of "open collective" on website index page 

dirkjf:
image

instead of

image


atomiks:
Thanks not sure how that happened 😄

PR #2001, opened 2022-12-05
feat(react-dom): `open` option & `isPositioned` return value

atomiks:
Right now, the only fully reliable way to ensure the position is ready during an effect is to use `requestAnimationFrame` to prevent unwanted scrolling to the top/left of the document (due to the floating element's initial layout):

```js
useLayoutEffect(() => {
  open && requestAnimationFrame(() => el.focus());
}, [open]);
```

(I know `preventScroll` exists, but it's not very widely supported).

This feels kind of like a hack and can cause some issues, so I am proposing the following:

```js
const [open, setOpen] = useState(false);

const {isPositioned} = useFloating({
  // if false, resets the state
  open,
});

useLayoutEffect(() => {
  isPositioned && el.focus();
}, [isPositioned]);
```

The `open` option allows you to subscribe to resets, where `false` will cause a reset. This plays nicely with the main `react`/interactions package, because it already accepts an `open` option (and will default to this behavior), but also allows consumers of only the positioning library (`react-dom`, not `react`) to hook into resets declaratively.

### DIFF
--- a/packages/react-dom/index.test-d.tsx
+++ b/packages/react-dom/index.test-d.tsx
@@ -5,7 +5,7 @@ App;
 function App() {
   const arrowRef = useRef(null);
   useFloating();
-  const {reference, floating, update} = useFloating({
+  const {reference, floating, update, reset} = useFloating({
     placement: 'right',
     middleware: [
       shift(),
@@ -42,6 +42,7 @@ function App() {
   reference(null);
   floating(null);
   update();
+  reset();
   return <div ref={arrowRef} />;
 }
 
--- a/packages/react-dom/index.test-d.tsx
+++ b/packages/react-dom/index.test-d.tsx
@@ -5,7 +5,8 @@ App;
 function App() {
   const arrowRef = useRef(null);
   useFloating();
-  const {reference, floating, update, reset} = useFloating({
+  const {reference, floating, update} = useFloating({
+    open: true,
     placement: 'right',
     middleware: [
       shift(),
@@ -42,7 +43,6 @@ function App() {
   reference(null);
   floating(null);
   update();
-  reset();
   return <div ref={arrowRef} />;
 }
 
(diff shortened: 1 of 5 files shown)


atomiks:
@lozinsky, I discussed with @FezVrasta  and we're likely making this change to make `x` and `y` default to `0` instead of `null`, you can see the changelog as to why. I think this affects the Vue package if they should match

lozinsky:
@atomiks yeah they should match. I'll make the same changes. Thank you ☺️

atomiks:
Hi @benoitgrelard 😄 , does this API look sound? It's especially meant to help in calling `.focus()` or `.scrollIntoView()` at the right time (`isPositioned` lets you wait for the floating element to move from the top/left initial position, to the proper position, to prevent unwanted scrolling to the top of the page, etc.)

atomiks:
The `open` option has issues with unmounting animations (not `framer-motion`, but `react-transition-group` `CSSTransition`) because the top/left state gets set to `0` while the close animation plays.

The other solution is to check if the `node` becomes `null` in the floating ref callback, so we know it's unmounted from the DOM for certain and can reset the initial data, but this has the caveat that unstable ref props will cause an infinite loop because React calls the `ref` with `null` and then with the new node, and so on in a loop, so we can't make any `setState` calls.

```js
<div
  // Works
  ref={floating}
  // Infinite loop, user must always memo the value
  ref={node => floating(node)}
/>
```

So could make it opt-in as `resetOnUnmount: true` or such, and explain the ref callback must be memo'd when the option is enabled, unless anyone has better ideas

atomiks:
I've decided to keep `number | null` for now because of some edge cases with SSR and unmounting animations, but can open an issue regarding those if we find a good solution.

atomiks:
Checking `x` and `y` works in most cases, but it has an edge case:

After positioning, if the reference element moved somewhere else to a different scroll location on the page, then the .focus() call will scroll to the old floating element position, causing the issue (just not the top-left 0,0 one). `isPositioned` ensures that also doesn't happen when the `open` option is specified. This does depend on your component API though, if `useFloating()` unmounts when the floating element also unmounts, that won't be an issue.

benoitgrelard:
Good to know though, has this change been released?

atomiks:
It's in the latest version yeah

PR #2016, opened 2022-12-08
Treat 'pen' pointerType like 'mouse' in useHover hook

Slapbox:
I made these changes in the github.dev editor, so apologies if any linter requirement might not be met.

I made these changes to the `dist/floating-ui.react-dom-interactions.esm.js` locally, tested them, and then copied the changes to the source file - so there's an exceedingly small chance there was a transcription error on my part - but I believe this code to be functional and ready to go.

@atomiks

### DIFF
--- a/packages/react-dom-interactions/src/hooks/useHover.ts
+++ b/packages/react-dom-interactions/src/hooks/useHover.ts
@@ -24,12 +24,14 @@ interface HandleCloseFn<RT extends ReferenceType = ReferenceType> {
   };
 }
 
+const mouseLikePointerTypes = ['mouse', 'pen'];
+
 export function getDelay(
   value: Props['delay'],
   prop: 'open' | 'close',
   pointerType?: PointerEvent['pointerType']
 ) {
-  if (pointerType && pointerType !== 'mouse') {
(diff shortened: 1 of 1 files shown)


Slapbox:
Good catch @atomiks! I accidentally made that typo and "fixed it" but I must have hit Ctrl+Z once too often and undid it.

Slapbox:
@atomiks I'm not sure if it would be `undefined` or an empty string. Do you still want me to stick `undefined` in there, or should we handle that as a separate issue if there's some uncertainty.

atomiks:
`pointerTypeRef.current` can be `undefined`, which is why we need it (if `onPointerEnter`/`onPointerDown` are never called). As for empty string, yeah I guess add that as well. I think it's better to fallback to showing the floating element than not showing it when it can't be determined. Hopefully all touch devices reliably send back `touch` though...

Slapbox:
Alright now we have both `undefined` and `''` in the array!

> Hopefully all touch devices reliably send back `touch` though...

One can never be certain, but I think it's a safe assumption. I think it's much less likely for `touch` devices to have any weirdness  like my edge case involving `pen` vs `mouse`.

PR #2019, opened 2022-12-09
fix(dom): use computed width/height for scale

### LINKED ISSUE #2012: [DOM] The positioning calculation is incorrect in the scale scene
## Reproduction link
https://github.com/clfeng/floating-ui-scale-bug-demo

### clone project
git clone https://github.com/clfeng/floating-ui-scale-bug-demo.git

### Startup project
yarn install;
yarn dev;

### explain 
The project provides two pages; 
http://localhost:5173/#/origin: This page is an example of the effect of float-ui without body scaling

http://localhost:5173/#/scale：This page is an example of the body scaling effect of float-ui

Through the comparison of the above two examples, it can be clearly seen that the position calculation in the scale scene is incorrect

more detail: https://github.com/clfeng/floating-ui-scale-bug-demo

### Environment Info

Browsers:
    Chrome: 107.0.5304.121
    Firefox: 77.0.1
    Safari: 14.1.2
  npmPackages:
    "@floating-ui/dom": "^1.0.8"

atomiks:
Closes #2012  @sallerli1 @clfeng


atomiks:
So I found two edge cases here:

- 3D transforms make the matrix a bit different
- If the reference is scaled AND the offsetParent is scaled, the logic breaks (which was already the case before too)

atomiks:
This new commit fixes all the different issues (as far as I can tell), but hurts performance a bit unfortunately.

With a 40 node-deep tree of `position: relative;` parents, the updates take ~2.2ms while before they're only ~0.7ms. 

So I'm thinking that we should use the old method which is prone to small abberrations (and thus incorrect positioning) with fractional values, and let the user manually pass in `getScale: () => ({x: number, y: number})` instead.

Issue #2021, opened 2022-12-09
useDismiss outsidePress is not reactive


mihkeleidast:
This is an issue that was introduced in https://github.com/floating-ui/floating-ui/pull/2010.

My usecase is that on large screens, the floating element is rendered inline, i.e. it is not actually floating. On smaller screens, it transforms to a floating element, rendered like a modal. This is controlled by reactive resize handlers (we use react-responsive library, but that's not too relevant). This means that the `outsidePress` value might change while the floating element is open, but currently this does not update the outside press event listeners, so one of two issues can happen:

1. When resizing the window from large to small (i.e. component changes from non-floating to floating), outside press does not work as expected.
2. When resizing the window from small to large (component changes from floating to non-floating), outside press handlers stay attached, meaning a click outside will close the floating element, even though it is not expected.

It's configured like this:
```
export const useDrawer = () => {
  const [open, setOpen] = React.useState(false);
  const { variables } = useTheme();
  const isLargeScreen = useMediaQuery({ query: variables.media.width.xlarge.min });
  const isModal = !isLargeScreen;

  const data = useFloating({
    open,
    onOpenChange: setOpen,
  });

  const context = data.context;
  const click = useClick(context);
  const dismiss = useDismiss(context, {
    outsidePress: !isLargeScreen,
    outsidePressEvent: "mousedown",
  });
  const role = useRole(context);

  const interactions = useInteractions([click, dismiss, role]);

  return React.useMemo(
    () => ({
      open,
      setOpen,
      modal: isModal,
      ...interactions,
      ...data,
    }),
    [open, setOpen, isModal, interactions, data],
  );
};
```

Any way that change could be reverted or made smarter in this regard, i.e. keep the boolean value reactive?

atomiks:
"Doubt this matters." LOL, well.... 😭

PR #2022, opened 2022-12-09
feat: allow disabling focusout event listeners in FloatingFocusManager

### LINKED ISSUE #2017: FloatingFocusManager focusout handler - closes when clicking another button, does not close on outside click
I created a small reproduction in a branch in my fork: https://github.com/mihkeleidast/floating-ui/commit/c5b92d4b029d8902e5b5a3178cb4cb4b3072ad37

Ran across an inconsistency in how the focusout handler works. Two different scenarios at play here:

Scenario 1
1. Open Popover from the button
2. Click outside of the Popover, observe that it does not close
3. Click the "Second button", observe that the Popover remains open

Scenario 2
1. Open Popover from the button
2. Click the "Second button" observe that the Popover closes

As far as I debugged, the focusout handler fires in both cases when clicking outside of the floating element, but in scenario 1, when clicking the background, `relatedTarget` is `null`, so it does not fire the open change event.

It is kind of an edge case, but would be good if the behavior was consistent, to know what I should expect, as from the users perspective a random click outside should not change anything. 

Some additional info: in my real usecase, I do not have such Popovers, rather it is a drawer-like component (ex https://shoelace.style/components/drawer) that can be configured to not render on top of other content, rather it would render aside the main page. In this context, I have floating-ui configured to not close the floating element on outside click, but this issue means it still does in some cases. I can create a more realistic sandbox, if necessary, though I think the above describes the issue in enough detail.

mihkeleidast:
Probably resolves #2017.

I added a somewhat realistic Drawer component to the playground as well, so the necessary behavior could be tested against and understood more easily. I don't really want to add too much of my specific components here, if they seem too much, so let me know if I should remove the example.

Additional comments:
- The added example can also be used to demo & test https://github.com/floating-ui/floating-ui/issues/2021 very well
- I kind of assumed that disabling the focusOut behavior here would keep the floating element open once I tab out of it, but it seems that there is another close hook in the focus guard that is rendered by FloatingPortal. Not too big of an issue at the moment, but maybe there needs to be another prop on FloatingPortal as well?


atomiks:
I think it should be called `closeOnFocusOut`, it also only applies to non-modal focus management but the modal prop has some side-effects that should be maintained here?

mihkeleidast:
> I think it should be called closeOnFocusOut

Renamed 👍 

> it also only applies to non-modal focus management but the modal prop has some side-effects that should be maintained here?

Did not quite get this, what sort of side effects of the modal prop should be mentioned here?

atomiks:
This prevents screen readers from accessing outside content. Usually they won't be using `modal={true}` and `closeOnFocusOut={true}` together, but in case they are it should be guarded.

https://github.com/floating-ui/floating-ui/blob/bc3fb7cb262340187c1847ce2c9d1a0e7877c246/packages/react-dom-interactions/src/FloatingFocusManager.tsx#L255-L262

Btw, the docs will update before release, so usually have to do them separately

mihkeleidast:
> Usually they won't be using modal={true} and closeOnFocusOut={true} together, but in case they are it should be guarded.

I assume you meant `modal={true} closeOnFocusOut={false}`, as both true is the default? Looking at the hideOthers/cleanup logic, why is it in this focusout effect at all, shouldn't it be in a separate effect?

I'll move the docs to a separate PR 👍

mihkeleidast:
Docs are now in #2023. I also moved the hideOthers/cleanup logic to a separate effect, I don't think it should be disabled together with the rest of closeOnFocusOut. Not sure about all the hook deps, though - there were previously some that weren't used in the effect, not sure if they were deliberately set there for some side effects or not, you probably now best.

packages/react-dom-interactions/package.json
```diff
@@ -85,7 +85,9 @@
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
     "react-merge-refs": "^1.1.0",
+    "react-responsive": "^9.0.2",
```
atomiks, line 88: Why's this here?

mihkeleidast: I added it to properly showcase a component that changes the floating hook & component config based on screen size. Could also remove it and write a one-off resize handler for this, if adding it is too much.

PR #2023, opened 2022-12-09
docs: closeOnFocusOut on FloatingFocusManager

mihkeleidast:
Docs for #2022.


atomiks:
Need to add:
```js
<Notice type="version">Available from v0.14.0</Notice>
```

PR #2027, opened 2022-12-10
refactor: rename `react-dom-interactions` to `react`

atomiks:
### Why?

- The package name was really long and unwieldy (super annoying to type imo and naming is important)
- Part of Floating UI’s rebranding was to help indicate it offers more than positioning, so this helps focus on that. Especially now that the interactions part is a bit more mature at this point. For example, libraries like `react-laag` combine both positioning and interactions into one library, so it'll be kind of like that.
- Interactions for React Native are likely not coming soon if ever due to lack of contribution (I do not work with RN, so it’d need to be maintained by someone else. That person has not come forward yet). If they do come, they’ll likely be packaged under the `@floating-ui/react-native` package along with the positioning.
- The `@floating-ui/react-dom` package will continue to exist just for positioning, as it’s a lean install (no other dependencies) for libraries like Radix. They also don’t need to worry if their consumer’s tree-shaking setup is broken for whatever reason, which is important for the web.

Pinging for extra opinions from @mihkeleidast @FezVrasta


FezVrasta:
Sound good to me 👍👌

PR #2029, opened 2022-12-11
Try out `esm-env`

atomiks:
I'm not sure if this is fully backwards-compatible, but it seems like a better solution than the one we came up with @Westbrook 

Will need to validate this against bundlers etc to ensure it actually eliminates the dead code when building for production, and keeps it in dev mode.

Reference: https://twitter.com/Rich_Harris/status/1601609347376652290

### DIFF
--- a/packages/core/rollup.config.js
+++ b/packages/core/rollup.config.js
@@ -41,6 +41,9 @@ const bundles = [
       name: 'FloatingUICore',
       file: path.join(__dirname, 'dist/floating-ui.core.umd.js'),
       format: 'umd',
+      globals: {
+        'esm-env': 'esmEnv',
+      },
     },
   },
   {
@@ -49,6 +52,9 @@ const bundles = [
       name: 'FloatingUICore',
       file: path.join(__dirname, 'dist/floating-ui.core.umd.min.js'),
       format: 'umd',
+      globals: {
+        'esm-env': 'esmEnv',
+      },
     },
   },
 ];
@@ -56,6 +62,7 @@ const bundles = [
 export default bundles.map(({input, output}) => ({
   input,
   output,
+  external: ['esm-env'],
   plugins: [
     nodeResolve({extensions: ['.ts']}),
     replace({
--- a/packages/core/rollup.config.js
+++ b/packages/core/rollup.config.js
@@ -62,7 +62,7 @@ const bundles = [
 export default bundles.map(({input, output}) => ({
   input,
   output,
-  external: ['esm-env'],
+  external: output.format !== 'umd' ? ['esm-env'] : [],
   plugins: [
     nodeResolve({extensions: ['.ts']}),
     replace({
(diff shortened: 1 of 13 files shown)


Westbrook:
This is definitely an interesting way to get rid of the `process.env` reliance, but it would be a bit of a breaking change, likely something to target when you go 2.0? It move ALL of the management of which code to load/build/dead code eliminate down to the final consumer, which ensure's they'll have the clearest path to the smallest build when they do build, but it doesn't line up to the multiple exports that are being leveraged now with `.mjs`,`.umd.js`, `.esm.js`, `.browswe.js`, and `.browser.min.js` all being available depending on consumption context.

benmccann:
> it doesn't line up to the multiple exports that are being leveraged now with .mjs,.umd.js, .esm.js, .browswe.js, and .browser.min.js all being available depending on consumption context.

I think this approach is compatible with multiple export targets. For the `.umd.min.js` file (and possibly `.browser.mjs` and `browser.min.js` if they're used by CDNs) since since it is used by a CDN if needs to be bundled ahead of time with the node condition being resolved then. For the others, you can mark `esm-env` as external leaving the node condition to be resolved by the end user.

> It didn't work properly with CJS so it made it non-viable

Hi @atomiks, I happened to stumble across this PR. I wrote `esm-env`. If you have an example where it doesn't work, I'd be interested in fixing it. I'm not aware of any issues using it with CJS currently and haven't gotten any bug reports about it, so am not quite sure how to address your issue otherwise.

packages/core/rollup.config.js
```diff
@@ -49,13 +52,17 @@ const bundles = [
...
+        'esm-env': 'esmEnv',
+      },
     },
   },
 ];
 
 export default bundles.map(({input, output}) => ({
   input,
   output,
+  external: output.format !== 'umd' ? ['esm-env'] : [],
```
benmccann, line 65: I think this isn't quite right and is why you'd have been having issues with the CJS version. This will replace `DEV` at bundle time for the file that you have listed in `main`. That's fine if you want to strip the dev stuff out at build time and it's only for your own use, but will not work if you want consumers of this library to be able to see dev-time warnings

You should probably include `&& output.file.includes('.min.')` in the condition here, so that the dev only gets resolved at bundle time for the `unpkg` users

benmccann:
omI think I just found the issue +89i-9*h/-=

benmccann:
I just took a closer look at this PR and think I see the issue you were likely having: https://github.com/floating-ui/floating-ui/pull/2029/files#r1148416432

What do you think about reopening this PR with the change I suggested?

We've had longstanding complaints about this library not working in the Svelte REPL because of `process.env` and I'd love to help get it fixed!

atomiks:
@benmccann feel free to make a PR if you want that works for both the esm and umd builds, would be cool to get it working

benmccann:
I won't bother given https://github.com/floating-ui/floating-ui/pull/2251, but if you ever want to try again in the future, I'm happy to help

PR #2034, opened 2022-12-12
Introduce `@floating-ui/vue` package

lozinsky:
This sources from the [`vue-floating-ui`](/lozinsky/vue-floating-ui/tree/main/packages/vue-floating-ui) package. So far, it provides positioning only.

Next, I'd like to research the possibility of adding interactions to this package. But I'm not sure that I can keep compatibility with Vue 2. Because development of components with `vue-demi` is too hard https://github.com/vueuse/vue-demi/issues/153.

- [x] `vue` installs with their own types for JSX that conflict with `react` (unfortunately I had to use `patch-package` to remove Vue JSX types https://github.com/vuejs/core/issues/1033)


atomiks:
It looks great to me, but my Vue knowledge is very limited. Anyways, if there are any issues, people can file them 😄  In the meantime, we can start it as `v0` major until a few months have passed with it being vetted by lots of usage, then bump it to `v1`

renatodeleao:
@lozinsky just want to thank you for doing this! Had to spin up my own package as well ([`floating-ui-x`](https://github.com/renatodeleao/floating-ui-x/tree/main/packages)), but it's much better to take advantage from the popularity of this repository: more people, more contributors = better code!

Developing Vue2/3 hybrid is indeed very painful when components are involved. Composables are "managable" from `vue@>2.7.x` but not sure if we should add that maintenance burden at this point. Been contributing to [a few](https://github.com/Grsmto/simplebar/pull/630) [packages](https://github.com/ndelvalle/v-click-outside/pull/563) with that approach and my advice is that is just not worth it. It's a rabbit hole, especially when it comes to testing.

I will try to help with a few PRs for the interactions package, made very basic attempts [here](https://github.com/renatodeleao/floating-ui-x/tree/main/packages/vue-interactions) but never got to the publishing part. 

Cheers! ✌️

PR #2038, opened 2022-12-16
docs: fix nav hierarchy

mihkeleidast:
Fixes the docs nav hierarchy. Right now, all the react hooks and examples are nested under "React Native", which seems incorrect.

Before:
image

After:
image


atomiks:
Good point, especially after we added 'Vue' below everything else.

PR #2040, opened 2022-12-16
feat(useDismiss): allow specifying different value for outside press bubbling and escape key bubbling

mihkeleidast:
Given a nested Popover pattern ([example](https://codesandbox.io/s/goofy-sunset-pplzyj?file=/src/Popover.tsx)), it is generally preferred to disable escape key close bubbling, as that is considered an explicit request to close the immediate open dialog (you say as much in the docs, I believe). However, the current `bubbles` flag also disabled bubbling of the outside press event. This seems counterintuitive for mouse users - if I intentionally click outside both of the open Popovers, I'd expect both of them to close as well. Currently, when disabling bubbling, it also necessitates multiple outside clicks, one for each level of open Popovers.

~~It seems like a good idea to allow configuring this for both events separately. I currently added a `outsidePressBubbles` option that inherits from the `bubbles` option. But it might be better to make it not inherit, and rename the current option to `escapeKeyBubbles` instead. I did not want to do that without an agreement, as that is a breaking change, albeit a small one.~~

This change allows specifying an object to the `bubbles` prop, to turn one or the other off specifically. Any omitted properties from the object default to `true`, i.e. the default value.

Also:
- ~~Any tests necessary for this?~~ Added tests as well.
- I'll file a separate PR for docs once this has a general approval: https://github.com/floating-ui/floating-ui/pull/2046

### DIFF
--- a/packages/react/src/hooks/useDismiss.ts
+++ b/packages/react/src/hooks/useDismiss.ts
@@ -38,6 +38,7 @@ export interface Props {
   referencePressEvent?: 'pointerdown' | 'mousedown' | 'click';
   outsidePress?: boolean | ((event: MouseEvent) => boolean);
   outsidePressEvent?: 'pointerdown' | 'mousedown' | 'click';
+  outsidePressBubbles?: boolean;
   ancestorScroll?: boolean;
   bubbles?: boolean;
 }
@@ -57,6 +58,7 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
     referencePressEvent = 'pointerdown',
     ancestorScroll = false,
     bubbles = true,
+    outsidePressBubbles = bubbles,
   }: Props = {}
 ): ElementProps => {
   const tree = useFloatingTree();
@@ -154,7 +156,7 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
       }
 
       if (
-        !bubbles &&
+        !outsidePressBubbles &&
         tree &&
         getChildren(tree.nodesRef.current, nodeId).length > 0
       ) {
@@ -237,6 +239,7 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
     ancestorScroll,
     enabled,
     bubbles,
+    outsidePressBubbles,
     refs,
     nested,
   ]);
--- a/packages/react/src/hooks/useDismiss.ts
+++ b/packages/react/src/hooks/useDismiss.ts
@@ -38,9 +38,8 @@ export interface Props {
   referencePressEvent?: 'pointerdown' | 'mousedown' | 'click';
   outsidePress?: boolean | ((event: MouseEvent) => boolean);
   outsidePressEvent?: 'pointerdown' | 'mousedown' | 'click';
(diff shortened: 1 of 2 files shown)


atomiks:
I think `bubbles` should be of type `boolean | object` (where you can turn one or the either off) to avoid too many props

```js
bubbles: false // both don't bubble
bubbles: true // both bubble
bubbles: {} // both bubble (truthy)
bubbles: {
  // non-specified options are implicitly `true`
  escapeKey: false
}
```

mihkeleidast:
Thanks for the feedback, made the changes. The tests question is still open, I did not find a test for the bubbles prop at the moment, not sure how easy it is to add tests for this behavior?

packages/react/src/hooks/useDismiss.ts
```diff
@@ -71,6 +71,18 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
...
+      ? bubbles
+      : bubbles.escapeKey === false
+      ? false
+      : true;
+  const outsidePressBubbles =
+    typeof bubbles === 'boolean'
+      ? bubbles
+      : bubbles.outsidePress === false
+      ? false
+      : true;
```
atomiks, line 85: This should work:
```js
const allBubble = typeof bubbles === 'boolean';
const escapeKeyBubbles = allBubble ? bubbles : bubbles.escapeKey;
const outsidePressBubbles = allBubble ? bubbles : bubbles.outsidePress;
```

atomiks:
It looks like only `test('always returns to the reference for nested elements',` (in `FloatingFocusManager.test.tsx`) has a test for the feature, but would be great to test it properly in `useDismiss.test.tsx`

packages/react/src/hooks/useDismiss.ts
mihkeleidast: For these scenarios:
```
bubbles: {} // both bubble (truthy)
bubbles: {
  // non-specified options are implicitly `true`
  escapeKey: false
}
```

Wouldn't this change cause omitted properties in the object to be defined to `false`, not `true`?

atomiks: Yeah, so I guess you'd add ` || true` at the end.

mihkeleidast: Of course, always forget about nullish coalescing. Will change!

mihkeleidast: Changed - only thing I wasn't able to do now was defining the `allBubbles` check as a separate const. The code works in newer versions of TS, but not in 4.1 that is installed here currently, so it started erroring in tests.

mihkeleidast:
Alright, added a bunch of tests as well, separately for the prop resolution and for the actual behavior. Most of the test content/component I borrowed from the FloatingFocusManager test, cleaned that up as much as possible though.

atomiks:
Oh yeah, that TS 4.1 typeof problem is annoying :\ we can technically upgrade I think, but I kept it to 4.1 to ensure we don't use anything the consumer can't also use if their TS version is 4.1 / lower than ours.

PR #2042, opened 2022-12-16
fix(vue): Add a type to vue arrow middleware

ErKeLost:
Add a type to vue arrow middleware


atomiks:
what's the purpose of this?

PR #2055, opened 2022-12-21
fix(FloatingDelayGroup): default delay in context

mihkeleidast:
Sets the default delay in the delay group context to 0, which is the same default delay used by `useHover`.

I was a bit surprised today that all my Tooltips got a delay after implementing the opt-in support for the delay group functionality. That is - I added the `useDelayGroupContext` and `useDelayGroup` hooks to my component, and I expect the delay group provider itself to be added as some outer wrapper, which I think is the expected implementation.

I think these should be 0, as in no delay, otherwise I don't understand how to default to 0. If one needs a different default delay, that's easier to implement in the custom Tooltip component that uses the hooks.

### DIFF
--- a/packages/react/src/components/FloatingDelayGroup.tsx
+++ b/packages/react/src/components/FloatingDelayGroup.tsx
@@ -21,8 +21,8 @@ const FloatingDelayGroupContext = React.createContext<
     setState: React.Dispatch<React.SetStateAction<GroupState>>;
   }
 >({
-  delay: 1000,
-  initialDelay: 1000,
+  delay: 0,
+  initialDelay: 0,
   currentId: null,
   setCurrentId: () => {},
   setState: () => {},


mihkeleidast:
> `<FloatingDelayGroup delay={0}` doesn't work? I agree with this regardless though

Yes but that means I need to wrap all tooltips in a delay group, or add this as an external wrapper for my entire application.

mihkeleidast:
@atomiks any chance of getting this released today, along with some of the other fixes? Would help me cross off quite a few bugs in my system :)

atomiks:
The next release of the packages is going to be delayed, probably not right until the end of the year/new year, because I have a big list of things I want to get in, and don't want to make a partial release (it's not possible to do individual releases of packages with Rolling Versions if others have merged changes). 

I would recommend using `patch-package` temporarily, patching the lib for now until next release

mihkeleidast:
Alright, thanks for the update!

PR #2076, opened 2022-12-28
fix: size modifier side check in calculations

heydarm:
Hi,

I just noticed there's a little 🐛 crawling around here.
`placement` represented as a string `{side}-{alignment}`, so searching that in `[left, right, top, bottom]` will not work as expected.
I replaced it with just the `side` variable, that is more appropriate for that union.

### DIFF
--- a/packages/core/src/middleware/size.ts
+++ b/packages/core/src/middleware/size.ts
@@ -66,7 +66,7 @@ export const size = (
     const dimensions = {
       availableHeight:
         rects.floating.height -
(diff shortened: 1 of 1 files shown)


atomiks:
Thanks for the PR, but the code is not checking for the side of the placement, rather whether it doesn't have an explicit alignment specified (i.e. center alignment). So the current code is correct, which is why the functional tests failed

heydarm:
@atomiks oh, now I got it 😅
Thank you very much for explanation
Have a great day

PR #2082, opened 2023-01-01
feat(flip): `fallbackAxisDimension` option

### LINKED ISSUE #2051: [Core] `flip` should allow axis change by default
Let's say I have this generic Popover component that allows specifying a placement, but defaults to "auto":

```tsx
import { Placement, flip, useFloating, autoPlacement } from '@floating-ui/react-dom-interactions';

export type PopoverPlacement = 'auto' | Placement;

interface Props {
  /**
   * Placement of the tooltip (default: 'auto')
   */
  placement?: PopoverPlacement;
}

const Popover = ({
  placement = 'auto',
}: Props) => {
  const [open, setOpen] = React.useState(false);
  const data = useFloating({
    placement: placement === "auto" ? undefined : placement,
    open,
    onOpenChange: setOpen,
    middleware: [
      placement === "auto" ? autoPlacement() : flip(),
    ],
  });

  return (
    <div />
  );
};
```

In order to allow both a specific placement and a specific placement, this conditional logic is necessary in several places, which already feels cumbersome to use TBH - 

Now, one instance of this component has specified `<Popover placement="right"`. This means that the flip logic will turn on instead of the fully automated logic. On some screen sizes, however, it won't fit on the right,. It won't fit on the left, either - which is the only side the flip modifier will check by default. The only option where it can fit is "top". Right now flip checks if it can fit on right, or left, which it does not, so it will put it on the right, which will still cause overflow.

Now, I know it is possible to specify the fallback placements to the modifier. To do this, there are three options:
1. I might make the fallback placements another prop on my component and leave this configuration fully to the consumers of my component.
2. I might create a static map of fallback placements and use it in my component, passing it always forward to floating-ui.
3. The static map of fallback placements could live in floating-ui, resolving such issues for all consumers of this library.

My question is - why not 3, why should this kind of logic have to live on the consumer side? I don't think anyone realistically wants to get overflow if there is any chance of avoiding it by changing the placement automatically. What I as a consumer of this lib would want, is to have a single placement modifier, that I can nudge to my preferred placement - but any other would also be fine. As an alternative, some simpler way of doing `fallbackPlacement: [oppositePlacement, ...restOfPlacements]` would also work, but as this needs to be computed based on the preferred placement, to me it seems like it will potentially be duplicated in a lot of codebases (or if not implemented, the components can be more buggy). Any thoughts?

atomiks:
Closes #2051  

@mihkeleidast 

See changelog below

### DIFF
--- a/packages/core/src/middleware/flip.ts
+++ b/packages/core/src/middleware/flip.ts
@@ -2,7 +2,7 @@ import {
   detectOverflow,
   Options as DetectOverflowOptions,
 } from '../detectOverflow';
-import type {Alignment, Middleware, Placement} from '../types';
+import type {Middleware, Placement} from '../types';
 import {getAlignmentSides} from '../utils/getAlignmentSides';
 import {getExpandedPlacements} from '../utils/getExpandedPlacements';
 import {getOppositeAxisPlacements} from '../utils/getOppositeAxisPlacements';
@@ -31,16 +31,17 @@ export interface Options {
    */
   fallbackStrategy: 'bestFit' | 'initialPlacement';
   /**
+   * Whether to allow fallback to the opposite axis, and if so, which
+   * dimension along the axis to prefer.
+   * @default 'none' (disallow fallback)
+   */
+  fallbackAxisDimension: 'none' | 'start' | 'end';
+  /**
    * Whether to flip to placements with the opposite alignment if they fit
    * better.
    * @default true
    */
   flipAlignment: boolean;
-  /**
-   * Whether to flip the axis, and if so, decide the direction to prefer.
-   * @default true
-   */
-  flipAxis: false | Alignment;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


atomiks:
In terms of creating "good" defaults, a component like a Tooltip or Popover might do this:

- Allow the user to override `flipAxis` on a per-usage basis
- Prefer `top` when the axis must flip, e.g. on mobile (tooltips are often on top)
- Don't allow the axis to flip if the preferred side is `top`/`bottom`, so `shift()` can take action instead of `flip()` to preserve the side

```js
function Tooltip({placement, flipAxis}) {
  const side = placement.split('-')[0];
  flip({
    // never flip the axis if the side is top/bottom
    // prefer the tooltip on top on mobile viewports
    flipAxis: flipAxis ?? ['top', 'bottom'].includes(side) ? false : 'start',
  });
}
```

mihkeleidast:
Nice!

atomiks:
I've decided this should be renamed before release, as reading about this more, I don't know if this option name is appropriate... this is kind of a hard concept to name

I first tried:

- `flipAxis`

Then came to:

- `fallbackAxisDimension`

Although, `fallbackAxisPreferredDimension` is technically better, but it's too long.

But now reading into "dimension" and "direction" I feel like "direction" is better than "dimension". But I also feel it's not fully appropriate either...

- `fallbackAxisDirection`
- `fallbackAxisSideDirection`

PR #2083, opened 2023-01-01
`open` option and `isPositioned` return value for Vue

lozinsky:
Sync with https://github.com/floating-ui/floating-ui/pull/2001

### DIFF
--- a/packages/vue/src/useFloating.ts
+++ b/packages/vue/src/useFloating.ts
@@ -36,6 +36,7 @@ export function useFloating<T extends ReferenceElement = ReferenceElement>(
   options: UseFloatingOptions<T> = {}
 ): UseFloatingReturn {
   const whileElementsMountedOption = options.whileElementsMounted;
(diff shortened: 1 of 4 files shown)


packages/vue/src/useFloating.ts
```diff
@@ -92,10 +95,22 @@ export function useFloating<T extends ReferenceElement = ReferenceElement>(
...
   }
 
+  function reset() {
+    if (!openOption.value) {
+      x.value = null;
+      y.value = null;
+      strategy.value = strategyOption.value;
+      placement.value = placementOption.value;
+      middlewareData.value = {};
+      isPositioned.value = false;
```
atomiks, line 105: Due to unmounting animations, only `isPositioned` should get reset. Otherwise it goes back to (0, 0) while it's animating out (depending on how the animation is performed, with CSS I noticed it in React, but not framer-motion).

lozinsky: Got it.

PR #2089, opened 2023-01-03
feat(react): `useTransition` hooks

### LINKED ISSUE #2088: Handling transitions
## Goal

I'm attempting to implement in-out transitions on a Tooltip component. I'd like to understand if there is some preferred library for implementing such transitions, or if it would be feasible to support this directly in floating-ui.

Sort of a discussion really - I did find https://github.com/floating-ui/floating-ui/discussions/1873, but that is also not solved.

## Current state

Example implementation I currently have: https://codesandbox.io/s/beautiful-jennings-zm83g8?file=/src/Tooltip.tsx

It uses the [react-transition-state](https://github.com/szhsin/react-transition-state) library that provides a hook that handles the animation. To be fair, I'm not too happy with that lib, but I haven't found any better hook-based libraries either. I suppose I could try out [react-transition-group](https://github.com/reactjs/react-transition-group) as well, though I do not like that component-based approach at all in 2023.

So, the Tooltip component allows both "auto" placement and some specific placement. If a specific placement is set, the positioning and transitions work fine. With the auto placement, however, the first transition is always with `placement="bottom"`, as the auto placement needs time to calculate the final placement. So this takes the transition and actual placement out of sync - the Tooltip is placed on the right, but the transition is for bottom. I assume that flip can also have this issue, whenever it actually needs to flip to some alternative placement.

I do know that there is now the `isPositioned` boolean returned from the hook that is true only when the final placement is calculated. So I attempted to make the transition state dependent on that: https://codesandbox.io/s/gracious-sound-9c52jf?file=/src/Tooltip.tsx

This version does not have the "in" transition for some reason. It might have something to do with the library I'm using for handling the transition, I don't really know. Either way it's quite cumbersome to mix'n'match these different libraries, it turns into a state variable conditional hell pretty quickly.

## Feature request?

I would assume that such transition and animation needs are quite common to be implemented together with different floating elements - at least in my memory this is usually the number 1 feature request from UI designers that I have to implement. So, it would be nice if we could provide some guidance and/or examples how these could be implemented - i.e. which hook/component would be a good match to use together with floating-ui for this.

Perhaps, if there is no good match out there, it could also be a new hook that lives inside floating-ui? As the positioning logic has to be quite tightly tied to the animation and mounting states, this might be a good approach?

```
const { isMounted, transitionState } = useFloatingTransition({
  open,
  placement,
  isPositioned,
  timeoutMs: 100,
});
```

atomiks:
Closes #2088

This enables easy creation of CSS transitions if not using a spring library like `framer-motion`, and also, easily enables side-aware transitions.


mihkeleidast:
Absolutely fantastic, I did not expect you'd implement this so fast!

Some thoughts:
- I'm not fully onboard with the fact that this works only with inline styles. Is there some deeper reason for this, i.e. when using classes for different states react re-rendering stuff might miss some steps, or something like that? I feel like it could (at least optionally) also return the side/placement and transition direction (enter/exit) from the hook, so it could be implemented in CSS as well, if there are no other technical obstacles. This would also kind of resolve the TODO you have noted down -by allowing a more complex transition to be implemented in CSS directly. The hook would then only need to know the full duration of the transition, not the individual steps.
- For the from/to/exit props, maybe these could be collated into a single or two props? It sort of looks weird when all three are set, I would at least want to combine from and to to a single `enter` object.

So it's either:
```js
{
  enter: {...} // implicit from
}
```

```js
{
  enter: {
      from: {...},
      to: {...}
  }
}
```

```js
{
  enter: {
      from: {...},
      to: {...}
  },
  exit: {
      from: {...},
      to: {...}
  }
}
```

atomiks:
> I'm not fully onboard with the fact that this works only with inline styles. Is there some deeper reason for this, i.e. when using classes for different states react re-rendering stuff might miss some steps, or something like that?

I feel it's the most natural way to style it since you get the full programmatic power of JS to compute the styles. The docs also use inline styles for the arrow, however it is necessary for pseudo-elements, so a CSS version might look like:

- `[data-placement="..."]` + 
- `[date-state="..."]`

```css
.class[data-placement^='top'][data-state='initial'] {
  transform: translateY(5px);
}
.class[data-placement^='bottom'][data-state='initial'] {
  transform: translateY(-5px);
}
```

> For the from/to/exit props, maybe these could be collated into a single or two props?

I feel 3 states is the most appropriate, I guess the naming is confusing though, and could be:

- `initial`
- `to`
- `exit`

So `initial` => `to` => `initial` (default), or optionally, `initial` => `to` => `exit`.

With the `enter` API, it's verbose and somewhat confusing:

- `enter.from` => `enter.to` => `exit.from` (this state likely doesn't need to exist) => `exit.to`

And if you only specify `enter`, it doesn't really make sense that it would also symmetrically _exit_ back to it as well. Whereas the 3 states are clearer (at least `initial` instead of `from`).

`framer-motion` appears to have 3 states also:

- `initial`
- `animate`
- `exit`

```js
<motion.div
  initial={{ opacity: 0 }}
  animate={{ opacity: 1 }}
  exit={{ opacity: 0 }}
/>
```

mihkeleidast:
> I feel it's the most natural way to style it since you get the full programmatic power of JS to compute the styles.  The docs also use inline styles for the arrow

Fair,  I agree that for some places defining this stuff in JS is basically unavoidable, and then it is good to have the option to use this method. But I'm not sure if it needs to be a part of this hook - at least for the other hooks / style-related stuff, only the data is returned from the hook, and the actual styles need to be assigned by the user themselves.  So I'm not sure if the styles object needs to be returned by this hook - maybe it could be just one documented implementation example? That would allow users to use whatever their preferred method for defining this stuff is.

mihkeleidast:
Though, I tried to modify your implementation to only return state and it seems to suffer from the same non-final first placement (i.e. first transition is always bottom instead of right): https://codesandbox.io/s/thirsty-leaf-f6kt4z?file=/src/useFloatingTransition.ts

Tested the implementation in this PR as well , just copied the code over - this seems to work OK: https://codesandbox.io/s/objective-mestorf-zytb39?file=/src/Tooltip.tsx

atomiks:
I feel like the best API would be one that defaults to inline styles (which solves 90%+ of use cases that want a nice basic transition) but can be switched to the data-based API with an option in `useCSSTransition`, where you compute it yourself for more advanced use cases (including the ability to use CSS as in your example).

> The naming of things gets complicated fast here, I would likely prefer "unmounted" -> "entering" -> "entered" -> "exiting" -> "unmounted" flow - this would make more sense to me if looking at only the state & styles sets that need to be applied at every time.

Do you have an example implementation of that? Ideally it should be as easy as the inline way unless they want more info to make an asymmetric transition, but symmetric ones should be just as easy. 

Since the rest of the lib uses `open`/`close`, why not:
`closed` -> `opening` -> `open` -> `closing` -> `closed`

I guess something like:

```js
const {isMounted, status, side} = useCSSTransition(context, {
  styles: false,
});
```

to switch to the data API may work, whereas inline will have the three states with styles given to you:

```js
const {isMounted, styles} = useCSSTransition(context, {
  initial: ...
  // optional
  open: ...
  close: ...
});
```

I guess a symmetric transition should ideally only require this?:

```css
.class[data-state="transitioning"] {
  transition-property: opacity, transform;
  transition-duration: 250ms;
}

.class[data-state="initial"][data-placement^='top'] {}
.class[data-state="initial"][data-placement^='right'] {}
.class[data-state="initial"][data-placement^='bottom] {}
.class[data-state="initial"][data-placement^='left'] {}
```

mihkeleidast:
> > The naming of things gets complicated fast here, I would likely prefer "unmounted" -> "entering" -> "entered" -> "exiting" -> "unmounted" flow - this would make more sense to me if looking at only the state & styles sets that need to be applied at every time.
> 
> Do you have an example implementation of that? Ideally it should be as easy as the inline way unless they want more info to make an asymmetric transition, but symmetric ones should be just as easy.
> 
> Since the rest of the lib uses `open`/`close`, why not: `closed` -> `opening` -> `open` -> `closing` -> `closed`

Yeah, open/close naming would also work - I mainly meant that this kind of states would make the underlying state machine & how the different states come after one another well understandable just by reading the state names. Not so clear when they are named "to", "initial", "exit" - especially when exit is an optional alias for initial as well.

I forked and edited my previous implementation to use this naming scheme, a lot nicer to read: https://codesandbox.io/s/serene-shadow-1n39ev?file=/src/useFloatingTransition.ts

> I guess a symmetric transition should ideally only require this?:
> 
> ```css
> .class[data-state*="transitioning"] {
>   transition-property: opacity, transform;
>   transition-duration: 250ms;
> }
> 
> .class[data-state*="initial"][data-placement^='top'] {}
> .class[data-state*="initial"][data-placement^='right'] {}
> .class[data-state*="initial"][data-placement^='bottom'] {}
> .class[data-state*="initial"][data-placement^='left'] {}
> ```
> 
> I guess the state might need the transitioning word separate like `"intial transitioning"`

Yeah. Just by reading the styles here, I don't understand what the difference between "transitioning" and "initial" is, i.e. why couldn't they be the same? We kind of need two different words that would be clear, one that combines "open" and "closing" (your "transitioning"), and other that combines "opening" and "closing" (your "initial"). I personally would rather duplicate my css selectors, because that makes it a lot clearer when some specific rule is applied.

> I feel like the best API would be one that defaults to inline styles (which solves 90%+ of use cases that want a nice basic transition) but can be switched to the data-based API with an option in useCSSTransition, where you compute it yourself for more advanced use cases (including the ability to use CSS as in your example).

https://codesandbox.io/s/sweet-artem-ojjmt4?file=/src/useFloatingTransition.ts

I just wonder if it could be implemented already internally with two hooks, one that deals with the main state logic stuff, and a second one that additionally deals with the simple inline style API as well. That would give the user the choice of what to pick, and keep the options of both quite lean. So, in that example, `useFloatingTransition` is the main state logic hook, whereas `useFloatingTransitionStyle` calculates the inline style based on the first hooks return value.

Note that I did not test if `useFloatingTransitionStyle` works in that example, it is mostly a mock implementation, but I hope the idea is understandable.

mihkeleidast:
> I feel like the transition states are confusing though, as a user I would be confused why the transition styles would need to be set on open, not opening, and then why the close style also needs to be set on opening.

Because that's how quirky CSS transitions are when being applied to elements just inserted to DOM 😅 But yeah, I agree that it's confusing, but I'm not sure what the best approach is.

For a completely symmetrical transition, applying the transition styles conditionally does not make sense to me at all - I would expect I could just slap it on the main class together with the rest of the styles, so like this:

```
.Tooltip[data-state="open"][data-direction="from"],
.Tooltip[data-state="close"][data-direction="to"] {
  opacity: 0;
}
.Tooltip {
  transition: opacity 200ms;
}
```

But I know there are technical limitations to doing this, and I don't think this can be worked around in the hook code.

atomiks:
So I believe the statuses that are most intuitive are:

- `unmounted`
- `initial`
- `open`
- `close`

This also maps to the inline styles hook API.

Here are some docs which explain it:

## useCSSTransition

This hook provides a `status{:.const}` string that determines if
the floating element is in one of four states:

```ts
type Status = 'unmounted' | 'initial' | 'open' | 'close';
```

The floating element cycles through the order as seen below:

```text
unmounted -> initial -> open -> close -> unmounted
```

Like the interaction hooks, this hook accepts the
`context{:.const}` object returned from `useFloating(){:js}`:

```js /context/
function App() {
  const {context, placement} = useFloating();
  const {isMounted, status} = useCSSTransition(context);

  return (
    isMounted && (
      <div id="floating" data-status={status}>
        Tooltip
      </div>
    )
  );
}
```

Above, we apply a `data-status{:.keyword}` attribute to the
floating element. This can be used to target the transition
status in our CSS.

To define a symmetrical opacity fade CSS transition:

```css
#floating {
  transition-property: opacity;
}

#floating[data-status='open'],
#floating[data-status='close'] {
  transition-duration: 200ms;
}
#floating[data-status='initial'],
#floating[data-status='close'] {
  opacity: 0;
}
```

- `transition-property: opacity{:sass}` is applied to **all**
  states. This allows the transition to be interruptible.

For the status, DOM insertion CSS transitions can be a bit
unintuitive, but the statuses map to the following:

- `'closed'{:js}` represents the element is unmounted from the
  screen. No transitions or styles need to be applied in this
  state.
- `'initial'{:js}` indicates the floating element has been
  inserted into the DOM. At this point, we apply the initial
  opacity style.
- `'open'{:js}` indicates the floating element is in the open
  state (1 frame after `'initial'{:js}`) and begins transitioning
  in.
- `'close'{:js}` indicates the floating element is in the close
  state and begins transitioning out.

An asymmetric transition may look like:

```js
const {isMounted, status} = useCSSTransition(context, {
  duration: {open: 500, close: 250},
});
```

```css
#floating {
  transition-property: opacity, transform;
}
#floating[data-status='initial'] {
  opacity: 0;
  transform: scale(0);
}
#floating[data-status='open'] {
  opacity: 1;
  transform: scale(1);
  transition-duration: 500ms;
}
#floating[data-status='close'] {
  opacity: 0;
  transform: scale(2);
  transition-duration: 250ms;
}
```

packages/react/src/hooks/useCSSTransition.ts
```diff
@@ -0,0 +1,201 @@
...
+  openStyles?: CSSStylesProperty;
+  closeStyles?: CSSStylesProperty;
+  commonStyles?: CSSStylesProperty;
+}
+
+/**
+ * Provides styles to apply CSS transitions to a floating element, correctly
+ * handling placement-aware transitions. Higher-level wrapper around
+ * `useCSSTransition`.
+ * @see https://floating-ui.com/docs/useCSSTransition
```
mihkeleidast, line 108:
```suggestion
 * @see https://floating-ui.com/docs/useCSSTransitionStyles
```

mihkeleidast:
Looks pretty great now!

packages/react/src/hooks/useCSSTransition.ts
mihkeleidast: Oh it's on the same page, so the suggestion is invalid - I wonder if this could hash-link to the specific heading, though?

atomiks:
I added a `symmetric` option for the status API to make it less verbose, but I don't think it can work because of the `transition-duration` 🤦 I'll revert it

packages/react/src/hooks/useCSSTransition.ts
```diff
@@ -0,0 +1,197 @@
+import type {
+  FloatingContext,
+  Placement,
+  ReferenceType,
+  Side,
+} from '@floating-ui/react';
```
mihkeleidast, line 6: This import should be from relative file, no?
```suggestion
} from './types';
```

atomiks:
Placement-aware transitions don't actually seem to be working anymore

packages/react/src/hooks/useCSSTransition.ts
atomiks: I've linked to the hases but the hash logic on the docs is broken for random edge cases. The hashes need to handle duplicate headings but I haven't found a solution to this that doesn't break in weird ways...

atomiks:
- Changed `closed` to `unmounted` because I feel it'd be prone to typos and confusion with `close`.
- Removed `Styles` from the options as it's kind of needlessy verbose and other transition hooks from other libs omit it.

But yeah I think this is pretty much good now

packages/react/src/hooks/useCSSTransition.ts
```diff
@@ -0,0 +1,192 @@
...
+      setStyles((styles) => ({
+        transitionProperty: styles.transitionProperty,
+        ...commonStyles,
+        ...initialStyles,
+      }));
+    }
+
+    if (status === 'open') {
+      setStyles({
+        transitionProperty: Object.keys(openStyles).join(','),
```
mihkeleidast, line 160: One caveat here - as the styles in options are JS objects, then multi-word keywords would need to be transformed from camelCase to kebab-case, for example `backgroundColor` should become `background-color` in order for it to work. But I guess it's possible to pass `transitionProperty` manually in options in that case?

atomiks: That's a good point, `opacity` and `transform` are by far the most common transitioned properties so didn't consider that. I don't think they should have to manually specify it though, seems like poor DX.

atomiks: Another thing to consider is `transitionTimingFunction`, however this can be done in `commonStyles` (for a symmetric transition). This is done in the `Tooltip.tsx` file.

mihkeleidast: Yeah, I came across this when implementing a transition for a Drawer that is wrapped in FloatingOverlay. I want the Drawer to slide in from the side, and the background overlay to fade in - but when using opacity, that would also fade in the Drawer itself, so need to transition backgroundColor on that instead.  (Not using this hook for that, but since I'm using vanilla-extract for styles, ran across a similar conversion issue.)

atomiks: I'm assuming Tailwind users for instance would also use the lower-level `useCSSTransition` hook if they want to use its classNames. The status string should work fine for them to conditionally add or remove different classNames.

atomiks:
I can't help but feel maybe `useCSSTransition` should be named something more generic... like `useMotion`? And `useCSSTransitionStyles` as `useTransitionStyles`?

The reason is because I feel like the `status` string and its ability to wait for `placement` to be ready could also handle CSS `animation` not just `transition`. Also maybe it works better than `isPositioned` if you want to make placement-aware animations using `framer-motion`/`react-spring`...

mihkeleidast:
I would want to say `useTransition`, i.e. just remove CSS from it, as what it does is help with the transition from closed to open state, and vice versa. How that is implemented afterwards, with CSS transitions, or animations, or some third CSS lib, is somewhat unimportant. But - `useTransition` is a core React hook now, so that name is sort of taken. `useFloatingTransition` reads clear to me, as it is also meant for use with `useFloating` hook, they are somewhat tied together, but still separate - but I think you don't like repeating that Floating prefix too much. That said, if I were to come across some conflicting hooks from multiple libs, that's what I would alias import it with. `useMotion` feels a bit wrong, as the hook itself does not offer any motion, by default it only delays the unmounting...

`useTransitionStyles` seems good though.

atomiks:
> But - useTransition is a core React hook now, so that name is sort of taken.

Yeah that's why I added `CSS` to it as I saw complaints about another library using that name.

---

It looks like in `framer-motion`, you need to use the `key` prop which changes based on `isPositioned`, so this hook isn't necessary.

Effectively, this renders the initial node which isn't positioned yet, then immediately unmounts and remounts the new one — since their sizes are the same, the `placement` measured from the initial node is the same as the remounted node, creating the desired result and it plays the animation from the final placement.

I guess if you're using `size()` then this may not work in certain scenarios though since it requires the dimensions before and after being positioned to remain the same.

```js
<motion.div key={isPositioned} />
```

> In React, changing a component's key makes React treat it as an entirely new component. So the old one is unmounted before the new one is mounted. So by changing the key of a single child of AnimatePresence, we can easily make components like slideshows.

mihkeleidast:
`useTransitionState` perhaps? And rename the returned value from `status` to `state` as well?

One other thing I started thinking about - is the returned `isMounted` variable just an alias for `status !== 'unmounted'`? Is it necessary at all as a separate variable in that case? Sort of confusing to say the same thing in two variables.

atomiks:
`useTransitionStatus` is great I think. And yeah we can omit `isMounted` then, only `useTransitionStyles` would need it.

mihkeleidast:
> And yeah we can omit `isMounted` then, only `useTransitionStyles` would need it.

Ahh I did not think about that. Would that mean that the useDelayUnmount hook is used in both of these hooks internally, resulting in two timeouts as well? I would keep it if that's the case, it seems better to reuse the exact same timeout and state variable.

atomiks:
Merging now, will release soon

Issue #2100, opened 2023-01-08
[React] Possible API enhancements


atomiks:
After `isPositioned` got added in v1.1.0 of the React DOM package, and `positionReference` in v0.15.0 of the main React package, plus some issues made here recently, I think some changes could help the API here.

This issue is created for feedback ideas, if you disagree with anything here feel free to comment.

---

## 1. Make `0` the default for `x` and `y` (not `null`)

🚨 Breaking (although rarely affects most usage). This is the only one that would actually justify bumping the package to v2, all the rest could be added in v1. So this could be done sometime in the future.

Removes the need for `?? 0`, making it simpler. 

`isPositioned` supplants `x == null` by default, but if the `open` option is specified and there are unmounting transitions, and you want to use `visibility` style for SSR ,`useEffect` + state flag works.

(Originally explored in the PR where `isPositioned` was added, but kept it non-breaking for now because I had additional ideas)

(cc: @lozinsky for Vue)

## 2. Prefix the callback refs with `set`

❓ Could either be a deprecation, or make the old names an alias — if deprecated, warnings could be added in the major version bump.

I originally avoided `set` because I felt like it encouraged doing this (creating an inline callback ref), which caused problems:

```jsx
// This:
<div ref={node => setFloating(node)} />

// Instead of:
<div ref={setFloating} />
```

It was also inspired by the Headless UI example. I also thought it was more declarative because placing it on the `ref` prop and `set` felt too imperative, plus is less characters to type out.

However, you can call `positionReference` with a virtual element in an effect or event handler, which ideally should be prefixed with `set`. It's also not clear they're callback refs, and that trying to use them as regular refs doesn't work which has tripped some up.

So the callback refs would become:

- `setFloating`
- `setReference`
- `setPositionReference`

But the old names could still exist as an alias `ref={reference}`/`ref={floating}` which feels cleaner? And keeps backwards-compat with barely any bundle size cost. Two ways of doing the same thing can be confusing though, so the docs might just switch to the `set*` ones entirely.

## 3. Allow synchronizing external elements in the hook

✅ Only an enhancement (I think? Not sure if overloaded types are an issue.)

This depends on your component API where one of the two techniques should be used — right you can use layout effects to synchronize external elements, but it's annoying:

```js
// Overloaded function that supports both techniques
const {x, y, setReference, setFloating} = useFloating(options);
const {x, y} = useFloating(referenceElement, floatingElement, options);

// Alternatively?
const {x, y} = useFloating({
[...]

PR #2101, opened 2023-01-08
feat(react): API enhancements

atomiks:
Implements 2, 3, and 4 of #2100 (all backwards-compatible)

### DIFF
--- a/packages/react-dom/src/useFloating.ts
+++ b/packages/react-dom/src/useFloating.ts
@@ -8,17 +8,39 @@ import type {
   UseFloatingData,
   UseFloatingProps,
   UseFloatingReturn,
+  VirtualElement,
 } from './types';
 import {deepEqual} from './utils/deepEqual';
 import {useLatestRef} from './utils/useLatestRef';
 
-export function useFloating<RT extends ReferenceType = ReferenceType>({
-  middleware = [],
-  placement = 'bottom',
-  strategy = 'absolute',
-  whileElementsMounted,
-  open,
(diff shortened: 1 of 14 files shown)


packages/react-dom/src/useFloating.ts
```diff
@@ -34,28 +38,44 @@ export function useFloating<RT extends ReferenceType = ReferenceType>({
...
-  const floating = React.useRef<HTMLElement | null>(null);
-  const cleanupRef = React.useRef<(() => void) | void | null>(null);
+  const referenceRef = React.useRef<RT | null>(null);
+  const floatingRef = React.useRef<HTMLElement | null>(null);
   const dataRef = React.useRef(data);
 
   const whileElementsMountedRef = useLatestRef(whileElementsMounted);
 
+  const [reference, _setReference] = React.useState<RT | null>(null);
+  const [floating, _setFloating] = React.useState<HTMLElement | null>(null);
```
Twixes, line 48: Hey @atomiks! Sadly this seems to have broken `@floating-ui/react`'s `useTransition` hooks, because `_setFloating` isn't being called. :(
And `useTransition` requires `floating` to be set for `status` to be changed from `'unmounted'`.
I believe this specifically is a problem when the floating element isn't mounted on first render, which is a common case with `isMounted`. When the floating element gets rendered at a later point, somehow the `floating` value isn't updated, so `status` always remains `'unmounted'`.

atomiks: @Twixes the `useTransition` hooks still work fine, do you have a CodeSandbox?

Issue #2094, opened 2023-01-06
[React] Define the root node for `FloatingPortal` instead of the portal subroot node


mihkeleidast:
Not sure what to call this, maybe some observation / feedback / question on a recent change.

From https://github.com/floating-ui/floating-ui/pull/1985:

> refactor(FloatingPortal): portal content by default no longer injects itself into a single root node with a default id of floating-ui-root. Instead the portals have a unique id and get removed from the DOM when the FloatingPortal’s host component unmounts — when specifying an id this logic can be bypassed however as it is assumed the node is static or controlled elsewhere. This better supports the new focus management system. (https://github.com/floating-ui/floating-ui/pull/1985)

Previously, the logic was that there was a single root node with id `floating-ui-root`. The portals themselves did not have subroots, i.e. they just forwarded children to the root.

```html
<body>
  <div id="floating-ui-root">
    <div class="my-component"></div>
    <div class="my-component"></div>
  </div>
</body>
```

When one specified a custom id to FloatingPortal, it made it possible to change the root node, but otherwise the structure would be the same i.e. the result would have been:

```html
<body>
  <div>
    <!-- custom root can be deep inside the rest of the app structure: https://github.com/floating-ui/floating-ui/blob/master/packages/react/test/visual/components/Drawer.tsx#L112 -->
    <div id="custom-root">
      <div class="my-component"></div>
      <div class="my-component"></div>
    </div>
  <div>
</body>
```

Now, it seems that there is no specific root node (it defaults to document.body), and the portals themselves are rendered into their own subroots, that are rendered into the root node.

```html
<body>
  <div id="subroot-1">
    <div class="my-component"></div>
  </div>
  <div id="subroot-2">
    <div class="my-component"></div>
  </div>
</body>
```

If I specify a custom ID to a component's FloatingPortal, it renders all of mu components to the same subroot, which results in a different HTML structure when viewed from the perspective of the subroot - there can be multiple focus guards in the same subroot in this case, which seems a bit weird? Also, there is no specific "root" in this case, really - it's whatever the parent node of the specified subroot is.

```html
<body>
  <div>
    <!-- custom root can be deep inside the rest of the app structure: https://github.com/floating-ui/floating-ui/blob/master/packages/react/test/visual/components/Drawer.tsx#L112 -->
    <div id="custom-root">
      <div class="my-component"></div>
      <div class="my-component"></div>
    </div>
  <div>
</body>
```

## Question / proposal

[...]

PR #2104, opened 2023-01-09
feat: change portal root node instead of subroot node

mihkeleidast:
Related to #2094 - example of what I'd change / expect to happen. Very raw changes, the tests probably fail as well, but hopefully it makes stuff more clear.

@atomiks 

This is a mix of the old (pre #1985) and the current behavior. 

Definitions:
- Root node - the container for multiple floating portals, usually specified on a floating component (i.e. Tooltip, Popover, Modal) level. What was previously `<div id="floating-ui-root">`, currently is `document.body`.
- Subroot node - the container for a single floating portal, appended to the root node. This was essentially a `<React.Fragment>` before, i.e. it did not exist. Right now it is the portal node created for each portal inside of `useFloatingPortalNode`.

Changes:
- This will bring back the default  `<div id="floating-ui-root">` container for all portal subroot nodes. Will clean up the DOM a bit, but would also be useful if the subroot node is disabled on a floating component.
- Specifying either `id` or `node` on a FloatingPortal will only change the root node. The subroot node is a functional requirement, as it has to potentially handle focus events and focus guards, and as such, it should not be customized / merged  / reused for multiple floating elements.
- For a Tooltip-like scenario that does not have to handle any focus elements, it is likely benefitial to provide a way to disable the subroot node. Not sure if this can be derived from some state automatically, or it needs to be a new prop on FloatingPortal. I added some commented out code for this, essentially, if the subroot is disabled, the portal should target the root directly.

### DIFF
--- a/packages/react/src/components/FloatingPortal.tsx
+++ b/packages/react/src/components/FloatingPortal.tsx
@@ -23,10 +23,13 @@ const PortalContext = React.createContext<null | {
   afterInsideRef: React.RefObject<HTMLSpanElement>;
   beforeOutsideRef: React.RefObject<HTMLSpanElement>;
   afterOutsideRef: React.RefObject<HTMLSpanElement>;
+  shouldRenderGuards: boolean;
 }>(null);
 
+const DEFAULT_ID = 'floating-ui-root';
+
 export const useFloatingPortalNode = ({
-  id,
+  id = DEFAULT_ID,
   enabled = true,
 }: {
   id?: string;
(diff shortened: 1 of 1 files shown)


atomiks:
Still confused on this, would be great to have a demo/tests set up that defines the expected behavior for each scenario

piecyk:
@mihkeleidast i think #2182 would kinda allow to have similar structure what you propose #2094

mihkeleidast:
@piecyk I think that PR has the same issues as this one, i.e. it is not immediately clear if/how changing the root node changes the focus handling. I think the current setup, in case the id is not customized, is set up so that the portals are nested in the parent portal if necessary, which probably catches some events. So the question is, how will using a different root node for some components affect this? Also, if you define a custom root for one component, but then render some other floating components inside of that one that do not have a custom root node defined, where should they be rendered? 

I haven't dropped this PR, I just have had to deal with some other work. Still planning to come back to this at some point and work on those questions and a comprehensive test suite. If you're willing to help out with this, it would be appreciated :)

piecyk:
> I think that PR has the same issues as this one, i.e. it is not immediately clear if/how changing the root node changes the focus handling. I think the current setup, in case the id is not customized, is set up so that the portals are nested in the parent portal if necessary, which probably catches some events. So the question is, how will using a different root node for some components affect this? 

It depends one structure, you can push everything under one element `<div id="floating-ui-root" />` but still focus would be handled as now with `usePortalContext` 
 
> Also, if you define a custom root for one component, but then render some other floating components inside of that one that do not have a custom root node defined, where should they be rendered?

It will be rendered like in current impl, inside the floating component, keeping correct order.

mihkeleidast:
> > Also, if you define a custom root for one component, but then render some other floating components inside of that one that do not have a custom root node defined, where should they be rendered?
> 
> It will be rendered like in current impl, inside the floating component, keeping correct order.

This can cause unexpected clipping, can it not? It is an assumption, but if it does, it needs to behave differently, as I wouldn't think clipping would be expected in most cases. Needs testing & validation at least.

tuures:
@mihkeleidast Being able to "sandbox" the app into under a single DOM node (vs `document.body`) would help us a lot. What kind of testing & validation you would be looking for? My suggestion would be to release the feature as "experimental" and gain some experiences using it in practice.

atomiks:
I think that `FloatingPortal` should have a `root` and a `container` prop, where `container` acts like `root` does right now, and `root` is  what `document.body` is now.

mihkeleidast:
> I guess there is that question of ergonomics though, which piecyk's context provider technique solves.
> 
> Assuming you're using the provider globally, everything gets put into a node at the end of `document.body`, so there shouldn't be any clipping.

Do you think the provider solution is better than how `root` worked previously, i.e. by just passing an ID of an element that should exist somewhere? My usecase was to portal a specific component to another root, which could not work with a context solution.

piecyk:
> > I guess there is that question of ergonomics though, which piecyk's context provider technique solves.
> > Assuming you're using the provider globally, everything gets put into a node at the end of `document.body`, so there shouldn't be any clipping.
> 
> Do you think the provider solution is better than how `root` worked previously, i.e. by just passing an ID of an element that should exist somewhere? My usecase was to portal a specific component to another root, which could not work with a context solution.

@mihkeleidast you still can pass ID, that logic is not changed. Basic context provider will make it easy to control the container root without custom wrapper.

mihkeleidast:
> @mihkeleidast you still can pass ID, that logic is not changed. Basic context provider will make it easy to control the container root without custom wrapper.

Generally floating-ui always requires custom component wrappers anyway, but gotcha. Maybe the provider should be more agnostic to potentially allow other options to be globally configured as well, not just the default portal root id? Not sure if there are such needs today, but would at least future-proof it a bit, I think.

piecyk:
> Maybe the provider should be more agnostic to potentially allow other options to be globally configured as well, not just the default portal root id? Not sure if there are such needs today, but would at least future-proof it a bit, I think.

Hmm hard to say, would keep it simple for now, what do you think @atomiks?

atomiks:
Starting simple seems good to me. If use cases emerge we can work from it.

piecyk:
> Starting simple seems good to me. If use cases emerge we can work from it.

Great, let me know if anything is missing here https://github.com/floating-ui/floating-ui/pull/2182

atomiks:
I think we're pretty close to merging https://github.com/floating-ui/floating-ui/pull/2182 instead, so closing this one

PR #2105, opened 2023-01-09
fix: account for closeOnFocusOut in FloatingPortal

mihkeleidast:
The focus guard in FloatingPortal still tries to close the floating element on focus out, even if `closeOnFocusOut={false}` is set on FloatingFocusManager. I think the context was meant for passing the values from FloatingFocusManager to FloatingPortal for scenarios like this, so reused it.

LMK if a test is necessary for this.


atomiks:
Tests are always great to prevent regressions, the more the better, especially since this lib has had too many already 😅

mihkeleidast:
Added a test based on the Drawer component example, quite simple really.

PR #2114, opened 2023-01-13
fix: improve useDismiss bubbles logic

### LINKED ISSUE #2112: bubbles=false does not as expected for different nested components
Reproduction: https://codesandbox.io/s/laughing-noyce-uoei3x?file=/src/App.tsx

I have two different floating components, `PopoverA` and `PopoverB`, that are nested like this:

```
<PopoverA>
  <PopoverB />
</PopoverA>
```

> With bubbling, the event is first captured and handled by the innermost element and then propagated to outer elements.

My understanding is that the dismissal bubbling should happen from inside to outside, i.e. when focus is inside `PopoverB`, and I press ESC, if `PopoverB`  has disabled bubbling, it should not bubble up to `PopoverA`, so `PopoverB` should close, while `PopoverA` stays open.

Instead, it seems that if I do not want `PopoverA` to close when `PopoverB` is closed with ESC, I need to disable the bubbling on `PopoverA`?

Something in this seems illogical to me, but then again, JS event bubbling (and capturing) is a complicated topic, so I might have something messed up in my head. 

I looked at the current tests, they seem to use the same `bubbles` configuration for both outer and inner dialog (haha, also saw I myself added those tests), so I think this sort of configuration is no-mans-land right now? What do you think, how should it work?

mihkeleidast:
Fixes #2112, I think.

Sort of forced my way through to this solution, but seems to work. Feel free to suggest any improvements.

- ~~`tree.nodesRef` => `tree.getNodesRef()` is necessary because the nodesRef was stale in the onKeyDown function that needs to look at the children bubbles values. Maybe this can be fixed in some other way, but seemed better to move it to a getter function that is always up to date, might fix some other stale data issues as well?~~
- `useDismiss` does not need to do anything with it's own `escapeKeyBubbles` value - I think it should only exist to be put into context, so parents can read it and decide whether to close or not.
- Not sure if `outsidePressBubbles` needs similar changes?

### DIFF
--- a/packages/react/src/hooks/useDismiss.ts
+++ b/packages/react/src/hooks/useDismiss.ts
@@ -102,7 +102,7 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
       return;
     }
 
-    dataRef.current.escapeKeyBubbles = escapeKeyBubbles;
+    dataRef.current.__escapeKeyBubbles = escapeKeyBubbles;
 
     function onKeyDown(event: KeyboardEvent) {
       if (event.key === 'Escape') {
@@ -113,7 +113,7 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
           children.forEach((child) => {
             if (
               child.context?.open &&
-              !child.context.dataRef.current.escapeKeyBubbles
+              !child.context.dataRef.current.__escapeKeyBubbles
             ) {
               shouldDismiss = false;
               return;
--- a/packages/react/src/hooks/useDismiss.ts
+++ b/packages/react/src/hooks/useDismiss.ts
@@ -70,6 +70,7 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
     events,
     nodeId,
     elements: {reference, domReference, floating},
+    dataRef,
   }: FloatingContext<RT>,
   {
     enabled = true,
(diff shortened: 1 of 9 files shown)


atomiks:
> tree.nodesRef => tree.getNodesRef() is necessary because the nodesRef was stale in the onKeyDown function that needs to look at the children bubbles values. Maybe this can be fixed in some other way, but seemed better to move it to a getter function that is always up to date, might fix some other stale data issues as well?

How can the mutable ref be stale, as it's not saved to a variable when it's accessed?

mihkeleidast:
> How can the mutable ref be stale, as it's not saved to a variable when it's accessed?

IDK, I could swear I did not have the children there when I started debugging this, but now it worked... maybe just the friday brain not working. Reverted those changes, sorry.

packages/react/src/hooks/useDismiss.ts
```diff
@@ -101,14 +102,27 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
...
     function onKeyDown(event: KeyboardEvent) {
       if (event.key === 'Escape') {
-        if (
-          !escapeKeyBubbles &&
-          tree &&
-          getChildren(tree.nodesRef.current, nodeId).length > 0
-        ) {
-          return;
+        if (tree && getChildren(tree.nodesRef.current, nodeId).length > 0) {
+          const children = getChildren(tree.nodesRef.current, nodeId);
```
atomiks, line 110:
```suggestion
        const children = tree ? getChildren(tree.nodesRef.current, nodeId) : [];
        if (children.length > 0) {
```

atomiks:
> Not sure if outsidePressBubbles needs similar changes?

Would be good to verify if it does need it

packages/react/src/hooks/useDismiss.ts
```diff
@@ -101,14 +102,27 @@ export const useDismiss = <RT extends ReferenceType = ReferenceType>(
       return;
     }
 
+    dataRef.current.escapeKeyBubbles = escapeKeyBubbles;
```
atomiks, line 105: I think this has no external purpose right? So can leave off the user-facing type (leave it as `any`) and prefix it like `__escapeKeyBubbles`

mihkeleidast: Yeah, I suppose it doesn't, so refactored. I suppose it can quite easily be renamed again to be public, if a usecase is found.

mihkeleidast:
> 



> > Not sure if outsidePressBubbles needs similar changes?
> 
> Would be good to verify if it does need it

Verified, added a similar test, which showed the need for a change there as well. I think the overall logic should be the same.

PR #2125, opened 2023-01-16
feat(useFloating): add enabled prop

mihkeleidast:
From https://github.com/floating-ui/floating-ui/issues/2100#issuecomment-1381495128.

Adds a "global" enabled prop on the main hook, which acts as the default value for the same option on all interaction hooks. This is helpful when all of the interaction hooks are disabled on the same condition, lessening the duplication.

Not sure if it needs deeper integration, i.e. if we can bail out of some of the compute work behind the scenes. Feel free to take over this one and make additional changes, you know the core better than me.

### DIFF
--- a/packages/react/src/hooks/useClick.ts
+++ b/packages/react/src/hooks/useClick.ts
@@ -25,9 +25,15 @@ export interface Props {
  * @see https://floating-ui.com/docs/useClick
  */
 export const useClick = <RT extends ReferenceType = ReferenceType>(
-  {open, onOpenChange, dataRef, refs}: FloatingContext<RT>,
   {
-    enabled = true,
+    enabled: enabledGlobal,
+    open,
+    onOpenChange,
+    dataRef,
+    refs,
+  }: FloatingContext<RT>,
+  {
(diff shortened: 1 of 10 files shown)


mihkeleidast:
@atomiks any thoughts on this? I can update the PR and resolve the conflicts, if this can move forward.

atomiks:
I'm not keen on this global value because it's not clear exactly what it means in the context of `useFloating`. I don't feel like individually configuring the enabled state of each hook is much of a problem despite the duplication. For example, how does the `enabled` state relate to the positioning, etc.

mihkeleidast:
I get that, I suppose one question is, what should happen when used like this?
```js
useFloating({ enabled: false, open: true })
```

Our Tooltip component does not render at all when disabled, even if controlled & set to open. But I don't think the hook can be that opinionated, some might want different behavior. So I'd say if disabled and open, the hook should still do all the work for positioning.

> I don't feel like individually configuring the enabled state of each hook is much of a problem despite the duplication.

True, not a huge issue, but it's easy to overlook / forget when adding another hook to a existing component.

PR #2130, opened 2023-01-19
fix(react): avoid adding `floating` element to interaction props dependency array

### LINKED ISSUE #2129: [React] `useTooltip` hook breaks after upgrading `@floating-ui/react` to `v0.17.0`
I wrote a custom `useTooltip` hook based on the [Tooltip examples](https://floating-ui.com/docs/tooltip) in the docs. It’s been working well for a few months.

When updating `@floating-ui/react` `v0.16.0 -> v0.17.0`, I noticed that my app was crashing whenever a tooltip opened. The app is entering an infinite render loop.

[I made a minimal reproduction in CodeSandbox here.](https://codesandbox.io/s/angry-cherry-eejht0?file=/src/useTooltip.js) If you upgrade the `@floating-ui/react` version from `v0.16.0 -> v0.17.0` then hover the preview frame, the app will crash with the error:

> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

I imagine that this is because of the [new refs APIs](https://github.com/floating-ui/floating-ui/pull/2101) mentioned in the [0.17.0 release notes.](https://github.com/floating-ui/floating-ui/releases/tag/%40floating-ui%2Freact%400.17.0) The changes are listed as “non breaking”, so I thought it would be worthwhile to provide a counterexample here.

atomiks:
Prevents `setState` loops when placing `getFloatingProps` as a dependency of a hook and the `ref` setter is inside. 

Fixes #2129

### DIFF
--- a/packages/react/src/hooks/useClick.ts
+++ b/packages/react/src/hooks/useClick.ts
@@ -25,7 +25,7 @@ export interface Props {
  * @see https://floating-ui.com/docs/useClick
  */
 export const useClick = <RT extends ReferenceType = ReferenceType>(
-  {open, onOpenChange, dataRef, refs}: FloatingContext<RT>,
+  {open, onOpenChange, dataRef, elements: {domReference}}: FloatingContext<RT>,
   {
     enabled = true,
     event: eventOption = 'click',
@@ -124,7 +124,7 @@ export const useClick = <RT extends ReferenceType = ReferenceType>(
             return;
           }
 
-          if (event.key === ' ' && !isSpaceIgnored(refs.domReference.current)) {
+          if (event.key === ' ' && !isSpaceIgnored(domReference)) {
             // Prevent scrolling
             event.preventDefault();
           }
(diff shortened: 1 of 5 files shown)


atomiks:
This fixes the `setState` loop but there's still some kind of rendering loop — looks like the `x` value is changing frequently causing a constant unmount-remount loop (just without an error or locking the browser up). Needs further investigation.

atomiks:
>  This fixes the setState loop but there's still some kind of rendering loop — looks like the x value is changing frequently causing a constant unmount-remount loop (just without an error or locking the browser up). Needs further investigation.

@rileyjshaw Resetting main to `v0.16.0` also has this problem... 🤔  locally, do the changes here fix it for you?

Maybe it was always broken in certain ways before... the API of this particular component doesn't seem to interact nicely with how the library operates

rileyjshaw:
Thanks @atomiks, and sorry for the delay! I got a concussion after posting this, so I’ve been avoiding screens as much as possible 🤕

I can confirm that the latest version fixes the issues I was seeing. And thanks for the additional tips! I really appreciate all the work you put into this library.

PR #2142, opened 2023-01-26
Export `composedOffsetParent()` as a convenience function

claviska:
Per @atomiks' comment [in this issue](https://github.com/shoelace-style/shoelace/issues/1135#issuecomment-1401017841), this PR restores the `composedOffsetParent()` function that [was previously removed](https://github.com/floating-ui/floating-ui/pull/1894/files) as an opt in convenience function for those using Floating UI in shadow roots.

The proposed usage is:

```ts
import { composedOffsetParent, computePosition, platform } from '@floating-ui/dom';

computePosition(anchor, popup, {
  // ...
  platform: {
    ...platform,
    getOffsetParent: composedOffsetParent
});
```

I suspect you'll see a lot of issues from the web component community like [this one](https://github.com/shoelace-style/shoelace/issues/1135) and [this one](https://github.com/Esri/calcite-components/issues/6300), so it couldn't hurt to add usage and a short explainer to the docs. I wasn't sure where you wanted that — [perhaps as a tip/note under this section?](https://floating-ui.com/docs/platform#getoffsetparent)

Also, I noticed three failing tests locally, but they don't seem to be related.

Let me know what you think!

### DIFF
--- a/packages/dom/src/utils/getOffsetParent.ts
+++ b/packages/dom/src/utils/getOffsetParent.ts
@@ -5,6 +5,7 @@ import {
   isContainingBlock,
   isHTMLElement,
   isLastTraversableNode,
+  isShadowRoot,
   isTableElement,
 } from './is';
 import {getWindow} from './window';
@@ -60,4 +61,52 @@ export function getOffsetParent(element: Element): Element | Window {
   }
 
   return offsetParent || getContainingBlock(element) || window;
+}
+
+/**
+ * Polyfills the old offsetParent behavior from before the spec was changed:
+ * https://github.com/w3c/csswg-drafts/issues/159
+ *
+ * This is exported as a convenience function for consumers that want to opt
+ * into the more accurate old behavior. Note that using this function in some
+ * DOM configurations may introduce performance issues.
+ */
+export function composedOffsetParent(element: HTMLElement): Element | null {
+  let {offsetParent} = element;
+
+  let ancestor: Element = element;
+  let foundInsideSlot = false;
+
+  while (ancestor && ancestor !== offsetParent) {
+    const {assignedSlot} = ancestor;
+
+    if (assignedSlot) {
+      let newOffsetParent = assignedSlot.offsetParent;
+
+      if (getComputedStyle(assignedSlot).display === 'contents') {
+        const hadStyleAttribute = assignedSlot.hasAttribute('style');
+        const oldDisplay = assignedSlot.style.display;
+        assignedSlot.style.display = getComputedStyle(ancestor).display;
+
+        newOffsetParent = assignedSlot.offsetParent;
+
+        assignedSlot.style.display = oldDisplay;
+        if (!hadStyleAttribute) {
+          assignedSlot.removeAttribute('style');
+        }
+      }
+
+      ancestor = assignedSlot;
+      if (offsetParent !== newOffsetParent) {
+        offsetParent = newOffsetParent;
+        foundInsideSlot = true;
+      }
+    } else if (isShadowRoot(ancestor) && ancestor.host && foundInsideSlot) {
+      break;
+    }
+    ancestor = ((isShadowRoot(ancestor) && ancestor.host) ||
+      ancestor.parentNode) as Element;
+  }
+
+  return offsetParent;
 }


atomiks:
I'm thinking the user should polyfill this externally themselves, what do you think?

```js
const originalOffsetParent = 
  Object.getOwnPropertyDescriptor(HTMLElement.prototype, 'offsetParent').get;

function composedOffsetParent(element: HTMLElement): Element | null {
  // ...
}

Object.defineProperty(HTMLElement.prototype, 'offsetParent', {
  get() {
    // If there's perf problems, opt out of it and change the layout if it breaks
    return this.hasAttribute('data-disable-offset-parent-polyfill')
      ? originalOffsetParent.apply(this) 
      : composedOffsetParent(this);
  }
});
```

claviska:
I can appreciate not wanting to include this to keep the package small. I'm OK with polyfilling it myself, but I think this problem will be extremely common in any custom element library. Whether Floating UI wants to provide an easy solution or not is the real question.

You've already exposed `platform` so we _can_ make this work on our own. And yes, polyfilling `HTMLElement` will also work, but it will cause everything else using it to not work according to spec. A modular solution that can be applied to just this use case is much more appropriate than overloading.

In [this comment](https://github.com/shoelace-style/shoelace/issues/1135#issuecomment-1401017841), you mentioned:

> To reduce the browser-based bundles' sizes (where tree-shaking can't be utilized), it could be a separate import path, but that'd complicate things a lot.

To understand better, is your reluctance based on package size, maintaining the `composedOffsetParent()` function, or a bit of both? If it complicates things too much to include this in a separate file, maybe the function itself should be a separate package. That way, it's opt in only and we can still add a section to the docs showing folks how to get that behavior through `platform.getOffsetParent`, along with the perf caveat of leaning on it too much.

My goal is to solve my own problem, but I would also like to make it easier for other devs to solve this. We're already seeing folks run into [the same problem](https://github.com/Esri/calcite-components/issues/6300), so it's probably not going away without a documented solution. And overloading `HTMLElement.offsetParent` is a massive overreach.

Personally, I think the value in utility libs is normalizing platform gaps like this. My preferred solution would be a higher level option:

```ts
computePosition(anchor, popup, {
  // ...
  useComposedOffsetParent: true // name TBD
});
```

But again, I'm OK with patching `platform.getOffsetParent`. It just sucks that we have to:

1. Copy the `composedOffsetParent()` code from a delete commit
2. Copy `isShadowRoot` because it's not exported
3. Copy `getWindow` and `isWindow` because they're not exported

…when all this code is/was already available through the lib.

atomiks:
I agree that an option like that is easy for a consumer, but is pretty bad from this library's perspective trying to stay pure/lean in API surface.

I think a separate npm package is best, especially if the polyfill can be optimized in different versions. I hope someone in the web component community could publish such a package:

`npm install composed-offset-parent`

```js
export function composedOffsetParent(element) {
  // ...
}
```

As to how this library could consume the polyfill, the function needs to be fed into `getTrueOffsetParent`, so the usage is a bit different:

```js
import {computePosition, platform, getOffsetParent} from '@floating-ui/dom';
import {composedOffsetParent} from 'composed-offset-parent';

computePosition(..., {
  platform: {
    ...platform,
    getOffsetParent: (element) => getOffsetParent(element, composedOffsetParent),
  }
});
```

Would you be willing to publish such a polyfill lib and edit the PR to change how the polyfill is passed? If it's undefined, use `element.offsetParent`, otherwise use the function.

claviska:
> Would you be willing to publish such a polyfill lib and edit the PR to change how the polyfill is passed? If it's undefined, use element.offsetParent, otherwise use the function.

I didn't write the code, so I'd rather not.

> As to how this library could consume the polyfill, the function needs to be fed into getTrueOffsetParent, so the usage is a bit different:
>
>
> ```ts
> import {computePosition, platform, getOffsetParent} from '@floating-ui/dom';
> import {composedOffsetParent} from 'composed-offset-parent';
>
> computePosition(..., {
>   platform: {
>     ...platform,
>     getOffsetParent: (element) => getOffsetParent(element, composedOffsetParent),
>   }
> });
> ```

Thanks for clarifying. To see how difficult this would for a consumer, I tried copying `composedOffsetParent()` into my own module. Then I had to pull in `isShadowRoot()` and `getWindow()`. Then I tried importing `getOffsetParent()` from Floating UI, but that doesn't appear to be exported in 1.1.0. So I copied that in too.

But then the following functions were also missing:

- `getTrueOffsetParent()`
- `isTableElement()`
- `getNodeName()`
- `isContainingBlock()`

And if I copied those over, who knows what else would be missing missing. Sure, you could add these as exports (and perhaps some already are), but I suspect you probably don't want to add all these to your public API, either.

At this point, I stopped because it's already too much effort and too much code that will be duplicated doing it this way. I doubt other devs will want to do this either.

> I agree that an option like that is easy for a consumer, but is pretty bad from this library's perspective trying to stay pure/lean in API surface.

I fully appreciate this. Unfortunately, at this point, I'd rather leave things broken than 1) overload a platform API or 2) copy a significant number of Floating UI's utility functions into my own code.

I'll close this since it's clear you don't want to incorporate it into the project. Hopefully the fixed positioning strategy will suffice for folks who are affected by this.

atomiks:
> Then I tried importing getOffsetParent() from Floating UI, but that doesn't appear to be exported in 1.1.0. So I copied that in too.

Oops sorry my example was wrong. It's on `platform`, so you'd just need to do `platform.getOffsetParent(element, composedOffsetParent)`.  I think that makes an external polyfill much simpler - you only need `getWindow` and `isShadowRoot`? And if you don't care for internal iframes, you don't need `getWindow`.

I do want to provide a built-in solution to fix this problem, but I'd rather the polyfill part be a separate library that can be iterated on and used by others because this problem doesn't just apply to this library, it affects all usage of `offsetParent`.

claviska:
After thinking about it more, no, it doesn't change anything because it offloads basic positioning logic to third parties and/or end users for features we expect to "Just Work."

I understand this is an unfortunate gap in the platform. I wouldn't have pushed so hard if it were some kind of custom positioning logic being asked for, but I'm still of the opinion that libraries exist to make things easier and solve these types of problems. If the maintainers don't want to support it, that's their decision and I'll respect that.

For now, I'm advising my users to apply the fixed positioning strategy when they stumble upon this issue.

atomiks:
> If the maintainers don't want to support it, that's their decision and I'll respect that.

But I do want to add support for it, I just don't want to include the polyfill itself as part of the library. As I said, it affects all usage of `offsetParent` whether they're using this library or not, polyfills are typically their own package (either pure or mutative).

Regardless though, isn't the end DX of using `fixed` also the same as the polyfill? Since the polyfill introduces a perf issue, it can't be defaulted to, so the user has to add a special attribute to opt-in to the polyfill every time there's a positioning problem and then decide if they "tolerate" the performance impact it might have for the single case.

So actually, defaulting to `fixed` for shadow DOM based components seems like it would not only almost always avoid the positioning problem, it would also prevent the user from having to think about opting-in to any polyfills and then deciding if they tolerate its perf impact. `fixed` and `absolute` work largely the same, but with some minor differences, such as with elastic overscroll.

claviska:
> Regardless though, isn't the end DX of using fixed also the same as the polyfill?

Not in all cases. The `fixed` strategy usually fails when the [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block) is anything but the viewport. So any ancestor with a translate, filter, backdrop filter, etc. won't work with it.

Note the erroneous tooltip position. The `<div>` that surrounds the tooltip simply has `translate: 1px`.

CleanShot 2023-02-02 at 12 04 18@2x

[View on CodePen](https://codepen.io/claviska/pen/QWBzgzP?editors=1000)

Positioning would still be correct with the `absolute` strategy (but, of course, it won't break out of the overflow in this example).

> Since the polyfill introduces a perf issue, it can't be defaulted to, so the user has to add a special attribute to opt-in to the polyfill every time there's a positioning problem and then decide if they "tolerate" the performance impact it might have for the single case.

Correct. And this is unfortunate, but the users that reported issues had menus and lists with a lot of children. I was considering letting them opt in through my abstraction. Maybe something like this:

```html
<sl-popup strategy="absolute"></sl-popup>
<sl-popup strategy="absolute:shadow"></sl-popup> <!-- uses the fix -->
<sl-popup strategy="fixed"></sl-popup>
```

I don't love it, but without it some users won't be able to position things correctly. I'm hoping this number remains small, but time will tell.

To be clear, I'm not upset and I somewhat agree with your decision to _not_ include it in Floating UI. But since `fixed` isn't a silver bullet and because there's no easy way to handle this edge case without maintaining my own positioning logic, it would be nice if the positioning library facilitated that either through an opt in API or even a separate package.

But no worries! I still love Floating UI and I appreciate the attention and support you've provided. ✌️

atomiks:
Ok thanks for providing cases where `fixed` doesn't work. I think the main disagreement here is not about providing a built-in to fix this, it's about how to deal with the polyfill.

I think for a library, you're 100% right — you shouldn't mutate the HTMLElement prototype. For an app, that probably doesn't matter, although maybe it could lead to a smooshgate problem somehow. So I agree that it should be a "ponyfill" (pure polyfill) from Shoelace's perspective. 

Where I disagree is that this should be part of Floating UI as a separate import path or a main export. That leads to a bunch of complexity with exports and would be much simpler as its own package that is consumed by Floating UI.

I know you said you didn't write the polyfill, but neither did I, and I kind of don't want to maintain a package like that. But clearly it seems invaluable for the web component community.

The benefits of a separate package external to Floating UI include:

- It can be iterated on and versioned independently of Floating UI.
- It affects all usage of `offsetParent`, so it would have a shared source if you need to use `offsetParent` for any other reason, and reduce duplication or conflicting polyfills.
- Can influence the need for a spec to fill this platform gap.

This is easy enough to be consumed by Floating UI once added:

```js
import {computePosition, platform} from '@floating-ui/dom';
import {composedOffsetParent} from 'composed-offset-parent';

function getComposedOffsetParent(element) {
  return platform.getOffsetParent(element, composedOffsetParent);
}

computePosition(a, b, {
  platform: {
    ...platform,
    getOffsetParent: strategy === 'absolute:shadow' 
     ? getComposedOffsetParent 
     : platform.getOffsetParent,
  }
});
```

If you don't want to make it a package, the external polyfill is easy enough to write internally in Shoelance, the only duplication you need is the `isShadowRoot` and related functions (because it's on `platform` as I said, it's easier).

atomiks:
Opened a PR here: https://github.com/floating-ui/floating-ui/pull/2160

The polyfill is very small, whether someone wants to make it a package or not is up to them, but this is what I'm willing to do.

PR #2149, opened 2023-01-28
fix(safePolygon): use `pointer-events` blocking

### LINKED ISSUE #2124: [React] Hover events possible for 1 tick after `safePolygon` element is injected to body
Before `v0.15.0`, a "global" holes strategy with `pointer-events: none` on the `<body>` and `pointer-events: auto` on elements was used for `blockPointerEvents: true`, but this caused unwanted side-effects in various scenarios.

Since then, it now injects a real `<polygon>` element to the body on `mouseleave` to block pointer events, but it seems to leave hover events possible for 1 tick before the browser will consider it blocking other content.  This means hover events e.g. `onMouseEnter` can now fire while trying to traverse the polygon.

A workaround is to ensure elements underneath that can be affected have a short open delay.

I can't really think of a good solution to this as the polygon needs to be generated only on mouseleave, but it seems like it needs to be generated just before? I don't want to return to the previous solution either due to the side-effects.

atomiks:
Fixes #2124

This reverts the polygon technique and goes back to the old `pointer-events` blocking technique. This does have side-effects, so I'm going to prominently display a warning about how to disable it and how to work around these side-effects. 

This PR adds an attribute to the `<body>` called `data-floating-ui-safe-polygon` which allows the user to target it in their CSS.

Both these two issues for scrolling containers (not the main `window`):

- Not being able to scroll while the cursor rests over the reference element while the floating element is open
- macOS' auto-hiding scrollbars forcefully appearing in Chrome

Can be solved like this:

```js
function ScrollContainer({children}) {
  return (
    <div data-scroll>
      <div>
        {children}
      </div>
    </div>
  );
}
```

```css
[data-floating-ui-safe-polygon] [data-scroll] {
  pointer-events: auto;
}

[data-floating-ui-safe-polygon] [data-scroll] > div {
  pointer-events: none;
}
```

Obviously, this requires setup, ensuring the user applies this attribute and the wrapping div to all relevant scrolling containers, and as a result isn't automagic, but I really cannot think of any other good solution to this problem. At least they can disable all blocking if necessary.

@mihkeleidast maybe any other ideas?

### DIFF
--- a/packages/react/src/hooks/useHover.ts
+++ b/packages/react/src/hooks/useHover.ts
@@ -235,6 +235,7 @@ export const useHover = <RT extends ReferenceType = ReferenceType>(
           x: event.clientX,
           y: event.clientY,
           onClose() {
+            clearPointerEvents();
             cleanupMouseMoveHandler();
             closeWithDelay();
           },
@@ -298,6 +299,7 @@ export const useHover = <RT extends ReferenceType = ReferenceType>(
     move,
     closeWithDelay,
     cleanupMouseMoveHandler,
+    clearPointerEvents,
     onOpenChange,
     open,
     tree,
--- a/packages/react/src/hooks/useHover.ts
+++ b/packages/react/src/hooks/useHover.ts
@@ -1,8 +1,10 @@
 import * as React from 'react';
 import useLayoutEffect from 'use-isomorphic-layout-effect';
 
-import {useFloatingTree} from '../components/FloatingTree';
-import {destroyPolygon} from '../safePolygon';
+import {
+  useFloatingParentNodeId,
+  useFloatingTree,
+} from '../components/FloatingTree';
 import type {
   ElementProps,
   FloatingContext,
@@ -13,15 +15,19 @@ import {getDocument} from '../utils/getDocument';
 import {isElement, isMouseLikePointerType} from '../utils/is';
 import {useLatestRef} from './utils/useLatestRef';
 
+const safePolygonIdentifier = 'data-floating-ui-safe-polygon';
+
 export interface HandleCloseFn<RT extends ReferenceType = ReferenceType> {
   (
     context: FloatingContext<RT> & {
       onClose: () => void;
       tree?: FloatingTreeType<RT> | null;
       leave?: boolean;
-      polygonRef: React.MutableRefObject<SVGElement | null>;
     }
   ): (event: MouseEvent) => void;
+  __options: {
+    blockPointerEvents: boolean;
+  };
 }
 
 export function getDelay(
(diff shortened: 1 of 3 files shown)


mihkeleidast:
Not sure I have any alternative ideas - you likely have done your research, so I don't suppose there are any other magic solutions in other libraries that do the same but better / without side-effects?

As for bringing the previous behavior back, one thing to consider: perhaps `blockPointerEvents` should default to `false` instead of the current `true`? My thinking on this is that since `true` has side effects, while `false` does not, opting into a special behavior with side effects that need to be taken care of would make more sense than always having to opt out of it, if one does not care about this.

I currently have two components that use useHover+safePolygon, one has pointer events blocking disabled, the other has not, but I'm not sure if it is intentional or not. I can test on monday to see if my other usecase is correctly not disabling it.

Also, note for the docs: with the previous behavior that is being restored here, we ran into a bunch of issues in tests (testing-library + user-event), for example when a component that uses safePolygon with `blockPointerEvents: true` is open, it is not possible to simply click on some other element in tests, you have to specify that the click should not care about pointert-events. This is certainly an annoyance for consumers of a ui library that uses floating-ui who don't know the specifics of why/how this behavior is implemented. I think that stuff like this should probably also be documented, if it can't be automatically worked around.

atomiks:
> solutions in other libraries that do the same but better / without side-effects?

I think they manually block the `pointer-events` of surrounding items, e.g. Menu component. I haven't seen this automatic blocking elsewhere.

> As for bringing the previous behavior back, one thing to consider: perhaps blockPointerEvents should default to false instead of the current true?

Yeah, I agree with this. I think you can use a condition like `safePolygon({ blockPointerEvents: isRealBrowser })` or such.

mihkeleidast:
> I can test on monday to see if my other usecase is correctly not disabling it.

Just as a follow-up, it is necessary for the other case, meaning I can't just disable it for all my usecases.

I wonder if we could offer options for this, however, each with its own drawbacks?

1. the now-implemented method of global pointer-events disabling
2. some more static [svg path exit area](https://www.smashingmagazine.com/2021/05/frustrating-design-patterns-mega-dropdown-hover-menus/#svg-path-exit-areas) logic. I think the staticness of it would also work around the 1 tick issue? This would have the benefit of being local, but the downside of sometimes being less obvious perhaps?

atomiks:
The main problem with the static path like that is when you have two items with submenus attached directly next to each other, it ends up blocking the other item undesirably for a portion of the element (hover events can't fire). Whereas the current dynamic polygon only blocks for ~1 pixel when moving up/down, although it can still be better.

PR #2153, opened 2023-01-31
Apply computed position only for current refs

dmitriyshx:


atomiks:
What bug does this fix/solve?

dmitriyshx:
Hello. When changed ref, the position is calculated incorrectly
Example bug: https://codesandbox.io/s/white-leaf-xptlyv?file=/src/App.tsx

Current position: 0,0 (x, y)
Correct position: 100, 100

atomiks:
You should be using state to store the external elements, not refs. React does not permit reading refs during render like in your demo.

Using state, it works correctly: https://codesandbox.io/s/wizardly-murdock-42j88s?file=/src/App.tsx

The docs mention this:
Screenshot 2023-01-31 at 10 50 51 pm

dmitriyshx:
Thanks, it works

Issue #2128, opened 2023-01-16
Focus management and prop getters for external reference


mihkeleidast:
Currently, the only way to get the correct props (both aria and interaction) and focus management is to use an "internal" reference element - i.e, the reference element must be in some context through which the floating hook return is shared. This works for lots of smaller floating components, where the reference and floating element can be co-located this way. Unfortunately, it does not work for some more complex floating elements, like modals and drawers, where there can be multiple references for a single floating element.

I put together a very simple example based on the drawer test: https://github.com/floating-ui/floating-ui/pull/2127, which should showcase the issues. If a more complex example is necessary, I can put together something where in a virtualized list (we use [react-virtuoso](https://github.com/petyosi/react-virtuoso) at the moment), each list item has a trigger that opens a drawer. The list items being rendered virtually would make even the thought of co-location of the drawer/reference impossible, as we can't have the drawer self-closing when user scrolls the list.

What I'd like to achieve, is:
- have an external reference on a floating component;
- have that external reference receive all the same props and behavior that an internal reference would get.

I guess it may be possible to do this in userland already, by syncing the hook return value to some more global state management system based on an unique ID, and then retrieving the reference props based on the same ID? I haven't gotten that far yet, wanted to file an issue anyway to get an opinion on this (should/could it be simpler?) and eventually provide some documented example for others who run into similar issues/questions.

This also sort of ties into the singleton issue (https://github.com/floating-ui/floating-ui/issues/1697), but as that is mostly described as an issue of transitions, while this is about functionality, I filed it separately. But feel free to merge these if they likely share the solution.

atomiks:
I guess as you said, you'd need to use an effect to synchronize the values with the parent context/state, like with external elements. Or lift the hook up higher in the component tree where the values are needed, then pass its values down to the child component.

mihkeleidast:
> Or lift the hook up higher in the component tree where the values are needed, then pass its values down to the child component.

That can be problematic if the values are passed through a simple context like is used in all of the examples, as it's rather easy to run into a conflicting scenario where you want multiple hooks/context values at a similar level, and with context, the innermost provider always wins.

But I'll see what I can do about implementing some sort of global store, and will report back in case of any issues. Or, if everything works wonderfully, we can document this.

mihkeleidast:
[...]

PR #2154, opened 2023-01-31
fix: focus management with tabbable elements in between reference and floating element

mihkeleidast:
Relates to #2128. This should fix cases where the hook/state is "lifted higher in the component tree", when the reference element and floating element are not directly next to each other in the DOM. Also fixes the external reference usecase, i.e. focus management with external refs should work a lot better.

Crossing my fingers that the existing tests covered enough of the previous logic, so the changes won't break anything. 

Maybe some additional code can be cleaned up after this as well, hints are welcome. I did not want to make too much unnecessary changes unless confirmed.

### DIFF
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -20,6 +20,7 @@ import {
   getNextTabbable,
   getPreviousTabbable,
   getTabbableOptions,
+  isNextTabbableFrom,
   isOutsideEvent,
 } from '../utils/tabbable';
 import {usePortalContext} from './FloatingPortal';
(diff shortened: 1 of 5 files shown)


atomiks:
I'm going to have to check how this works with VoiceOver virtual cursor on iOS/Safari, because I have a feeling it may not work (and it needs to)

atomiks:
So with iOS VoiceOver (and possibly other touch-based screen readers) there are two problems with the existing Popover demo:

1. It doesn't focus the first tabbable element (`initialFocus` no longer works, also for `modal: true` as well)
2. Then trying to move focus from the reference element (swipe right), the virtual cursor doesn't move into the floating element properly, and if it did also won't exit to the correct element.

They have no concept of `tab` key, so it can't be relied upon to move focus

mihkeleidast:
Huh. How is the current code is detecting the virtual cursor movement? I would be helpful to understand how the events and focus movement is working at the moment for VoiceOver, then it could maybe give some ideas on how to fix the issues.

I would assume (but maybe incorrectly?) this not specifically a touch-based screen reader issue, virtual cursors exist in desktop screen readers as well. As I don't have any Apple devices to test on, any way I can reproduce this on my Windows + NVDA setup?

Right now I tested (with NVDA) that when the Popover is open and my virtual cursor is on the button, then the next element is the dialog in both master and this branch, so no change in behavior.

For the initialFocus issues as well, I don't think I changed anything relating to that part of the code, so not sure how the behavior could have changed 🤔 

Or, any other ideas on how to make the test I added work properly? One thing I came up with was that it would be useful if we had a focus guard right after the reference element, but that would likely mean another wrapper component that has to be used around it?

mihkeleidast:
Alright, that makes sense, thank you 👍 I'll try to add some semi-mocked (i.e. programmatically focusing the outside guards) tests about this, so it would be evident that this needs to work in a specific way.

But as you said, the outside guards need to be rendered somewhere after the reference element. Currently they are around the portal, which breaks when there are any tabbable elements between the reference and portal in the component tree. The only alternative I can see right now is adding a  `<FloatingReference>` component that should be wrapped around the reference, which would then render both before and after guards immediately after its children. Do you think this would be a viable solution, or are there other more magical solutions to consider?

mihkeleidast:
> What's the main use case for not being able to immediately tab into the floating element after the reference btw?

External refs - right now floating-ui expects that the reference is rendered right before the portal/focus manager, and nothing can be in between. Realistically though that is very hard to enforce when providing a component that allows either an internal or external reference - they can be far away from each other in an app.

Example: we have pages where there are tables, each row has a button that opens a "detail view". That detail view is a floating element that is shared between all the rows (there can only be one open at a time, plus the table is virtualized so it would be very hard to keep the detail view open when the user scrolls down and the row unrenders). There can be lots of rows in that table, so if a user opens the detail view from the first row, currently they'd have to tab through the entire table before arriving at the floating element, which feels like bad UX to me.
```jsx
<>
  <table>
    <tr><td><button>trigger</button></td></tr>
  </table>
  <drawer>
    some dynamic content
  </drawer>
</>
```

Also - the `order` prop on FloatingFocusManager does not make sense if elements in between are not allowed, specifically when it is used with `order={['reference', 'content']}`. The docs for that tell that the focus should be moved from reference to floating, but this is not actually the case when there are other tabbable elements in between.

atomiks:
Yeah I see, and I also see how that relates to singletons as well. I think modal focus management, instead of non-modal, is the best solution to make it work properly for now. But it would be good to somehow make it work in all cases

mihkeleidast:
Yeah, using modal focus management is a somewhat usable workaround for the moment, though not ideal.

I'll close this, as this solution in itself is not viable, but I have the linked issue still open, so we can track the problem and potential solutions there.

PR #2158, opened 2023-02-01
fixed available width bug

### LINKED ISSUE #1897: [Core] `size` middleware gives wrong `availableWidth`
availableHeight is right, but availableWidth is wrong
https://codesandbox.io/s/floating-ui-dom-template-forked-k2y1nu?file=/src/index.js

porov:
Fixes #1897

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ stats
 .parcel-cache
 test-results
 out
+BOBO_DEBUG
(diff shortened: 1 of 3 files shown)


atomiks:
I think this should be an opt-in option, as it doesn't seem suitable for aligned placements, and requires `shift()` to be somewhere in the middleware array to work as expected - but I'm not sure what to call the option.

PR #2160, opened 2023-02-03
feat(dom): add ability to polyfill `offsetParent` access

atomiks:
This enables the ability to add a `composedOffsetParent` polyfill, which can also be used conditionally when perf problems arise.

## Polyfill

```js
function getWindow(node) {
  return node.ownerDocument?.defaultView || window;
}

function isShadowRoot(node) {
  return node instanceof getWindow(node).ShadowRoot;
}

/**
 * Polyfills the old offsetParent behavior from before the spec was changed:
 * https://github.com/w3c/csswg-drafts/issues/159
 */
function composedOffsetParent(element) {
  let {offsetParent} = element;
  let ancestor = element;
  let foundInsideSlot = false;

  while (ancestor && ancestor !== offsetParent) {
    const {assignedSlot} = ancestor;

    if (assignedSlot) {
      let newOffsetParent = assignedSlot.offsetParent;

      if (getComputedStyle(assignedSlot).display === 'contents') {
        const hadStyleAttribute = assignedSlot.hasAttribute('style');
        const oldDisplay = assignedSlot.style.display;
        assignedSlot.style.display = getComputedStyle(ancestor).display;

        newOffsetParent = assignedSlot.offsetParent;

        assignedSlot.style.display = oldDisplay;
        if (!hadStyleAttribute) {
          assignedSlot.removeAttribute('style');
        }
      }

      ancestor = assignedSlot;
      if (offsetParent !== newOffsetParent) {
        offsetParent = newOffsetParent;
        foundInsideSlot = true;
      }
    } else if (isShadowRoot(ancestor) && ancestor.host && foundInsideSlot) {
      break;
    }
    ancestor = ((isShadowRoot(ancestor) && ancestor.host) ||
      ancestor.parentNode);
  }

  return offsetParent;
}
```

## Usage

```js
computePosition(a, b, {
  platform: {
    ...platform,
    getOffsetParent: element => platform.getOffsetParent(element, composedOffsetParent)
  }
});
```

### DIFF
--- a/packages/dom/src/types.ts
+++ b/packages/dom/src/types.ts
@@ -44,7 +44,7 @@ export interface Platform {
   }) => Promisable<Rect>;
   getOffsetParent?: (
     element: Element,
-    polyfill?: (element: Element) => Element | null
+    polyfill?: (element: HTMLElement) => Element | null
   ) => Promisable<Element | Window>;
   isElement?: (value: unknown) => Promisable<boolean>;
   getDocumentElement?: (element: Element) => Promisable<HTMLElement>;
--- a/packages/dom/src/types.ts
+++ b/packages/dom/src/types.ts
@@ -42,7 +42,10 @@ export interface Platform {
     offsetParent: Element;
     strategy: Strategy;
   }) => Promisable<Rect>;
-  getOffsetParent?: (element: Element) => Promisable<Element | Window>;
+  getOffsetParent?: (
+    element: Element,
+    polyfill?: (element: Element) => Element | null
+  ) => Promisable<Element | Window>;
   isElement?: (value: unknown) => Promisable<boolean>;
   getDocumentElement?: (element: Element) => Promisable<HTMLElement>;
   getClientRects?: (element: Element) => Promisable<Array<ClientRectObject>>;
--- a/packages/dom/src/utils/getOffsetParent.ts
+++ b/packages/dom/src/utils/getOffsetParent.ts
@@ -9,12 +9,21 @@ import {
 } from './is';
 import {getWindow} from './window';
 
-function getTrueOffsetParent(element: Element): Element | null {
+type Polyfill = (element: Element) => Element | null;
+
+function getTrueOffsetParent(
+  element: Element,
+  polyfill?: Polyfill
+): Element | null {
   if (
     !isHTMLElement(element) ||
     getComputedStyle(element).position === 'fixed'
   ) {
     return null;
+  }
+
+  if (polyfill) {
+    return polyfill(element);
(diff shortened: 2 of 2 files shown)


zengguirong:
@atomiks I can't understand your polyfill.
I write another simple version, it works fine with shadow dom in my project.
I have performance issues with your polyfill, but so far this one doesn't
```typescript
function composedOffsetParent(element: HTMLElement) {
  let ancestor: Node | null = element
  while (ancestor) {
    if (ancestor === document.documentElement || ancestor === document.body) {
      return null
    } else if (ancestor instanceof Element) {
      const style = getComputedStyle(ancestor)
      if (style.display === 'none' || style.position === 'fixed') {
        return null
      } else if (ancestor !== element && style.position !== 'static') {
        return ancestor
      }
    }

    ancestor = getParentNode(ancestor)
  }

  return null
}

function getParentNode(current: Node) {
  return current instanceof ShadowRoot ? current.host : (current as Element).assignedSlot || current.parentNode
}
```

atomiks:
I don't understand it either as I didn't really write it. I think that one may fail in some cases with `display: contents;` ancestors, but if it works for you then that's good. To handle iframes it needs to use the `getWindow` function I provided above. Consider publishing it as a package for others to use 🙏

jcfranco:
@zengguirong @atomiks Unfortunately, I confirmed that the above polyfill doesn't cover all of our use cases (namely [this one](https://github.com/Esri/calcite-components/issues/3667)). 

### TL;DR: I published https://www.npmjs.com/package/composed-offset-position with ponyfills that can be passed into `platform.getOffsetParent(element, composedOffsetParent)`

On the bright side, I did spend some time today testing  [`offsetParentPolyfill`](https://github.com/josepharhar/offsetparent-polyfills/) (from https://bugs.chromium.org/p/chromium/issues/detail?id=1331803#c9 and also mentioned in https://github.com/shoelace-style/shoelace/issues/1135) and it's working great (thanks @josepharhar!). I know there were [concerns about the polyfill patching HTMLElement's prototype](https://github.com/shoelace-style/shoelace/issues/1135#issuecomment-1405868933), so I went ahead and converted them into ponyfills, which fit perfectly with the changes made in this PR. The ponyfills are [published on NPM](https://www.npmjs.com/package/composed-offset-position) for further testing and feedback. I still need to add usage doc, but for now the [test page](https://github.com/jcfranco/composed-offset-position/blob/main/test/test.html#L94-L98) is the closest thing.

atomiks:
Thanks @jcfranco I added this to the docs and edited the OP

Issue #1440, opened 2021-12-12
Support boundaries with “holes” aka detectOverflow relative to arbitrary elements


atomiks:
Expands on https://github.com/floating-ui/floating-ui/issues/1381

There have been feature requests related to collision detection of other floating elements, not just a surrounding boundaries. To do this it looks like we have to think of boundaries as a polygon with "holes", the holes represent other arbitrary elements to detectOverflow to, not just a surrounding rectangle boundary.

A picture is worth a thousand words:

<img width="400" alt="Screen Shot 2021-12-12 at 7 48 26 pm" src="https://user-images.githubusercontent.com/22450188/145706115-22047459-b2ba-46c8-99c8-bd8ea49bd8d5.png">

The holes represent other elements "eating" into the main boundary.

---

Right now you can either use `padding` in the `shift`/`flip` modifiers (only works for static things like a sticky header or navbar), or let the floating element overlap those other floating elements with a higher z-index.

KutnerUri:
Yup I also come across this - some banners or header might have a high z-index, and could cover the tooltip, making bad UX.

Ideally, popper would be "smart" enough to detect all z-index items around it and avoid them. Obviously, this is not practical.

Any best practices? 

for example:
![Screenshot 2022-04-30 at 2 55 18 PM](https://user-images.githubusercontent.com/27219302/166121558-2217dabb-9c03-4b00-8bf2-f5dc22b1fc14.png)

atomiks:
The only solution is using `padding` in `flip` or `shift` which takes into account the other positioned element, or if your app is not scrollable and instead has individual scrolling containers (e.g. like a desktop UI), you can use a custom boundary to limit it from overflowing its container, thereby not overflowing other positioned elements. 

I think this problem is hard to solve and I'm not sure how to even start on the "holes" thing.

wlib:
Maybe I misunderstood the problem, but here are my immediate thoughts:

Ideally, the DOM would provide a `document.nodesWithinRect()` sort of like [`document.elementsFromPoint`](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementsFromPoint), and then we could just test if there is any intersection of the returned set with the set of nodes that shouldn't be overlapped. Since we don't have this DOM API, the reverse could be done: maintain the set of nodes that shouldn't be overlapped, tracking overlap of their bounding boxes with our target floating element. If any of these tracked nodes are positioned by floating-ui, then that's a little performance shortcut.

[...]

PR #2161, opened 2023-02-03
feat(detectOverflow): accept virtual `Rect` boundaries

### LINKED ISSUE #2084: [DOM] Improve visual vs layout viewport logic (pinch-zooming)
Right now, `'viewport'` in `detectOverflow#rootBoundary` corresponds to the [Visual Viewport](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API) mainly because of the iOS software keyboard — things like combobox dropdowns shouldn't be obscured beneath the keyboard, but in general it's more reliable on mobile for all floating elements.

This also has the positive (in my opinion) side-effect of ensuring the floating element doesn't get cut off outside of the visual viewport if the user pinch-zoomed **before** opening a floating element. However, this has the downside of the floating element shifting and moving around while pinch-zooming **after** it's open, which can be distracting or not desired at all in certain cases.

So we should:

- Possibly add a `layoutViewport` option for `rootBoundary`.
- Possibly add ability to ignore the `visualViewport` in `autoUpdate`.

Or even better, find the specific combination of the above two that manages to preserve the behavior on iOS/mobile viewports and initial pinch-zooming, while removing the undesired behavior of causing it to adjust while pinch-zooming if it's already open, without needing any configuration.

atomiks:
Closes #2084

This enables "virtual" boundaries, and applies to both `boundary` and `rootBoundary` options.  Useful for more than just the linked issue though.

So if you wanted to use the layout viewport rather than the visual viewport, you could do:

```js
detectOverflow(args, {
  rootBoundary: {
    x: 0,
    y: 0,
    width: document.documentElement.clientWidth,
    height: document.documentElement.clientHeight,
  }
});
```

I think it's best to keep the default `viewport` `rootBoundary` if the visual viewport and layout viewport dimensions aren't the same upon first render of a floating element.  This heuristic isn't perfect, but this way it will correctly re-adjust its position when pinch-zooming out if it was flipped or shifted initially when opening, while preventing it from changing position while pinch-zooming _in_ which is the main issue.

```js
  useEffect(() => {
   if (!isOpen) return;

    const visualWidth = visualViewport?.width;
    const visualHeight = visualViewport?.height;
    const layoutWidth = document.documentElement.clientWidth;
    const layoutHeight = document.documentElement.clientHeight;

    // They aren't pinch-zoomed in / software keyboard is not open, etc, so we
    // use the layout viewport now to prevent the floating element from moving
    // when pinch-zooming around. Could also use a threshold of difference.
    // 
    // I haven't actually tested this on iOS, but seems to work on macOS Chrome.
    if (visualWidth === layoutWidth && visualHeight === layoutHeight) {
      setRootBoundary({
        x: 0,
        y: 0,
        width: layoutWidth,
        height: layoutHeight,
      });
    }
  }, [isOpen]);
```

Going even further, you could also check if `middlewareData.shift` contains coords other than `0`, and if the final `placement` is not the preferred `placement`, etc to determine to use the layout viewport for a given render.

### DIFF
--- a/packages/dom/src/types.ts
+++ b/packages/dom/src/types.ts
@@ -58,7 +58,7 @@ export interface NodeScroll {
   scrollTop: number;
 }
 
-export type Boundary = 'clippingAncestors' | Element | Array<Element>;
+export type Boundary = 'clippingAncestors' | Element | Array<Element> | Rect;
 
 export type DetectOverflowOptions = Omit<
   CoreDetectOverflowOptions,
--- a/packages/core/src/detectOverflow.ts
+++ b/packages/core/src/detectOverflow.ts
@@ -14,12 +14,12 @@ const DEBUG_RECTS = false;
 
 export interface Options {
   /**
-   * The clipping element(s) in which overflow will be checked.
+   * The clipping element(s) or area in which overflow will be checked.
    * @default 'clippingAncestors'
    */
   boundary: Boundary;
   /**
-   * The root clipping element in which overflow will be checked.
+   * The root clipping area in which overflow will be checked.
    * @default 'viewport'
    */
   rootBoundary: RootBoundary;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -151,7 +151,7 @@ export interface MiddlewareArguments extends Coords {
 export type ClientRectObject = Rect & SideObject;
 export type Padding = number | Partial<SideObject>;
 export type Boundary = any;
-export type RootBoundary = 'viewport' | 'document';
+export type RootBoundary = 'viewport' | 'document' | Rect;
 export type ElementContext = 'reference' | 'floating';
 
 export {computePosition} from './computePosition';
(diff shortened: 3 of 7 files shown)


packages/dom/src/types.ts
```diff
@@ -55,7 +55,7 @@ export interface NodeScroll {
   scrollTop: number;
 }
 
-export type Boundary = 'clippingAncestors' | Element | Array<Element>;
+export type Boundary = 'clippingAncestors' | Element | Array<Element> | Rect;
```
FezVrasta, line 58: Don't you need to know the position in the DOM of the boundary for the logic to work?

atomiks: No, clipping is detected only relative to the viewport (`.getBoundingClientRect()`)

atomiks:
This should also enable a tree-shakeable API for the "easy" algorithm in #1440:

```js
rootBoundary: createViewportRect({
  // avoid these elements! e.g. sticky navbar or sidebar
  holes: '[data-floating-ui-viewport-hole]',

  // other options could be:
  // use or not use the visualViewport?
  visual: false,
})
```

PR #2163, opened 2023-02-04
fix(size): take into account `shift()` for center aligned placements

### LINKED ISSUE #1897: [Core] `size` middleware gives wrong `availableWidth`

atomiks:
The `crossAxis` of `size()`, i.e. the y-axis for left/right placements, or x-axis for top/bottom placements poses some problems.

For center aligned placements, this PR considers whole viewport space when `shift()` appears before `size()`, which fixes #1897.

For edge aligned placements (`start` / `end`), I found the following:

- It's not desirable to use `shift()` **before** `size()`, because then the placement loses its alignment. Placing it after works fine with desirable behavior. Perhaps should be documented.
- When using `flip()`, which is much more useful, it "resets" at a point when it overflows on both sides, which looks like it was caused by https://github.com/floating-ui/floating-ui/pull/2151. Will try to include a fix for that in this PR also.

### DIFF
--- a/packages/core/src/middleware/flip.ts
+++ b/packages/core/src/middleware/flip.ts
@@ -134,10 +134,11 @@ export const flip = (
         };
       }
 
-      // First, try to use the one that fits on mainAxis side of overflow.
-      let resetPlacement = overflowsData.find(
-        (d) => d.overflows[0] <= 0
-      )?.placement;
+      // First, find the candidates that fit on the mainAxis side of overflow,
+      // then find the placement that fits the best on the main crossAxis side.
+      let resetPlacement = overflowsData
+        .filter((d) => d.overflows[0] <= 0)
+        .sort((a, b) => a.overflows[1] - b.overflows[1])[0]?.placement;
 
       // Otherwise fallback.
       if (!resetPlacement) {
--- a/packages/core/src/middleware/size.ts
+++ b/packages/core/src/middleware/size.ts
@@ -4,6 +4,7 @@ import {
 } from '../detectOverflow';
 import type {Middleware, MiddlewareArguments} from '../types';
 import {getAlignment} from '../utils/getAlignment';
+import {getMainAxisFromPlacement} from '../utils/getMainAxisFromPlacement';
 import {getSide} from '../utils/getSide';
 import {max} from '../utils/math';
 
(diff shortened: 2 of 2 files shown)


Westbrook:
Thanks for this! Seems to fix my repro AND my source case, though I was _certain_ they were examples of different problems 🥳

benface:
> For edge aligned placements (`start` / `end`), I found the following:
> It's not desirable to use `shift()` **before** `size()`, because then the placement loses its alignment. Placing it after works fine with desirable behavior.

@atomiks – Interesting; why is keeping the alignment prioritized over maximizing the width of the floated element only for edge-aligned placements? I would love to have this new `center` behavior with `left` and `right` as well, where the floated element is allowed to shift until there is no more available space, and only then should it shrink. To clarify:

- With `size()` before `shift()`...
  - all placements get the "wrong" behavior, where the element shrinks even though there's enough space for it to be shifted.
- With `shift()` before `size()`...
  - `center` gets the "right" behavior, where the element is shifted until there's _really_ no more available space, after which it starts shrinking.
  - `left` and `right`, however, still get the "wrong" behavior, at least on the initial layout/computation... on subsequent ones (so if I resize the viewport while a floating element is visible), it actually works like I want. It's kinda strange, see this video:

https://github.com/floating-ui/floating-ui/assets/1059139/43c8fdda-82ff-4c27-88ec-a65e6132eaad

atomiks:
@benface might be similar to this https://github.com/floating-ui/floating-ui/discussions/2953

benface:
Huh, actually, I sent that comment bit too soon. Strangely, it only seems to work well in Firefox (and Firefox-based browsers, like the one I use, Zen). I just tried my CodeSandbox in Chrome and Safari and I'm getting a slightly different issue than [the one I mentioned initially](https://github.com/floating-ui/floating-ui/pull/2163#issuecomment-2192560190), but potentially related? Here's a video of the strange behavior:

https://github.com/user-attachments/assets/e6aa3fcb-3ab7-4233-b483-1fd2d7af911c

Looks like when there's not enough space for the whole menu, the `align="end"` starts behaving like `align="start"`. Interestingly, there's no issue with `align="start"`.

PR #2164, opened 2023-02-04
chore: try to speed up ci

atomiks:
This might, or probably will, break, but I'm trying to speed up the test suite in CI as it's slow when trying to add new tests and update the snapshots, and performing a release.

- **Switches (back) to Linux for visual snapshots**: macOS is much slower than Linux, it was used because Linux was flakey sometimes (although macOS can be too, it's far less frequent). Hopefully it's not anymore?
- **Upgrade Node and Playwright versions**: Uses Node 18, not 17 as there were warnings being emitted. Also Playwright was locked to 1.17.x because something was broken in 1.18.x. Can't remember, but I'm hoping it's fixed in the latest version as it's now pretty old.
- **Caching**: I think npm install caching was broken? Upgrades to v3 versions of the actions and uses `npm` cache option
- **Increase workers and use "fullyParallel" option**: I don't know if this will work or break something, but let's try it

---

### Might also implement these in this PR as they seem important for speed:

- Reduce number of screenshots for the shadow DOM test suite — this one appears to be by far the slowest as it has a combinatorial explosion of state being tested.
- Don't use screenshots for `offset()`, just use regular unit tests. It's not really something that needs screenshots of, and it's the second slowest test suite.


atomiks:
Looks much faster now:

- Linting and type checking ~2x faster
- Unit tests ~1.2x faster
- Functional tests ~3x faster

Going to check for validity and flakiness then merge

PR #2170, opened 2023-02-07
Prevent fixed ancestors from being a clipping ancestor

### LINKED ISSUE #2167: Should `strategy: fixed` be contained by elements with `transform: ...`?
As seen in https://github.com/floating-ui/floating-ui/blob/db4b7ad129b3f79b85184c3c30f352daab824cc3/packages/dom/src/utils/is.ts#L49-L74 an element is considered containing whenever is has been given a `transform` rule that does not equal `none`. This makes sense for `strategy: "absolute"` positioned content, but when `strategy: "fixed"` is used this can errantly contain the `fixed` content to an ancestor with `overflow` rules when an interceding element has a `transform` rule. While https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block does say that an element with a `transform` rule _"MAY"_ form a containing block, I'm not so sure that is actually does for our purposes when using `strategy: "fixed"`.

Repro: https://codesandbox.io/s/magical-sky-vbbnqk?file=/index.html

Do you think that `strategy` could/should be taken into account when gathering whether an element is a containing block? Or do you think there is some other reasonable API that could be leveraged to ignore content of this type as being a containing block? It seems that [`getClippingRect`](https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/utils/getClippingRect.ts#L149-L152) is incorrectly calculated without one of these two options being available to consumers.

Happy to submit a PR if we can work out an acceptable path forward for this.

Westbrook:
fixes #2167

Previously ancestors with `transform` values and interceding decendents with `position: fixed` would errantly count as clipping ancestors. This releases the styles of a `position: fixed` ancestor so that it doesn't allow the `transform` ancestor to join the clipping ancestors array.

Repro in previous context: https://codesandbox.io/s/youthful-wiles-dc6oy4?file=/index.html

### DIFF
--- a/packages/dom/src/utils/getClippingRect.ts
+++ b/packages/dom/src/utils/getClippingRect.ts
@@ -105,7 +105,7 @@ function getClippingElementAncestors(
     const computedStyle = getComputedStyle(currentNode);
     const containingBlock = isContainingBlock(currentNode);
 
-    const shouldIgnoreCurrentNode = ['fixed'].includes(computedStyle.position);
+    const shouldIgnoreCurrentNode = computedStyle.position === 'fixed';
     if (shouldIgnoreCurrentNode) {
       currentContainingBlockComputedStyle = null;
     } else {
--- a/packages/dom/src/utils/getClippingRect.ts
+++ b/packages/dom/src/utils/getClippingRect.ts
@@ -105,21 +105,26 @@ function getClippingElementAncestors(
     const computedStyle = getComputedStyle(currentNode);
     const containingBlock = isContainingBlock(currentNode);
 
-    const shouldDropCurrentNode = elementIsFixed
-      ? !containingBlock && !currentContainingBlockComputedStyle
-      : !containingBlock &&
-        computedStyle.position === 'static' &&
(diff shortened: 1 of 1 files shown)


packages/dom/src/utils/getClippingRect.ts
```diff
@@ -105,21 +105,26 @@ function getClippingElementAncestors(
...
-        computedStyle.position === 'static' &&
-        !!currentContainingBlockComputedStyle &&
-        ['absolute', 'fixed'].includes(
-          currentContainingBlockComputedStyle.position
-        );
-
-    if (shouldDropCurrentNode) {
-      // Drop non-containing blocks.
-      result = result.filter((ancestor) => ancestor !== currentNode);
+    const shouldIgnoreCurrentNode = ['fixed'].includes(computedStyle.position);
```
atomiks, line 108:
```suggestion
    const shouldIgnoreCurrentNode = computedStyle.position === 'fixed';
```

atomiks:
Thanks!

PR #2173, opened 2023-02-08
fix(useTypeahead): allow type ahead when currentTarget contains target

### LINKED ISSUE #2172: [React] useTypeahead with different currentTarget
Having combobox with structure below, would like to enable useTypeahead when searchable is disabled, 

```jsx
<div {...getReferenceProps({ className: styles.wrapper })>
  <Input readOnly={!searchable}>
  <div>clear<div>
  <div>toggle menu<div>
</div>
```
Issue is here https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/hooks/useTypeahead.ts#L87-L96 basic activeElement is input, but currentTarget is the wrapper.

As wrapper will have role `combox`, when adding it to list of `target.closest` should work as expected https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/hooks/useTypeahead.ts#L91

piecyk:
fixes #2172

### DIFF
--- a/packages/react/src/hooks/useTypeahead.ts
+++ b/packages/react/src/hooks/useTypeahead.ts
@@ -32,7 +32,7 @@ export interface Props {
  * @see https://floating-ui.com/docs/useTypeahead
  */
 export const useTypeahead = <RT extends ReferenceType = ReferenceType>(
-  {open, dataRef}: FloatingContext<RT>,
+  {open, dataRef, refs}: FloatingContext<RT>,
   {
     listRef,
     activeIndex,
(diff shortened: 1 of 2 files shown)


atomiks:
Doesn't that break this comment requirement?

```js
      // Correctly scope nested non-portalled floating elements. Since the nested
      // floating element is inside of the another, we find the closest role
      // that indicates the floating element scope.
```

If so, maybe a callback filter to let the user decide whether it should work is best?

piecyk:
> Doesn't that break this comment requirement?
> 
> ```js
>       // Correctly scope nested non-portalled floating elements. Since the nested
>       // floating element is inside of the another, we find the closest role
>       // that indicates the floating element scope.
> ```
> 
> If so, maybe a callback filter to let the user decide whether it should work is best?

hmm good questions. This contains should only be checked in context of reference, current code works because assumes that reference is the active element when validating useTypeahead, otherwise we are in floating elements.

atomiks:
Thanks! Can you add a test in `useTypeahead.test.tsx`?

piecyk:
> Thanks! Can you add a test in `useTypeahead.test.tsx`?

@atomiks added.

atomiks:
Thanks!

PR #2179, opened 2023-02-11
feat(react): `useClientPoint` hook

### LINKED ISSUE #2150: [React] Hook for cursor/point-based virtual element position
It's already possible to position a floating element at the cursor point, but this would provide a simpler API for things like:

- Following the cursor
- Context menus
- etc

atomiks:
Closes #2150

### DIFF
--- a/packages/react/src/hooks/useClientPoint.ts
+++ b/packages/react/src/hooks/useClientPoint.ts
@@ -21,6 +21,7 @@ function createVirtualElement(
   let isAutoUpdateEvent = false;
 
   return {
+    contextElement: domRef.current || undefined,
     getBoundingClientRect() {
       const domRect = domRef.current?.getBoundingClientRect() || {
         width: 0,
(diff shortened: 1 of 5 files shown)


emroot:
Hey @atomiks, thanks for working on this PR.
looks like this should work with virtual elements right?
Looking for this to land, trying to build a floating toolbar for prosemirror where I only have coordinates and not a real dom element.
Maybe this could be renamed `useVirtualPoint` instead to expand use-cases?
what do you think?

atomiks:
I can see you might want to use a virtual rect in general, so `useVirtualRect` / `useRect`, but I think that is a separate hook. Something like `useRangeSelection` is higher-level with different behavior, and the concerns should be separated from a "point"-like hook, as this is.

```js
useRect(context, {
  x: number,
  y: number,
  width: number,
  height: number
});
```

```js
function useVirtualRect(context, rect) {
  useLayoutEffect(() => {
    context.refs.setPositionReference({
      contextElement: context.elements.domReference || undefined,
      getBoundingClientRect() {
        return ...;
      }
    });
  }, [context.refs, context.elements, rect.x, rect.y, rect.width, rect.height]);
}
```

Issue #2078, opened 2022-12-30
Export more useful utilities to simplify usage


atomiks:
There are utilities like `autoUpdate`, but the library is missing many more that could be useful. Of course, the library still gives full control, but common use cases should be handled more easily. This issue applies to most of the packages, although the React package added several utilities recently.

## Virtual elements

A point reference is the most common one:

```js
createVirtualElement({
  width: 0,
  height: 0,
  x: clientX,
  y: clientY,
});
```

I think range selections and modifying/deriving `getBoundingClientRect` on a real element are already simple enough?

## Arrows

These are by far the biggest pain — we should simplify both applying arrow styles and creating an actual arrow element in the first place.  We should export function(s)/components that either generate an SVG or provide data to generate one. 

SVGs are less limited than the most basic CSS styles, as they can:

- Have a border
- Allow the floating element to be partially transparent (no overlapping problem)
- Are flexible when it comes to roundedness, shape, etc.

Options could include:

- type (isosceles or right-angle)
- angle
- size
- roundness
- color
- border

Since `offset.mainAxis` usually needs to equal the hypotenuse length of the arrow, a utility for that could also be included.

## Other

If you have ideas, please comment!

mihkeleidast:
From the [testing docs](https://floating-ui.com/docs/react#testing), the `waitForPosition` testing utility would perhaps be useful to export.

atomiks:
For arrows, I'm not actually sure if this should be part of the library code, but rather, just some generator thing on the docs. 

I have no idea how to generate unique rounded shapes like this programmatically in code (made this in Figma):

<img width="805" alt="Screenshot of rounded arrows on the four sides" src="https://user-images.githubusercontent.com/22450188/211826931-f29987ed-d9ca-4089-90c7-1885eca48dd2.png">

The docs could maybe just have a bunch of copy & pasteable SVG shapes with some CodeSandbox links showing how to use each one of them.

PR #2195, opened 2023-02-26
feat(react): `FloatingArrow` component

atomiks:
Partially solves #2078

This component is designed to be flexible and support the vast majority of use cases.

- **Simple to setup**: it renders out a 14x7px black-filled SVG arrow by default. 

Placing this inside your tooltip will position everything without any styling required:

```js
<FloatingArrow ref={arrowRef} context={context} />
```

- **Customizable width and height (and therefore `angle`)** via `width` and `height` attributes.
- **Rounded tip**: the `tipRadius` prop accepts a number similar to `border-radius`.
- **Borders**: the `stroke`/`strokeWidth` props can add a border to the arrow.
- **Custom shapes**: the `d` prop accepts a custom path string to create more fancy rounding - the docs will have an example that can be edited via Figma or similar
- **Static offsets**: the `staticOffset` prop allows the dynamic position to be overridden by a static offset under any condition, usually when the floating element is smaller than the reference element along the relevant axis and has an edge alignment.
- **Transparency**: possible as it's a real triangle not a rotated square which would cause overlap darkening.

---

What it doesn't solve but could in the future:

- Skewing via `centerOffset` or even `edgeOffset` when using `staticOffset`, etc.
- Advanced `clip-path` use cases to create an arrow out of a `polygon()` (Wikipedia's hovercards)
- Other things I haven't thought of?

## Caveats

- A gap can appear in Safari when the `x`/`y` coords aren't integers (or placed evenly on the device's subpixel grid). Will document this.
- In Firefox, there's a 0.5px gap that needs a stroke on the "fill" triangle, which I've added as a fix, though it makes it 1px bigger than the value specified when using a stroke.

### DIFF
--- a/packages/react/src/components/FloatingArrow.tsx
+++ b/packages/react/src/components/FloatingArrow.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import {platform} from '../';
+import {platform} from '..';
 import type {Alignment, FloatingContext, Side} from '../types';
 
 export interface Props extends React.SVGAttributes<SVGSVGElement> {
@@ -47,9 +47,9 @@ export interface Props extends React.SVGAttributes<SVGSVGElement> {
 
 /**
  * Renders a pointing arrow triangle.
- * @see https://floating-ui.com/docs/ArrowComponent
+ * @see https://floating-ui.com/docs/FloatingArrow
  */
-export const Arrow = React.forwardRef(function Arrow(
+export const FloatingArrow = React.forwardRef(function FloatingArrow(
   {
     context: {
       placement,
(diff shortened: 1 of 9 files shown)


packages/react/src/components/FloatingArrow.tsx
```diff
@@ -0,0 +1,139 @@
...
+   * Height of the arrow.
+   * @default 7
+   */
+  height?: number;
+
+  /**
+   * The roundness of the arrow (0 - 100+)
+   * @default 0 (sharp)
+   */
+  roundness?: number;
```
mihkeleidast, line 25: Did you consider "radius" (as in border radius or corner radius) for the prop name as well?

PR #2203, opened 2023-03-03
fix: share root id via context

mihkeleidast:
Relates to this: https://github.com/floating-ui/floating-ui/issues/2128#issuecomment-1452077329

When an external reference is using its own set of interactions hooks because it needs to configure them differently per reference, they should still use the same floating element ID reference throughout.

### DIFF
--- a/packages/react/src/types.ts
+++ b/packages/react/src/types.ts
@@ -129,6 +129,7 @@ export type FloatingContext<RT extends ReferenceType = ReferenceType> =
       events: FloatingEvents;
       dataRef: React.MutableRefObject<ContextData>;
       nodeId: string | undefined;
+      rootId: string;
       refs: ExtendedRefs<RT>;
       elements: ExtendedElements<RT>;
     }
--- a/packages/react/src/types.ts
+++ b/packages/react/src/types.ts
@@ -129,7 +129,7 @@ export type FloatingContext<RT extends ReferenceType = ReferenceType> =
       events: FloatingEvents;
       dataRef: React.MutableRefObject<ContextData>;
       nodeId: string | undefined;
-      rootId: string;
+      floatingId: string;
       refs: ExtendedRefs<RT>;
       elements: ExtendedElements<RT>;
     }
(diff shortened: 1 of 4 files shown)


mihkeleidast:
Not sure if the functional test failure (timeout) is something I did.

packages/react/test/unit/useRole.test.tsx
```diff
@@ -80,6 +116,33 @@ describe('dialog', () => {
 
     cleanup();
   });
+
+  test('sets correct aria attributs with external ref, multiple useRole calls', () => {
```
atomiks, line 120:
```suggestion
  test('sets correct aria attributes with external ref, multiple useRole calls', () => {
```

packages/react/src/useFloating.ts
```diff
@@ -3,8 +3,9 @@ import * as React from 'react';
 import useLayoutEffect from 'use-isomorphic-layout-effect';
 
 import {useFloatingTree} from './components/FloatingTree';
+import {useId} from './hooks/useId';
 import {useEvent} from './hooks/utils/useEvent';
-import type {
+import {
```
atomiks, line 8: `type`?

mihkeleidast: Oops, my VS Code has a bad habit of removing these annotations when auto importing stuff.

packages/react/src/types.ts
```diff
@@ -129,6 +129,7 @@ export type FloatingContext<RT extends ReferenceType = ReferenceType> =
       events: FloatingEvents;
       dataRef: React.MutableRefObject<ContextData>;
       nodeId: string | undefined;
+      rootId: string;
```
mihkeleidast, line 132: One thing I started thinking abut is if this variable name makes the most sense, because this will become a part of the public API. Perhaps `floatingId` or something like that would be more logical, as it would connect with the reference/floating element naming logic better? There is also `nodeId` already in here, which to me feels like something similar, though this needs to be provided externally currently. Perhaps there is a possibility to use the same ID for both usecases?

atomiks: I do think `floatingId` is better indeed 👍

PR #2204, opened 2023-03-04
React/Vue v1

atomiks:
Opening this for discussion, it's not ready for merging. /cc @mihkeleidast

The positioning parts here also apply to the `react-native` and `vue` packages, everything else is for `react`. 

As Mihkel pointed out, we need to make sure features like singletons and other things he pointed out don't require the API to break to work in the future, so let's make sure we future-proof this 😅 

---

## `x`/`y` default to `0` instead of `null`

`isPositioned` is a better way of checking if the floating element has been positioned. 

This removes the need to do `?? 0` although that will still work anyway, so not breaking for many use cases.

```js
const {x, y, strategy} = useFloating();
<div 
  style={{
    position: strategy,
    left: x,
    top: y,
  }} 
/>
```

## Enforce a cleanup function for `whileElementsMounted` type

Many people accidentally omitted the cleanup function for `autoUpdate` so I had to highlight it strongly in the docs — I don't know if you ever want to use it without a cleanup function. If there are complaints I can easily re-allow it without any breakage anyway.

## Renamed `FloatingOverlay` to `FloatingBackdrop`

More descriptive of its function, matches `::backdrop` CSS selector.

## Renamed `FloatingDelayGroup` to `FloatingGroup`

As discovered by @mihkeleidast it has some useful properties and could be extended beyond using delays. This more generic name allows it to do more things if we want in the future.

## Renamed `useHover#mouseOnly` to `ignoreTouch`

I still need to investigate if these need changing:

- `ignoreMouse` in `useClick`
- `keyboardOnly` in `useFocus`
- Should `move: true` be default in `useHover`? It's useful in some cases, but I'm not sure if it should be default.
- Should `restMs` be `openRestMs` as it doesn't apply to `close`? And `restMs` in `safePolygon()` be `destroyRestMs`? Think certain names can be more descriptive here.

## Made `useInteractions` and `useMergeRefs` variadic instead of accepting an array

Looking online, most of these functions use `...rest` instead of an array, looks a bit cleaner as the array isn't necessary. Neither accept a second argument and I doubt they ever will.

## Remove `reference()` virtual element support

Replaced by `refs.setPositionReference(...)`

## Replace public refs with instance variables

Convenience:

- `tree.nodesRef.current` => `tree.nodes`
- `context.dataRef.current` => `context.data`

---

- I haven't removed `reference`, `floating`, `positionReference` top-level callbacks. I deprecated them in the types, should we remove them?

### DIFF
--- a/packages/react-dom/src/types.ts
+++ b/packages/react-dom/src/types.ts
@@ -70,9 +70,7 @@ type Prettify<T> = {
 } & {};
 
 export type UseFloatingData = Prettify<
-  Omit<ComputePositionReturn, 'x' | 'y'> & {
-    x: number | null;
-    y: number | null;
+  ComputePositionReturn & {
     isPositioned: boolean;
   }
 >;
@@ -115,7 +113,7 @@ export type UseFloatingProps<RT extends ReferenceType = ReferenceType> =
         reference: RT,
         floating: HTMLElement,
         update: () => void
-      ) => void | (() => void);
+      ) => () => void;
       open?: boolean;
     }
   >;
--- a/packages/react-dom/src/useFloating.ts
+++ b/packages/react-dom/src/useFloating.ts
@@ -30,8 +30,8 @@ export function useFloating<RT extends ReferenceType = ReferenceType>(
   } = options;
 
   const [data, setData] = React.useState<UseFloatingData>({
-    x: null,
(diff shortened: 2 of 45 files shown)


atomiks:
Closed in favor of #2300

PR #2215, opened 2023-03-09
fix(FloatingFocusManager): mutate internal preventReturnFocusRef instead of returnFocusRef

mihkeleidast:
We ran into a problem with a Popover that uses the transition hook, the main structure was like this:

```
<Popover id="a">
  <Popover id="b">
  </Popover>
</Popover>
<Popover id="c">
</Popover>
```

The bug occurred when:
1. open Popover a
2. open nested Popover b
3. open Popover c, which should close both a and b with an outside press
4. but Popover c also closed

Since a and b closed with transition timeouts, the layout effect cleanup function also run with a timeout. This surfaced an issue where in the cleanup function, `returnFocusRef.current` was not `false` anymore, even though it was set to false in the `onDismiss` function. I assume this is because `returnFocusRef` depends on the prop value, and with the transitions, it gets rerendered and reset at some point. I think this is a bug, internally we shouldn't mutate the ref that comes from the prop. Instead, we should mutate the internal `preventReturnFocusRef` when necessary.

I tried writing a test for it, but failed, since this issue only seemed to happen with transition timeouts, which would need to be mocked with fake timers in tests. Hoping the change makes sense without a test/repro as well.

### DIFF
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -366,10 +366,10 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
       const returnFocus = payload.data.returnFocus;
 
       if (typeof returnFocus === 'object') {
-        preventReturnFocusRef.current = false;
+        returnFocusRef.current = true;
         preventReturnFocusScroll = returnFocus.preventScroll;
       } else {
-        preventReturnFocusRef.current = !returnFocus;
+        returnFocusRef.current = returnFocus;
       }
     }
 
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -366,10 +366,10 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
       const returnFocus = payload.data.returnFocus;
 
       if (typeof returnFocus === 'object') {
-        returnFocusRef.current = true;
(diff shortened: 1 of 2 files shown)


mihkeleidast:
I pushed a repro in the playground (that I reverted as well) if you want to verify this issue/fix in more detail.

1. Checkout https://github.com/floating-ui/floating-ui/pull/2215/commits/c38e2878c5b953d9d967f0d5d3969b287e72f371
2. Go to Popover in the playground, uncheck "modal" checkbox
3. Click on "first button"
4. Click on "first second button"
5. Click on "second button"
6. Wait a second or so, so all the transition timeouts would finish.
7. Observe that all Popovers are closed, though it is expected that the second popover would remain open.

packages/react/src/components/FloatingFocusManager.tsx
```diff
@@ -383,6 +383,7 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
       }
 
       if (
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
atomiks, line 386: Why is this here?

mihkeleidast: ESLint started complaining after my change, so I figured disabling it with a comment is better than to leave the warning in the console. If I copy it to a variable in the hook, some tests fail (see the first commit in this PR).

atomiks:
Thanks!

PR #2229, opened 2023-03-14
fix: Use correct window to get the devicePixelRatio

ling1726:
The `roundOffsetsByDPR` utility was using the global window keyword instead of the `getWindow` utility. This meant that applications that render to childe windows would have blurry poppers sincee the device pixel ratio being used wouldn't be correct.Two cases where that could happen:

* Parent window in another monitor
* Parent window has a different zoom than the child window

### DIFF
--- a/src/modifiers/computeStyles.js
+++ b/src/modifiers/computeStyles.js
@@ -46,8 +46,7 @@ const unsetSides = {
 // Round the offsets to the nearest suitable subpixel based on the DPR.
 // Zooming can change the DPR, but it seems to report a value that will
 // cleanly divide the values into the appropriate subpixels.
-function roundOffsetsByDPR({ x, y }): Offsets {
-  const win: Window = window;
+function roundOffsetsByDPR({ x, y }, win: Window): Offsets {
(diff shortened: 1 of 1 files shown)


FezVrasta (approved):

ling1726:
@FezVrasta do you mind taking a look at the CI failures? I don't have enough understanding of the repo to know what action to take on those errors

ling1726:
@FezVrasta  @atomiks  would you be able to help with the errors I'm seeing in the checks? 🙏🙏

FezVrasta:
It seems to be a problem with CI itself but I'm not sure what's going on to be fair. @atomiks may have some ideas?

atomiks:
Don't know sorry, guess some packages/actions had some breaking changes not in semver or something :| 
Probably can just merge and release regardless, @FezVrasta has to release locally though as Rolling Versions somehow can't release the v2.x branch.

FezVrasta:
https://github.com/floating-ui/floating-ui/releases/tag/v2.11.7

PR #2244, opened 2023-03-22
fix(useInteractions): merging function props that return values

mihkeleidast:
Our usecase for this is that our buttons accept Promise-returning onClick handlers, which put the button to a loading state while the promise resolves. This behavior broke when a button was wrapped with a Tooltip, as the prop merging here did not forward the returned promise.


mihkeleidast:
@atomiks can you release this? :)

PR #2264, opened 2023-03-30
fix(FloatingFocusManager): respect child with autoFocus

mihkeleidast:
I suppose this could be configured with initialFocus, but this would seem like an unnecessary configuration to me. I think users would expect autoFocus inside a dialog to just work.

### DIFF
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -350,12 +350,15 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
       previouslyFocusedElement
     );
 
-    // If the `useListNavigation` hook is active, always ignore `initialFocus`
-    // because it has its own handling of the initial focus.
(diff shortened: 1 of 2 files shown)


packages/react/src/components/FloatingFocusManager.tsx
```diff
@@ -345,9 +345,15 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>({
...
+      floating,
+      previouslyFocusedElement
+    );
+
     // If the `useListNavigation` hook is active, always ignore `initialFocus`
     // because it has its own handling of the initial focus.
     !ignoreInitialFocus &&
+      !focusAlreadyInsideFloatingEl &&
       open &&
       enqueueFocus(elToFocus, {preventScroll: elToFocus === floating});
```
atomiks, line 358: Guess this should be made an `if` now as it's getting unwieldy

mihkeleidast: Refactored.

PR #2269, opened 2023-04-02
feat: `FloatingList`

atomiks:
@pierreburel @Marsony

The `sortByDocumentPosition` technique is found in other libs to enable this feature, and this one doesn't write or read any refs during render, so hopefully good for concurrent features.

Usage:

`FloatingList` is a context provider for children

- `elementsRef` is `useListNavigation`'s `listRef`
- `labelsRef` is `useTypeahead`'s `listRef` which uses the `label` passed into `useListItem` or falls back to the node's `textContent`

```js
<FloatingList elementsRef={elementsRef} labelsRef={labelsRef}>
  ...
</FloatingList>
```

- `useListItem` reads that context and returns an `index` and a `ref`:

```js
const {activeIndex} = useSelectContext();
const {ref, index} = useListItem();
const isActive = activeIndex === index;

// No need to manually assign an index:
<div ref={ref} tabIndex={isActive ? 0 : -1} />
```

For controlled (pre-selected values) is there anything else we should do?  @pierreburel - your demo has `listValuesRef` but this is not included here

### DIFF
--- a/packages/react/src/components/FloatingList.tsx
+++ b/packages/react/src/components/FloatingList.tsx
@@ -1,0 +1,150 @@
+import * as React from 'react';
+import useLayoutEffect from 'use-isomorphic-layout-effect';
+
+function sortByDocumentPosition(a: Node, b: Node) {
+  const position = a.compareDocumentPosition(b);
+
+  if (
+    position & Node.DOCUMENT_POSITION_FOLLOWING ||
+    position & Node.DOCUMENT_POSITION_CONTAINED_BY
+  ) {
+    return -1;
+  }
+
+  if (
+    position & Node.DOCUMENT_POSITION_PRECEDING ||
+    position & Node.DOCUMENT_POSITION_CONTAINS
+  ) {
+    return 1;
+  }
+
+  return 0;
+}
+
+function areMapsEqual(
+  map1: Map<Node, number | null>,
+  map2: Map<Node, number | null>
+) {
+  if (map1.size !== map2.size) {
+    return false;
+  }
+  for (const [key, value] of map1.entries()) {
+    if (value !== map2.get(key)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+export const FloatingListContext = React.createContext<null | {
+  register: (node: Node) => void;
+  unregister: (node: Node) => void;
+  map: Map<Node, number | null>;
+  elementsRef: React.MutableRefObject<Array<HTMLElement | null>>;
+  stringsRef?: React.MutableRefObject<Array<string | null>>;
+}>(null);
+
+export function FloatingList({
+  children,
+  elementsRef,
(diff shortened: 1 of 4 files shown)


pierreburel:
Nice job! 

I really like the idea of passing an optional string for `useTypeahead` to `useListItem`, but i'm not sure about the `value` name tho? Maybe more `content` or `label`? In a `Select` component for example, `value` would more likely be an id or slug, so it may feels weird:

```jsx
const Option = ({ value, label, description }) => {
  const { setSelectedIndex, setValue } = useSelectContext();

  const { ref, index } = useListItem({ value: label }); // 👈🏻 here

  const handleSelect = () => {
    setSelectedIndex(index);
    setValue(value);
  };

  return (
    <div ref={ref} onClick={handleSelect}>
      <strong>{label}</strong>
      <small>{description}</small>
      <svg …/>
    </div>
  );
};
```

> For controlled (pre-selected values) is there anything else we should do? @pierreburel - your demo has listValuesRef but this is not included here

I guess for now we have to do that?

```jsx
const Option = ({ value, label, description }) => {
  const { selectedIndex, setSelectedIndex, selectedValue, setValue } = useSelectContext();

  const { ref, index } = useListItem();

  const handleRef = (node) => {
    ref(node);
    if (selectedValue === value && selectedIndex !== index) {
      setSelectedIndex(index);
    }
  };

  const handleSelect = () => {
    setSelectedIndex(index);
    setValue(value);
  };

  return <div ref={handleRef} onClick={handleSelect}>{label}</div>;
};
```

Any chance we could find a way to simply the ref part? Maybe by adding a parameter to `ref` or a prop to `useListItem`?

And yeah the `listValuesRef` was something i forgot to remove when i did some tests.

atomiks:
I like `label` indeed. For `selectedIndex` it could be a prop like `onMount`, or you could just a layout effect instead:

```js
useListItem({
  label: str,
  onMount() {
    if (selectedValue === value && selectedIndex !== index) {
      setSelectedIndex(index);
    }
  }
});
```

pierreburel:
Yeah it works well with a layout effect, maybe there's no need to overcomplicate things (like, would `onMount` needs a `onUnmount` too?).

In any case, LGTM! Amazing work as always 👍

atomiks:
Last question here is what's the best way to allow typeahead to work **before** the list has ever been rendered/opened, while maintaining conditional rendering. e.g. focus the button then type anything to select automatically without opening the list — a feature of native `<select>`.

The following works, but isn't very perf-friendly (children are rendered into a hidden div to register themselves). You could also avoid conditional rendering altogether, although `FloatingFocusManager` would need to be conditionally rendered as a parent still. 

```js
{!open && <div hidden>{children}</div>}
```

In some components I've also seen a hidden `select` being rendered so it works with form autofills too

Issue #2192, opened 2023-02-21
[React] Support other floating element use cases


atomiks:
https://github.com/floating-ui/floating-ui/discussions/2191

Since this library is called "Floating UI", we can and should do more than just support elements that use anchor positioning — for example, a non-anchored floating Dialog or Drawer component are already possible to build using the existing interaction primitives, but a component like a Toast or anything draggable isn't despite them also being "floating" elements.

Non-anchored UI elements that also float:

- Dialog (already supported)
- Drawer (already supported)
- Toast
- Draggable window
- etc

takurinton:
@atomiks 
Hi atomiks, Thanks for this great package.
I wanted a toast component using floating-ui and was actually working on a prototype at hand.

source code: https://github.com/takurinton/floating-ui-toast
demo: https://floating-ui-toast.vercel.app
codesandbox: https://codesandbox.io/s/infallible-swanson-fbdx15?file=/src/App.tsx

This prototype is in the prototype stage at this time and has issues with accessibility, state management, and interface.
I am inclined to clean up this prototype a bit more and am actually trying to work on it. 

I am inclined to contribute to floating-ui, can I add this as a demo for the toast component of floating-ui??

atomiks:
Nice start @takurinton, any component should follow the [ARIA design pattern](https://www.w3.org/TR/wai-aria/#aria-live) and these features:

- Pauses closing on hover, focus and window blur
- Supports hotkey to jump to toast viewport 
- Supports closing via swipe gesture

I think an absolute flexbox column is better for stacking than each being individually positioned with a dynamic `top`?

atomiks:
The `useRole` hook doesn't support it because it only references `aria-haspopup` roles, whereas the toast element needs to have this structure:

```js
<ul role="region" aria-live="polite">
  <li role="status" aria-atomic="true">
     Toast content
  </li>
</ul>
```

`useDismiss` works for `esc` dismissal, and `FloatingFocusManager` allows it to be interactive (e.g. to press an Undo button inside of it), though. The pattern isn't very complex however, and should be straightforward to support as seen above.

takurinton:
Thanks support. I am working on implementing the above.  

One question,(in https://github.com/floating-ui/floating-ui/issues/2192#issuecomment-1495005516)

> - Supports closing via swipe gesture

Is this supported by the existing floating-ui hooks? Or do I need to implement my own hooks (e.g. `useDrag`, `useSwipe`, or other name)?

PR #2279, opened 2023-04-07
feat(react): add toast component

takurinton:
#2192

@atomiks 

supports

- [x] Pauses closing on hover, focus and window blur
- [x] Supports hotkey to jump to toast viewport
- [x] Pressing esc when focused on toast closes it
- [x] Fully customizable toast components, positions, delay times, and animations
- [x] Global values can be specified from the Provider and reflected in individual toasts.
- [x] Globally defined values can be updated by individual toasts
  - If global and toast values do not exist, default values will be applied
- [ ] Fully follow the ARIA design pattern

not supported

- Supports closing via swipe gesture
- Toast with close button
- Default toast (nothing is rendered when render function is empty)

> Since this library is called "Floating UI", we can and should do more than just support elements that use anchor positioning — for example, a non-anchored floating Dialog or Drawer component are already possible to build using the existing interaction primitives, but a component like a Toast or anything draggable isn't despite them also being "floating" elements.

I agree with you. I was wanting a toast component that uses floating-ui.
I created the prototype as an experiment.

### usage

Brief description of how it is used (how it is intended to be used)

(edited)

```tsx
export const Main = () => (
  <ToastProvider placement="top-center">
    <Component />
  </ToastProvider>
);

const Component = () => {
  const {toast} = useToast();

  return (
    <div>
      <button
        onClick={() =>
          toast({
            placement: 'top',
            render: () => <div>Hello</div>,
          })
        }
      >
        Add Toast
      </button>
  </div>
  );
};
```

### worries

- I currently have this implementation in the visual test directory, is that correct?
- Should I create hooks and incorporate swipe gestures into floating-ui?
- I need assistance as my knowledge of accessibility is a little lacking.
- The API to manipulate the toast is not yet fully prepared here

### DIFF
--- a/packages/react/test/visual/components/Toast.tsx
+++ b/packages/react/test/visual/components/Toast.tsx
@@ -1,14 +1,12 @@
 import {
   ElementProps,
   ExtendedRefs,
-  inner,
   Placement,
   ReferenceType,
   useDismiss,
   useFloating,
   UseFloatingReturn,
   useInteractions,
-  useListNavigation,
   useTransitionStyles,
 } from '@floating-ui/react';
 import {
@@ -29,38 +27,21 @@ import {Button} from '../lib/Button';
 
 export type AppearanceTypes = 'error' | 'info' | 'success' | 'warning';
 export type Id = string;
-type Options = {
-  appearance?: AppearanceTypes;
-  autoDismiss?: boolean;
-  id?: Id;
-  onDismiss?: (id: Id) => void;
-  [key: string]: any;
-};
 
-type AddToast = (
-  content: ReactNode,
-  options?: Options,
-  callback?: (id: string) => void
-) => void;
-
-type ToastType = Options & {
-  appearance: AppearanceTypes;
-  content: ReactNode;
+type ToastType = {
   id: Id;
+  placement: Placement;
+  render: () => ReactNode;
 };
 type ToastsType = Array<ToastType>;
 
 type ContextType = {
   listRef: MutableRefObject<Array<HTMLElement | null>>;
-  index: number | null;
   toasts: ToastsType;
   dismiss: ElementProps;
-  setIndex: (index: number) => void;
-  listNavigation: ElementProps;
-  addToast: AddToast;
-  removeToast: (toastId: string) => void;
+  toast: (toast: Omit<ToastType, 'id'>) => void;
+  close: (toastId: string) => void;
   refs: ExtendedRefs<ReferenceType>;
-  autoDismissTimeout: number;
 } & UseFloatingReturn;
 
 export const Main = () => (
@@ -70,35 +51,26 @@ export const Main = () => (
 );
 
 const Component = () => {
-  const {addToast} = useToastsContext();
-
-  const handleClick = (text: string, options: Options) => () => {
-    addToast(text, options);
-  };
+  const {toast} = useToast();
 
   return (
     <ToastProvider>
       <>
         <h1 className="text-5xl font-bold mb-8">Toast</h1>
         <div className="grid place-items-center border border-slate-400 rounded lg:w-[40rem] h-[20rem] mb-4">
-          <Button onClick={handleClick('info', {appearance: 'info'})}>
+          <Button
+            onClick={() =>
+              toast({
+                placement: 'top',
+                render: () => (
+                  <div className="w-[300px] bg-white p-4 mb-4 rounded-lg shadow-lg flex">
+                    Hello
+                  </div>
+                ),
+              })
+            }
+          >
             Add Toast
-          </Button>
-          <Button
-            onClick={handleClick('info', {
-              appearance: 'info',
-              autoDismiss: true,
-            })}
-          >
-            Add Toast autoDismiss
-          </Button>
-          <Button
-            onClick={handleClick('error', {
-              appearance: 'error',
-              autoDismiss: true,
-            })}
-          >
-            Add Error Toast autoDismiss
           </Button>
         </div>
       </>
@@ -112,93 +84,42 @@ export function generateUEID() {
   return `${first}${second}`;
 }
 
-const getColor = (appearance?: AppearanceTypes) => {
-  switch (appearance) {
-    case 'info':
-      return '#CDF0FE';
-    case 'success':
-      return '#ECFCD3';
-    case 'error':
-      return '#FFE2E5';
-    case 'warning':
-      return '#FEFACF';
-    default:
-      return '#ffffff';
-  }
-};
-
-const getCountDownColor = (appearance?: AppearanceTypes) => {
-  switch (appearance) {
-    case 'info':
-      return '#00B4D8';
-    case 'success':
-      return '#5CB85C';
-    case 'error':
-      return '#D9534F';
-    case 'warning':
-      return '#F0AD4E';
-    default:
-      return '#ffffff';
-  }
-};
-
-export function useToasts({placement = 'bottom'}: {placement?: Placement}) {
-  const [index, setIndex] = useState<number | null>(0);
+export function useToasts({placement = 'top'}: {placement?: Placement}) {
   const [toasts, setToasts] = useState<ToastsType>([]);
   const listRef = useRef<Array<HTMLElement | null>>([]);
 
   const data = useFloating({
     placement,
     open: true,
-    middleware: [
-      inner({
-        listRef,
-        index: index ?? 0,
-      }),
-    ],
   });
 
-  const addToast: AddToast = useCallback((content, options) => {
-    const id = options?.id ?? generateUEID();
-    const newToast = {id, content, ...options} as ToastType;
-    setToasts((prev) => [...prev, newToast]);
+  const toast = useCallback(({placement, render}: Omit<ToastType, 'id'>) => {
+    setToasts((prev) => [
+      ...prev,
+      {
+        id: generateUEID(),
+        placement,
+        render,
+      },
+    ]);
   }, []);
 
-  const removeToast = useCallback((toastId: string) => {
+  const close = useCallback((toastId: string) => {
     setToasts((prev) => prev.filter((toast) => toast.id !== toastId));
   }, []);
 
   const dismiss = useDismiss(data.context);
-  const listNavigation = useListNavigation(data.context, {
-    listRef,
-    activeIndex: index,
-    onNavigate: setIndex,
-    enabled: false,
-  });
 
   return useMemo(
     () => ({
       listRef,
-      index,
       toasts,
       dismiss,
-      listNavigation,
-      setIndex,
-      addToast,
-      removeToast,
+      toast,
+      close,
       ...data,
     }),
-    [
-      listRef,
-      index,
-      toasts,
-      dismiss,
-      listNavigation,
-      setIndex,
-      addToast,
-      removeToast,
-      data,
-    ]
+    [listRef, toasts, dismiss, toast, close, data]
   );
 }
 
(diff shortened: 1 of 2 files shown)


atomiks:
What do you think of [Chakra UI](https://chakra-ui.com/docs/components/toast#custom-component) custom component API or [Radix's](https://www.radix-ui.com/docs/primitives/components/toast)?

@mihkeleidast, @FezVrasta any thoughts?

One thing we should do is make sure we only provide functional styles and no other (decorative) styling. i.e. just the position/stacking, to be in line with the rest of how the library works.

```js
const toast = useToast();
return (
  <button 
    onClick={() => {
      toast({
        placement: 'top',
        render: () => <div>Hello</div>
      });
    }} 
  />
);
```

packages/react/test/visual/components/Toast.tsx
```diff
@@ -0,0 +1,426 @@
...
+    case 'success':
+      return '#5CB85C';
+    case 'error':
+      return '#D9534F';
+    case 'warning':
+      return '#F0AD4E';
+    default:
+      return '#ffffff';
+  }
+};
```
atomiks, line 143: Let's leave all of this up to the consumer to stay in line with the lib's philosophy of being unstyled and only providing behavioral hooks

packages/react/test/visual/components/Toast.tsx
```diff
@@ -0,0 +1,426 @@
...
+  const listRef = useRef<Array<HTMLElement | null>>([]);
+
+  const data = useFloating({
+    placement,
+    open: true,
+    middleware: [
+      inner({
+        listRef,
+        index: index ?? 0,
+      }),
```
atomiks, line 157: Interesting usage 🤔 , but toasts are only positioned relative to the viewport, not anchored to a reference element, right?

packages/react/test/visual/components/Toast.tsx
```diff
@@ -0,0 +1,426 @@
...
+    setToasts((prev) => prev.filter((toast) => toast.id !== toastId));
+  }, []);
+
+  const dismiss = useDismiss(data.context);
+  const listNavigation = useListNavigation(data.context, {
+    listRef,
+    activeIndex: index,
+    onNavigate: setIndex,
+    enabled: false,
+  });
```
atomiks, line 177: Do toasts actually need list navigation? If there's multiple interactive ones, I guess it's somewhat useful, but I figured tabbing would suffice

takurinton:
Thanks @atomiks 
> What do you think of [Chakra UI's](https://chakra-ui.com/docs/components/toast#custom-component) custom component API, [Radix's](https://www.radix-ui.com/docs/primitives/components/toast), or [Sonner](https://sonner.emilkowal.ski/)?

I have used [react-toast-notification](https://jossmac.github.io/react-toast-notifications/)'s as a reference here.
However, development of this library has stopped, and we certainly feel it makes sense to refer to a library that is still active today.

> One thing we should do is make sure we only provide functional styles and no other (decorative) styling

I think this is correct, and a chakra-ui interface might be nice.

packages/react/test/visual/components/Toast.tsx
takurinton: Sorry! I was mistaken about this!
It is indeed unnecessary. No anchor is needed this time.

atomiks:
I just realized an absolute flexbox column won't work with transitions to adjust toast positions (no layout animation, without `framer-motion` at least)... so `translateY(...)` to adjust the position of toasts in a stack as toasts are removed from the stack is required. I imagine the logic would be: 

- Enter: check the height of all toasts in the stack to add it to the end
- Leave: calculate the leaving toast's height, then propagate the change to all other entered toasts as necessary

FezVrasta:
Depends, if you make the toast disappear by first making it invisible and then reducing its height until it reaches 0 you can use flexbox

takurinton:
@FezVrasta 

Thanks comment.
One question I have is that the current toast has the following structure, with `useFloating` in the `<ul>` section.

```html
<ul role="region" aria-live="polite">
  <li role="status" aria-atomic="true">
     Toast content
  </li>
</ul>
```

So I can't manipulate open, is this structure wrong? Should I manage `<li>` with `useFloating`? 
Also, is it possible to operate open or close on a listRef like the current one that holds `<li>` refs?

packages/react/test/visual/components/Toast.tsx
```diff
@@ -0,0 +1,353 @@
...
+          ? '20px'
+          : 'auto',
+      right:
+        placements[1] === undefined
+          ? 0
+          : placements[1] === 'end'
+          ? '20px'
+          : 'auto',
+    };
+  }
```
takurinton, line 170: @atomiks 
I implemented to get the position of toast.
This doesn't look very smart, is it best?
I would be happy to use the floating-ui feature to specify the position relative to the viewport, but I can't think of anything. Any suggestions?

takurinton:
Currently, this is how it behaves locally for me.
The toast add and remove appears to be implemented correctly.

I feel that the style is a bit forced and a11y is not guaranteed in some areas, but I feel that it is generally good.
https://user-images.githubusercontent.com/50351271/230880202-4e7633ec-0bb3-4b9b-81a7-65339a470d1d.mov

There are several points that I would like to confirm.

- Should we be able to specify the position from a global context as in [Configuring toast globally](https://chakra-ui.com/docs/components/toast#configuring-toast-globally)?
- Is it sufficient to specify the same `placement` as that of the `placement` in floting-ui using the `toast` function?
- Do we need to attach useFloating to each individual toast? (I think this is necessary if you use useDismiss to close the toast)

takurinton:
Currently, this is how it behaves locally for me.
The toast add and remove appears to be implemented correctly.

I feel that the style is a bit forced and a11y is not guaranteed in some areas, but I feel that it is generally good.

https://user-images.githubusercontent.com/50351271/230882146-a5b565d4-39b6-44e7-b7bc-c64200620792.mov

<br />

@atomiks 
There are several points that I would like to confirm.

- Should we be able to specify the position from a global context as in [Configuring toast globally](https://chakra-ui.com/docs/components/toast#configuring-toast-globally)?
- Is it sufficient to specify the same `placement` as that of the `placement` in floting-ui using the `toast` function?
- Do we need to attach useFloating to each individual toast? (I think this is necessary if you use useDismiss to close the toast)

atomiks:
> Should we be able to specify the position from a global context as in [Configuring toast globally](https://chakra-ui.com/docs/components/toast#configuring-toast-globally)?

I imagine this would be useful yes

> Is it sufficient to specify the same placement as that of the placement in floting-ui using the toast function?

The `placement` strings should be all `bottom` and all `top` sides (e.g. `top`, `top-end`, `bottom-start`, etc). `right` and `left` sides don't apply since the toasts are always vertically stacked.

> Do we need to attach useFloating to each individual toast? (I think this is necessary if you use useDismiss to close the toast)

For transition styles to work properly yes, however I'm not sure if pressing `esc` should dismiss all of them simultaneously or the most recent (or old) one only.

---

Some notes on the API design:

- There shouldn't be any Tailwind styling, only inline styles for functional things (like positioning/stacking). The consumer should specify all decorative styles themselves externally
- Transitions should be external as well with full customizability - the `max-height` should start transitioning out as the same time as `opacity` but it seems there's a delay atm?

takurinton:
Thank you @atomiks !!!
I'll make some revisions regarding that.

I have some more questions:

- Should we also apply useFloating to the entire element (the `ul` node)?
  - I feel like it's unnecessary as long as we apply it to the `li` nodes, but what do you think?
  - I believe this depends on which toast you want to close as you say
- Should toast really be navigable with tabs and closed with esc? For example, [react-toastify](https://fkhadra.github.io/react-toastify/introduction/) only allows focusing on the close button, while [chakra-ui](https://chakra-ui.com/docs/components/toast#custom-component) doesn't seem to allow focusing at all. What do you think??

takurinton:
current location

You can use placement to specify the position. You can specify `top`,`top-start`,`top-end`,`bottom`,`bottom-start`,`bottom-end`.
Example

```tsx
<button 
  onClick={() => {
    toast({
      placement: 'top',
      render: () => <div>Hello</div>
    });
  }} 
/>
```

Added deley to the `toast` function argument to allow individual adjustment of the time until the toast closes.
Example

```tsx
<button 
  onClick={() => {
    toast({
      deley: 1000
      render: () => <div>Hello</div>
    });
  }} 
/>
```

Next, I made the transition fully customizable. This allows you to pass the same properties that you can pass to ``UseTransitionStylesProps``. This is due to the useTransitionStyle when specifying open and close animations in the toast component.
If you write a partial, only that will be applied, and the default values will be applied for the other parts.
Example

```tsx
<button 
  onClick={() => {
    toast({
      transition: {
        duration: { open: 500, close: 1000 }
      },
      render: () => <div>Hello</div>
    });
  }} 
/>
```

The render function can specify an onClose callback, which can be used to close the toast at the implementer's timing.
Example

```tsx
<button 
  onClick={() => {
    toast({
      render: (onClose) => (
        <div style={{ display: 'flex' }}>
          <div>hello</div>
          <button onClick={onClose}>close</button>
        </div>
      )
    });
  }} 
/>
```

It would be useful to be able to specify these functions globally via Provider. I will work on that.
Also, clearAll, close, id, . . etc., it would be useful if APIs such as clearAll, close, id, ... etc. are provided, so I will consider their implementation as well.

takurinton:
Enabled `dismiss`, which only works when focusing on toast.

packages/react/test/visual/components/Toast.tsx
```diff
@@ -0,0 +1,392 @@
...
+    };
+
+    const {start, stop} = useTimeout(closeToast, delay);
+
+    return isMounted ? (
+      <FloatingFocusManager
+        context={context}
+        modal={false}
+        order={['floating']}
+        initialFocus={-1}
```
takurinton, line 352: When toast is displayed, the focus is on toast by default.
Then, when you press `esc`, the newest toast will close automatically, and then the focus is removed so that no toast will close when you press `esc`.
Therefore, the toast closes only when the focus is intentionally set to floating element.

takurinton:
The `closeAll` function has been created. This allows you to close all toasts.
Example

```tsx
const Component = () => {
  const {closeAll} = useToast();

  return <Button onClick={closeAll}>Close All</Button>;
};
```

Next, the `render` callback in the toast function can now accept an id.
This is useful when you want to do something with an id. The following can do the same thing.
Example

```tsx
const Component = () => {
  const {close} = useToast();

  return (
    <Button
      onClick={() =>
        toast({
          render: (id, onClose) => (
            <div className="w-[300px] bg-white p-4 mb-4 rounded-lg shadow-lg">
              Hello 
              {/* same behavior */}
              <button onClick={onClose}>close using onClose</button>
              <button onClick={() => close(id)}>close using close function</button>
            </div>
          ),
        })
      }
    >
    Add Toast
  </Button>
};
```

takurinton:
Added `autoClose` prop.
If this is set to false, the toast will not close automatically. default true.
Example

```tsx
<button 
  onClick={() => {
    toast({
      autoClose: false,
      render: () => <div>Hello</div>
    });
  }} 
/>
```

atomiks:
@takurinton sorry for the delay looking into this and thank you so much for working on it. I will take a proper look and review soon

atomiks:
I've had the chance to dive deeper into this, and I really appreciate the effort you've put into it. However, there are quite a few aspects that need to be adjusted to align better with the library's goals (and I'm personally not even sure how this will work — it needs some more exploration), but I believe your contribution has potential.

I won't be able to merge it at the moment, but feel free to continue developing this as an external package for others to use in the meantime. Integrating it seamlessly with the other hooks and components in the library is challenging in terms of how it will be designed, and could take some time to figure out the best API, but I will work on it eventually.

Thank you :))

takurinton:
@atomiks 
Thanks for looking this!! And sorry for the many things I have not done.
I look forward to a future implementation of floating-ui with absolute positioning relative to the viewport!
Thanks again!

atomiks:
Currently closing old unmerged PRs :)

PR #2280, opened 2023-04-07
fix(getClippingRect): correct fixed ancestors logic

### LINKED ISSUE #2278: The hide middleware hides the floating element when the referenced element has fixed position
**Describe the bug**
The [hide middleware](https://floating-ui.com/docs/hide) hides the floating element when the referenced element has `position: fixed;` and it's inside an element with `overflow: hidden;`.

**To Reproduce**

https://codesandbox.io/s/upbeat-mclaren-s9q0n0

Note that button 1 with fixed position does not display the tooltip, but button 2 with absolute position does.

**Expected behavior**
The tooltip must be displayed as long as the referenced element is also visible.

**Screenshots**
image

A real example to exemplify our use case:
real-example

**Context:**
 - OS: OS: macOS Monterey 12.1
 - Browser: Google Chrome
 - Version: 112.0.5615.49

**Additional context**

atomiks:
Fixes #2278


packages/dom/src/utils/getClippingRect.ts
```diff
@@ -105,26 +105,25 @@ function getClippingElementAncestors(
     const computedStyle = getComputedStyle(currentNode);
     const containingBlock = isContainingBlock(currentNode);
 
-    const shouldIgnoreCurrentNode = computedStyle.position === 'fixed';
-    if (shouldIgnoreCurrentNode) {
+    if (computedStyle.position === 'fixed') {
       currentContainingBlockComputedStyle = null;
+    }
```
atomiks, line 110: Instead of skipping the whole `else` branch as before, unsetting this variable fixes both #2167 and #2278

PR #2294, opened 2023-04-19
fix(FloatingList): properly update index even when there is only one item

chuganzy:
This PR aims to fix a bug where `useListItem` returns index set to `-1` when there is only one list item, and add a test to verify that.

It 1. creates an entry when `register` is called with node as key and `null` as value initially, and 2. `useLayoutEffect` updates each index respecting the node's position. However, when there is only one item there it skips 2.

Although I agree running sort etc in this case might be redundant, I though doing the same flow regardless of the size is simpler and efficient enough.


chuganzy:
Wow that was quick! Thank you!

PR #2297, opened 2023-04-21
fix(vue): add deep watch option to address @vue/compat warnings

kilmc:
My team is in the process of migrating our apps from Vue2 to Vue3 and we have been using `@vue/compat` to give us insights. In the process we discovered the following warning from `@vue/compat`

```
[Vue warn]: (deprecation WATCH_ARRAY) "watch" option or vm.$watch on an array value will no longer trigger on array mutation unless the "deep" option is specified.
If current usage is intended, you can disable the compat behavior and suppress this warning with:
```

We traced the source to be within the `@floating-ui/vue` package. This PR adds `deep: true` to the watch options object to address the warnings.

I tried to follow the conventions of the repo as best I could but if I messed anything up I'm happy to address it. I also didn't add tests as it seemed that the Vue ones are aimed at visual testing and this should have no affect but if those are needed I'm happy to add them.

### DIFF
--- a/packages/vue/src/useFloating.ts
+++ b/packages/vue/src/useFloating.ts
@@ -102,9 +102,13 @@ export function useFloating<T extends ReferenceElement = ReferenceElement>(
   }
 
   watch([middlewareOption, placementOption, strategyOption], update, {
+    deep: true,
     flush: 'sync',
   });
-  watch([referenceElement, floatingElement], attach, {flush: 'sync'});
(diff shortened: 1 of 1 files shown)


atomiks:
The array doesn't get mutated though? Shouldn't you just ignore the warning

kilmc:
Ah, I mis-understood how it was being used. I'll close then and suppress it on my end.

PR #2298, opened 2023-04-21
fix(FocusGuard): aria-hidden must not be focusable

chuganzy:
This PR aims to solve an accessibility issue in `FocusGuard` as titled.

References:
* https://dequeuniversity.com/rules/axe/4.7/aria-hidden-focus
* https://accessibilityinsights.io/info-examples/web/aria-hidden-focus/
* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

The intention is slightly different, but Chakra's [VisuallyHidden](https://chakra-ui.com/docs/components/visually-hidden) does not have `aria-hidden` set, and I think this can be also the case here.

### DIFF
--- a/packages/react/src/components/FocusGuard.tsx
+++ b/packages/react/src/components/FocusGuard.tsx
@@ -59,7 +59,6 @@ export const FocusGuard = React.forwardRef<
       tabIndex={0}
       // Role is only for VoiceOver
       role={role}
-      aria-hidden={role ? undefined : true}
       data-floating-ui-focus-guard=""
       style={HIDDEN_STYLES}
     />


atomiks:
The focus guards are special and don't abide by those rules

chuganzy:
Thanks for the comment and that is interesting. On the other hand, the current FocusGuard does not let axe pass.. Do you think it's ok to exclude it?

atomiks:
Yes it should be ignored, as the focus guards are only used to move focus somewhere else, not keep focus placed on them where `aria-hidden` would be a problem.

chuganzy:
Got it. Thank you for the input! I will close this soberly.

PR #2300, opened 2023-04-23
feat!: React v2 + Vue v1

atomiks:
This bumps `@floating-ui/react-dom` to v2 with various API improvements (supersedes #2204)

## Changes

- `x` and `y` default to `0` (React + Vue)
- Remove deprecated  `reference`, `floating`, `positionReference` top-level ref setters (React)
- Require `whileElementsMounted` to return a cleanup function in types, and renamed some other types (React + Vue)

---

- Add ability to sync external elements (React)

Instead of using a layout effect separately, you can synchronize external elements directly in the hook:

```js
useFloating({
  // Both of these are optional.
  // You can mix and match `refs.set{Name}` with external syncing
  elements: {
    reference: referenceElement,
    floating: floatingElement,
  }
});
```

---

- Add object with positioning styles available for you (React + Vue)

This covers most use cases with good defaults to help with simplifying the API, while the raw coordinates are still available for full customization:

```js
const {refs, floatingStyles} = useFloating();
return <div ref={refs.setFloating} style={floatingStyles} />;
```

The `transform` option will use `top/left` instead:

```js
useFloating({transform: false});
```

`willChange` is added for DPRs of >=1.5


FezVrasta (approved):

packages/vue/src/types.ts
```diff
@@ -94,6 +94,12 @@ export type UseFloatingOptions<T extends ReferenceElement = ReferenceElement> =
      * @default undefined
      */
     middleware?: MaybeReadonlyRef<Middleware[] | undefined>;
+    /**
+     * Whether to use `transform` instead of `top` and `left` styles to
+     * position the floating element (`floatingStyles`).
+     * @default true
+     */
+    transform?: boolean;
```
lozinsky, line 102: I think we should use `MaybeReadonlyRef<boolean | undefined>` type here.

packages/vue/src/useFloating.ts
```diff
@@ -73,7 +104,6 @@ export function useFloating<T extends ReferenceElement = ReferenceElement>(
   function cleanup() {
     if (typeof whileElementsMountedCleanup === 'function') {
       whileElementsMountedCleanup();
-      whileElementsMountedCleanup = undefined;
```
lozinsky, line 76: Maybe we should leave this? If we do not clear `whileElementsMountedCleanup`, it can be called again.

packages/vue/src/useFloating.ts
```diff
@@ -40,16 +42,45 @@ export function useFloating<T extends ReferenceElement = ReferenceElement>(
   const middlewareOption = computed(() => unref(options.middleware));
   const placementOption = computed(() => unref(options.placement) ?? 'bottom');
   const strategyOption = computed(() => unref(options.strategy) ?? 'absolute');
+  const transformOption = computed(() => unref(options.transform) ?? true);
   const referenceElement = computed(() => unwrapElement(reference.value));
   const floatingElement = computed(() => unwrapElement(floating.value));
-  const x = ref<number | null>(null);
-  const y = ref<number | null>(null);
+  const x = ref(0);
+  const y = ref(0);
```
lozinsky, line 49: Should we update the `UseFloatingReturn` type?

packages/vue/src/useFloating.ts
```diff
@@ -40,16 +42,45 @@ export function useFloating<T extends ReferenceElement = ReferenceElement>(
...
   const strategy = ref(strategyOption.value);
   const placement = ref(placementOption.value);
   const middlewareData = shallowRef<MiddlewareData>({});
   const isPositioned = ref(false);
+  const floatingStyles = computed(() => {
+    const initialStyles = {
+      position: strategyOption.value,
```
lozinsky, line 56: Maybe it should be `strategy.value`?

lozinsky: `strategy` syncs with `computePosition().strategy` instead of `strategyOption`.

`strategy` fallbacks to `'absolute'` initially, the same as `strategyOption`.

atomiks: Oh sorry, I missed where `strategy` was actually coming from

packages/vue/src/useFloating.ts
```diff
@@ -118,6 +149,7 @@ export function useFloating<T extends ReferenceElement = ReferenceElement>(
     placement: shallowReadonly(placement),
     middlewareData: shallowReadonly(middlewareData),
     isPositioned: shallowReadonly(isPositioned),
+    floatingStyles: shallowReadonly(floatingStyles),
```
lozinsky, line 152: I think that `shallowReadonly` is unnecessary here because `floatingStyles` is computed ref unlike the rest. Computed refs are readonly by default with the exception of writable computed refs 😀

lozinsky (approved):

atomiks:
I need to make a version indicator on the website for the packages, then I will merge/release these changes. I don't think I'll keep old docs around, though.

PR #2334, opened 2023-05-20
fix(dom): change userAgent checks

### LINKED ISSUE #2333: Positioning issue when using fixed strategy within a new containing block (`filter: drop-shadow()`)
When using `strategy: 'fixed'`, there is an issue with positioning when within a new containing block created by `filter`.

**To Reproduce**
- @floating-ui/dom: https://codesandbox.io/s/floating-ui-fixed-strategy-bug-f4hey0?file=/src/index.js

**Expected behavior**
Should be positioned correctly.

**Screenshots**
CleanShot 2023-05-17 at 21 08 19@2x

**Context:**
Looks like it's only an issue in Chrome.

**Additional context**
See https://github.com/radix-ui/primitives/issues/2018#issuecomment-1551243712

atomiks:
Would really prefer these checks could be feature-detected instead to ensure no problems in the future...

@benoitgrelard Fixes #2333


packages/dom/src/utils/is.ts
```diff
@@ -69,21 +67,9 @@ export function isContainingBlock(element: Element): boolean {
...
-  // this can fail in various ways, making the userAgent check the most
-  // reliable:
-  // • Always-visible scrollbar or not
-  // • Width of <html>
-
-  // Is Safari.
-  return /^((?!chrome|android).)*safari/i.test(getUAString());
+export function isSafari(): boolean {
+  if (typeof CSS === 'undefined' || !CSS.supports) return false;
+  return CSS.supports('-webkit-backdrop-filter', 'none');
```
atomiks, line 72: This seems to be a decent way to detect WebKit/Safari browsers, as both FF and Chrome don't support the `-webkit` prefix, avoiding the `userAgent` check. There's a chance Safari removes the `-webkit-` prefix support in the future, but the same can be said for the things being checked for to begin with. Neither one is **critical** necessarily, but I really wish they could be feature detected.

PR #2335, opened 2023-05-20
fix(autoUpdate): prevent ResizeObserver loop with size

### LINKED ISSUE #1740: ResizeObserver loop limit exceeded
@atomiks 

> You could disable the elementResize option and update manually when content changes.

Unfortunately I can't do this. I'm not changing the content, it's just due to the repositioning of the same tooltip itself

> I've never seen that error when it's near the edge of the viewport. Does it happen in incognito mode? Seems like extensions could interfere based on the SO thread?

I did my best to replicate the issue in codesandbox: https://codesandbox.io/s/elated-sunset-1ugji2?file=/src/App.tsx

See the video with the steps:

https://user-images.githubusercontent.com/9648559/172205861-d11354db-01dd-48f3-af33-28e804b38f09.mov

FYI: I had to add this to correctly track the "loop limit exceeded":
```
  useEffect(() => {
    window.onerror = console.error;
  }, []);
```

_Originally posted by @csantos1113 in https://github.com/floating-ui/floating-ui/issues/1739#issuecomment-1147656061_

atomiks:
@npearson72 can you let me know if this change also fixes the issue, it's less hackish

Fixes #1740

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -147,9 +147,22 @@ export function autoUpdate(
   const cleanupIo =
     referenceEl && layoutShift ? observeMove(referenceEl, update) : null;
 
+  let reobserveFrame = -1;
   let resizeObserver: ResizeObserver | null = null;
+
   if (elementResize) {
-    resizeObserver = new ResizeObserver(update);
+    resizeObserver = new ResizeObserver((entries) => {
+      if (entries[0].target === reference) {
+        // Prevent update loops when using the `size` middleware.
+        // https://github.com/floating-ui/floating-ui/issues/1740
+        resizeObserver && resizeObserver.unobserve(floating);
+        reobserveFrame = requestAnimationFrame(() => {
+          resizeObserver && resizeObserver.observe(floating);
+        });
+      }
+      update();
+    });
+
     if (referenceEl && !animationFrame) {
       resizeObserver.observe(referenceEl);
     }
@@ -191,6 +204,7 @@ export function autoUpdate(
     cleanupIo && cleanupIo();
     resizeObserver && resizeObserver.disconnect();
     resizeObserver = null;
+    cancelAnimationFrame(reobserveFrame);
 
     if (animationFrame) {
       cancelAnimationFrame(frameId);
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -205,7 +205,6 @@ export function autoUpdate(
     cleanupIo && cleanupIo();
     resizeObserver && resizeObserver.disconnect();
     resizeObserver = null;
-    cancelAnimationFrame(reobserveFrame);
 
     if (animationFrame) {
       cancelAnimationFrame(frameId);
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -156,6 +156,7 @@ export function autoUpdate(
         // Prevent update loops when using the `size` middleware.
         // https://github.com/floating-ui/floating-ui/issues/1740
         resizeObserver.unobserve(floating);
+        cancelAnimationFrame(reobserveFrame);
         reobserveFrame = requestAnimationFrame(() => {
           resizeObserver && resizeObserver.observe(floating);
         });
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -151,11 +151,11 @@ export function autoUpdate(
   let resizeObserver: ResizeObserver | null = null;
 
   if (elementResize) {
-    resizeObserver = new ResizeObserver((entries) => {
-      if (entries[0].target === reference) {
+    resizeObserver = new ResizeObserver(([firstEntry]) => {
+      if (firstEntry && firstEntry.target === referenceEl && resizeObserver) {
         // Prevent update loops when using the `size` middleware.
         // https://github.com/floating-ui/floating-ui/issues/1740
-        resizeObserver && resizeObserver.unobserve(floating);
+        resizeObserver.unobserve(floating);
         reobserveFrame = requestAnimationFrame(() => {
           resizeObserver && resizeObserver.observe(floating);
         });


npearson72:
@atomiks It appears to be happening less frequently, but unfortunately I am still seeing the error. This fix does however fix the other issue I mentioned here: https://github.com/floating-ui/floating-ui/issues/1740#issuecomment-1553651491

>  Interestingly it also seems to have resolved a UI quirk I was seeing with flip firing off in unpredictable ways in React when my component is first mounted

It's tricky because it is intermittent. I'll work on creating a codepen to reproduce this (if possible), but perhaps for now this issue should be marked as "unable to reproduce"?

In the meanwhile, as I mentioned, this does fix that other issue of flip acting unpredictably (at least in my case). So perhaps it's worth merging this one anyway.

atomiks:
@npearson72 so does the error never happen with the previous hackish one mentioned in the issue? I know you said it's intermittent, but you mentioned you couldn't reproduce it with that technique. If so, I'll just add the DOM property hack.

npearson72:
Yes, in that first hackish suggestion I was unable to reproduce it. It's possible I didn't exhaust every case while testing, but I did bang on it quite a bit and no errors appeared.

atomiks:
@npearson72 I think the previous solution didn't actually work because the `delete` line was running before `ResizeObserver` even managed to update. I'm not sure how it was fixing anything 🥲  

I thought of a much cleaner potential solution, which is to only ever run `update()` once in a given frame, and block further updates within a frame. This should hopefully prevent the error in all cases - can you give it a try (latest commit)?

npearson72:
> @npearson72 I think the previous solution didn't actually work because the `delete` line was running before `ResizeObserver` even managed to update. I'm not sure how it was fixing anything 🥲
> 
> I thought of a much cleaner potential solution, which is to only ever run `update()` once in a given frame, and block further updates within a frame. This should hopefully prevent the error in all cases - can you give it a try (latest commit)?

Sweet! I'm heading out to dinner, but will test when I get in tonight.

npearson72:
I wish I could report the error went away, but sadly I'm still seeing it. Maybe less frequently... hard to tell because again, it's intermittent. I need to click to expand/collapse an accordion, switch locales a few times, click on the popover, resize the screen, etc.. There is no clear combination of these steps to reproduce it. Somewhere in the course of these interactions the error fires off.

Sorry, wish I had something better to report.

atomiks:
@npearson72 just to confirm, if you switch the `if (elementResize) {...}` block to `if (false) {...}` so it never runs, you don't see the error? Just need to confirm if it's actually this library's `ResizeObserver` causing the issue and not something else. Otherwise, I'm not really sure what to do tbh. Since it's benign, you can just ignore it anyway...

npearson72:
@atomiks Yes, I confirm. I actually commented out the observer and indeed, no errors appear:

```ts
  // let observer: ResizeObserver | null = null;
  // let hasObserverRunInFrame = false;

  // if (elementResize) {
  //   observer = new ResizeObserver(() => {
  //     // Prevent update loops
  //     // https://github.com/floating-ui/floating-ui/issues/1740
  //     if (!hasObserverRunInFrame) {
  //       hasObserverRunInFrame = true;
  //       requestAnimationFrame(() => {
  //         hasObserverRunInFrame = false;
  //       });
  //       update();
  //     }
  //   });
  //   isElement(reference) && !animationFrame && observer.observe(reference);
  //   if (!isElement(reference) && reference.contextElement && !animationFrame) {
  //     observer.observe(reference.contextElement);
  //   }
  //   observer.observe(floating);
  // }

  let frameId: number;
  let prevRefRect = animationFrame ? getBoundingClientRect(reference) : null;

  if (animationFrame) {
    frameLoop();
  }

  function frameLoop() {
    const nextRefRect = getBoundingClientRect(reference);

    if (
      prevRefRect &&
      (nextRefRect.x !== prevRefRect.x ||
        nextRefRect.y !== prevRefRect.y ||
        nextRefRect.width !== prevRefRect.width ||
        nextRefRect.height !== prevRefRect.height)
    ) {
      update();
    }

    prevRefRect = nextRefRect;
    frameId = requestAnimationFrame(frameLoop);
  }

  update();

  return () => {
    ancestors.forEach((ancestor) => {
      ancestorScroll && ancestor.removeEventListener('scroll', update);
      ancestorResize && ancestor.removeEventListener('resize', update);
    });

    // observer?.disconnect();
    // observer = null;

    if (animationFrame) {
      cancelAnimationFrame(frameId);
    }
  };
```

This also re-introduces the strange flip/layout issue on first mount. The initial render of the popover is below the target element (rather than above with the "top" setting), but the arrow is in on the bottom. Resizing the window causes the popover to adjust into the proper position, but initially the layout is incorrect.

Each variation you've tried so far has fixed this layout issue, but the `ResizeObserver loop limit exceeded` error was present.

Also, perhaps my previous comments of how I go about reproducing the error lead you to think it may be due to some other library, etc., but I only noticed this error appear once I added a new feature that uses a popover and that library (Mantine) uses floating-ui as the engine for this behavior.

Again, sorry about this, I wish I had better news to report.

What are your thoughts on wrapping in a try/catch and swallowing the error? Or do you think it's to heavy-handed? I could also bring up that option at the Mantine level.

atomiks:
It looks like Mantine doesn't change the floating element size which might be why nothing I've done fixes the issue and you're not experiencing a problem with `size`, but something else:

https://github.com/mantinedev/mantine/blob/167dd7c5a9d5c6dbd1b64c456c03a6da8627084a/src/mantine-core/src/Popover/use-popover.ts#L83-L85C20

Mantine has a lot of custom logic around Floating UI, especially effects etc, so I think something is going wrong there.

Importantly, does the popover specify fixed width, or `width: max-content`? When using layout, and the popover is near the edge of the document, without a specified width, it can resize and result in loops. Also the popover must have `top: 0; left: 0` initially, which I'm not sure if Mantine is doing currently.

npearson72:
@rtivital ^

npearson72:
@atomiks this is helpful information. Regarding the width, I'm actually using `fit-content` because I'm displaying a variable width image for desktop vs mobile sizes.

> As far as I know it can't be swallowed this way, only `window.onerror` which can't be done in a library.

👍 

Thanks again for looking into this. Hoping @rtivital might be willing to make some adjustments in the popover component.

Cheers!

rtivital:
Hi @atomiks, since the PR was merged, I'm not really sure whether I should do something on Mantine side besides dependencies update. Can you please explain what I should do with these most recent changes?

atomiks:
@rtivital AFAIK, there are 2 cases where this error can occur:

- **What this PR fixes**: `size()` middleware + reference element resizes - no need to do anything in Mantine
- **Layout positioning with resizable width**: If you're not using `transform` to position the floating element, then layout can affect the floating element, causing it to resize and throw a loop error at the boundary edge. To fix it, the floating element must have `width: max-content` or a fixed width like `250px` specified. However, if you use the new `floatingStyles` object in the latest version, it defaults to `transform` styles and avoids that error without needing to specify a width at all, since layout doesn't get affected. That would be the easiest option in case the user overrides the width as `fit-content` or something.

There might be other cases but I'm not aware of them right now

PR #2347, opened 2023-05-26
Change `vue-demi` version range and remove `peerDependencies` and `peerDependenciesMeta`

lozinsky:
Due to the `@vue/composition-api` peer dependency, Nuxt 3 users cannot install `@floating-ui/vue` without the `--force` flag. The installation fails with the error:

```sh
npm ERR! Conflicting peer dependency: vue@2.6.14
npm ERR! node_modules/vue
npm ERR!   peer vue@">= 2.5 < 2.7" from @vue/composition-api@1.7.1
npm ERR!   node_modules/@vue/composition-api
npm ERR!     peerOptional @vue/composition-api@"^1.0.0-rc.1" from @floating-ui/vue@1.0.0
npm ERR!     node_modules/@floating-ui/vue
npm ERR!       @floating-ui/vue@"*" from the root project
```

Context: https://github.com/floating-ui/floating-ui/discussions/2339

This configuration of `peerDependencies` is recommended by [`vue-demi`](https://github.com/vueuse/vue-demi#usage). But [`@vueuse/core`](https://github.com/vueuse/vueuse/blob/b7b985b198d9e72136f65fae39b2125cd3abef5b/packages/core/package.json#L41-L46) doesn't follow this recommendation. I have found couple issues about Yarn warnings like this https://github.com/vueuse/vueuse/issues/2781.

I'm not sure that removing of `peerDependencies` and `peerDependenciesMeta` is the right way, but it will allow Nuxt 3 users to install `@floating-ui/vue`.

Do you have any thoughts on this?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21516,9 +21516,9 @@
       }
     },
     "node_modules/vue-demi": {
-      "version": "0.13.11",
-      "resolved": "https://registry.npmjs.org/vue-demi/-/vue-demi-0.13.11.tgz",
-      "integrity": "sha512-IR8HoEEGM65YY3ZJYAjMlKygDQn25D5ajNFNoKh9RSDMQtlzCxtfQjdQgv9jjK+m3377SsJXY8ysq8kLCZL25A==",
+      "version": "0.14.5",
+      "resolved": "https://registry.npmjs.org/vue-demi/-/vue-demi-0.14.5.tgz",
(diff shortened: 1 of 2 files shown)


atomiks:
Following what `@vueuse/core` does seems good to me. That includes the semver range for `vue-demi` to prevent it from being installed multiple times due to its `v0` status?

lozinsky:
Sure, updated.

atomiks:
I think `>=0.13.0` might be better based on npm downloads to reduce dupes for people? (Since we don't need `0.14.x` features, presumably?)

lozinsky:
I agree, we don't need the features of `0.14.x` now.

PR #2348, opened 2023-05-27
[wip] feat: obstacles

### LINKED ISSUE #1440: Support boundaries with “holes” aka detectOverflow relative to arbitrary elements

atomiks:
Closes #1440

As far as I can tell, this feature is most useful when another fixed / absolutely-positioned element is _also_ on the top-layer, or the floating element is "sharing" the layer it's on with something else. For example, you want to avoid a fixed navbar and not let it go under or over it, but instead avoid a collision like a boundary.

This is rudimentarily already possible using an option like `padding: {top: obstacleRect.height}` in `flip`/`shift`, etc, but this is cleaner and works better.

---

**The hardcoded `document.querySelectorAll` is temporary**. Each middleware will get an option like this, for tree-shaking purposes:

```js
flip({
  experimental_obstacles: experimental_avoidObstacles(
    Array.from(
      document.querySelectorAll('[data-obstacle]')
    ).map(el => el.getBoundingClientRect())
  ),
});
```

---

Needs to be added to:

- `size`
- `autoPlacement`

---

The algorithm is not perfect and struggles when there are multiple intersecting obstacles.

As mentioned, it's not ideal when you want multiple floating elements to avoid each other in a nice layout, as this technique technique is completely uncoordinated - the floating elements positioned first take full precedence instead of being nicely grouped. This behavior can still useful in some cases though it seems.

### DIFF
--- a/packages/core/src/detectIntersections.ts
+++ b/packages/core/src/detectIntersections.ts
@@ -11,10 +11,10 @@ import {rectToClientRect} from './utils/rectToClientRect';
 
 export interface Options {
   /**
-   * The collidable rects or area in which intersection will be checked.
+   * The obstacle rects or area in which intersection will be checked.
    * @default []
    */
-  collidables: Array<ClientRectObject>;
+  obstacles: Array<ClientRectObject>;
 
   /**
    * Virtual padding for the resolved intersection detection offsets.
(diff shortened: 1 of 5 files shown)


catdevnull:
Hi! Is there anything that's blocking this? Thanks for working on this!

atomiks:
It's half-baked and doesn't live up to the necessary standard to be added when compared to the rest of the library. It's also missing the second possibly crucial part which is [coordination](https://github.com/floating-ui/floating-ui/discussions/2719) for multiple floating elements to avoid each other in a nice layout.

A compromise is adding this avoidance detection as a feature only for `flip` since that's where it appears to be most useful, and is the smallest code addition-wise.

Issue #1842, opened 2022-08-06
[DOM] Support for :top-layer content, e.g. <dialog> and [popup]


Westbrook:
Content in a `<dialog>` that is opened with [`showModal()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal) or `showPopUp()` via the [`popup` attribute](https://open-ui.org/components/popup.research.explainer) will not actually have "offsetParents" or "clippingAncestors". How do you think it is best for the library to support positioning that content?

Particularly, if a reference element and a floating element are both bound in an element with a matrix transform, a la https://codesandbox.io/s/floating-ui-dom-template-forked-es3dfl?file=/src/index.js, do you foresee adding extra checks for this reality, or surfacing a flag so a use can opt-out of ancestor math or something else.

thepassle:
+1 on this, I was hoping to use floating-ui in combination with the native `<dialog>` element, for which `computePosition` works nicely, but middleware like `flip` and `shift` don't seem to work. It'd be excellent if this was supported as well 🙂

Westbrook:
I'm testing out this fairly naive middleware, based on the `offset` middleware, to manage this context:
```ts
import { Middleware, MiddlewareArguments } from "@floating-ui/dom";

export const topLayer = (): Middleware => ({
  name: 'topLayer',
  async fn(middlewareArguments: MiddlewareArguments) {
    const { x, y, elements: { reference, floating } } = middlewareArguments;
    let onTopLayer = false;
    const diffCoords = {
        x: 0,
        y: 0,
    }
    // browsers will throw when they do not support the following selectors, catch the errors.
    try {
        onTopLayer = onTopLayer || floating.matches(':open');
    // eslint-disable-next-line no-empty
    } catch (e) {}
    try {
        onTopLayer = onTopLayer || floating.matches(':modal');
    // eslint-disable-next-line no-empty
    } catch (e) {}

    if (onTopLayer) {
        // :top-layer content will be positioned from `window` get the reference's position from there, too.
        const rect = reference.getBoundingClientRect()
        diffCoords.x = rect.x;
        diffCoords.y = rect.y;
    }

    return {
      x: x + diffCoords.x,
      y: y + diffCoords.y,
      data: diffCoords,
    };
  },
});
```
This feels like a situation that will come up more and more as these specs solidify and expect. It feels like this should be core library functionality, or at least a core library middleware.

atomiks:
[...]

PR #2351, opened 2023-05-31
Draft topLayerOverTransforms plugin

Westbrook:
Beginning of the work to address #1842 

**Approach:**
_If_ a floating element is on the top layer the way we calculate its position will need to be revisited.
- coalesce "is top layer" with the following:
  - floating element is `:popover-open`
  - floating element is `:open` (only needed for "pre-release" versions of the `[popover]` API, IIUC)
  - floating element is `:modal`
    - currently only `:modal` is walking the ancestors, but it's possible that all of them should
- if their is a containing block, assume the floating element is being positioned over transforms (while this code works in my project, reading it right now made it hard to explain why, so I've got some naming refactors to do here)
- calculate the alternative offset based off of the transforms applied to the floating and reference elements (probably needs additional clarity in the variable names and comments)
- profit. 

Questions:
- is it OK to surface more of the utils code to the project? e.g. `packages/dom/src/utils/getContainingBlock.ts`
- `topLayerOverTransforms()` is very much a DOM plugin, both in that it can only happen with a DOM and that it requires some of the helpers from `@floating-ui/dom`. Does it make sense to make a "DOM middleware" or do you have thoughts on a better path here?
- have I missed any tooling for performance checking? It would be great to get a reading on the cost of this plugin, if only so that it could be further tuned one a test suite for it is ready.
- I'd like to also add tests for the `popover` attribute, should I add flags to Playwright or force > Chrome 114 or something else?
- I'm looking at adding tests for the following situation, let me know if you know of any I should add:
  - `position: fixed` inside of a “clipping ancestor” (this functionality regressed for me recently, so I need a tighter repro to prevent that in the future)
  - `<dialog>.openModal()` over nothing
  - `<dialog>.openModal()` over transforms
  - `<dialog>.openModal()` over `<dialog>.openModal()`
  - `<dialog>.openModal()` over `<[popover]>.showPopover()`
  - `<[popover]>.showPopover()` over nothing 
  - `<[popover]>.showPopover()` over transforms
  - `<[popover]>.showPopover()` over `<dialog>.openModal()`
  - `<[popover]>.showPopover()` over `<[popover]>.showPopover()`
- Testing in a React context is super foreign to me, so the sooner you can catch a weird approach in the test code the less times I'll do it in further testing. 🙏🏼 
- How do I update test images at CI time for the new tests I'll be adding?


atomiks:
Thank you for the PR!

Some things:

- Can you make sure you've run Prettier on all of the files? The formatting is off on most of them.
- Should `OverTransforms` be in the name? Shouldn't it be `topLayer`, with an option to handle transforms, or simply detect it automatically?

---

> is it OK to surface more of the utils code to the project? e.g. packages/dom/src/utils/getContainingBlock.ts

Yeah, I'm assuming you mean internally and not exported for package consumers?

> have I missed any tooling for performance checking? It would be great to get a reading on the cost of this plugin, if only so that it could be further tuned one a test suite for it is ready.

Perf testing isn't automated (would be cool if it were). You need to manually go to the `Performance` route right now and log `console.time('perf')` + `console.timeEnd('perf')` in the `computePosition` (core) function start and end, then run DevTools performance monitoring.

> I'd like to also add tests for the popover attribute, should I add flags to Playwright or force > Chrome 114 or something else?

Not familiar with this but sounds like that would be required

> How do I update test images at CI time for the new tests I'll be adding?

- Go to **Actions** in the repo
- **Update Visual Snapshots** workflow on the left sidebar
- **Run workflow** dropdown
- **Select your branch**
- **Run workflow** button

Westbrook:
> Should OverTransforms be in the name? Shouldn't it be topLayer, with an option to handle transforms, or simply detect it automatically?

This is a great question, but feels like it would be better answered by you than me. Top Layer content that doesn't live in a container that has been transformed does not need this plugin. Top Layer content that does, needs it. When I originally envisioned it as a plugin, `[popover]` ([spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover)) was still just coming together, and while `<dialog>.openModal()` was more widely available, the act of positioning a `<dialog>` is _slighly_ less common, so I thought it should be opt-in. However, it's possible that this concept needs to be part of the standard positioning workflow, no matter the added cost of doing so. To that end:

- would you like it to continue to be a middleware? or as `[popover]` is likely to see rapid uptake, should this be part of the main calculations?
- if it stays a middleware, it might not only be for "transform" (I'll do more testing), but it's only for specific cases, should those cases be baked into the name in some way?

atomiks:
So most top-layer content already works then? I thought it didn't work at all but it seems to work here: https://codesandbox.io/s/ancient-water-0y60lj?file=/src/App.tsx 

Isn't putting it inside a transformed container kind of a niche use-case? Maybe it'd be better as a separate package for others to install if so

Westbrook:
As for leveraging an external package... `transform` is the main way to make animations in the web, any content that may have or may be animated will likely have transforms on it making likely less niche than `contain: strict` et al, which already have direct support in the library. On top of that, the work to calculate this as a plugin benefits from access to the util methods already build and maintained in library, if supporting this context is forced into an external package the next request will be to export more of the lib for use there. I might be reading too deeply into things, but your note about "and not exported for package consumers?" sounds like you're not overly interested in doing so. Where would you stand on that with this info?

As for whether this effects middleware, which seems likely... I've not gotten to testing them in isolation. I've always approached this as a middleware that went at the end of other middle ware, so it's likely that beyond `transform` there are other concepts that disrupt the current calculation logic. If that were the case, it would seems even more appropriate to move this work into the main library.

I'll expand my efforts to clarify boundaries of the currently available API.

Westbrook:
Update: funny I referenced `contain` in that it does seem like it's any rule that will create a containing ancestor for the floating content that can effect this calculation. I'll have updated tests in the branch soon that outline this.

atomiks:
https://github.com/floating-ui/floating-ui/pull/2747

PR #2357, opened 2023-06-01
feat(dom): export getRectRelativeToOffsetParent function

layershifter:
This PR exports `getRectRelativeToOffsetParent()` function from `@floating-ui/dom`.

TL;DR Without it you can't compose `getElementRects()` in `platform`:
https://github.com/floating-ui/floating-ui/blob/b8990250568043b876e1c8fe42358fe337847ede/packages/dom/src/platform.ts#L20-L31

### Why?

We need this for a hack in https://github.com/microsoft/fluentui/pull/28045 that prevents jumps, see #26579 and https://github.com/microsoft/fluentui/issues/26579#issuecomment-1503562558.

- We need to set `position: fixed` for initial positioning so elements in React portals will not appear off screen
- This should be done after `getOffsetParent()` as `getOffsetParent()` will always return `null` when an element has `position: fixed`
- This should be done before `getRectRelativeToOffsetParent()`
- `getRectRelativeToOffsetParent()` is not exported 💥

### DIFF
--- a/packages/dom/src/index.ts
+++ b/packages/dom/src/index.ts
@@ -32,6 +32,7 @@ export const computePosition = (
 export {autoUpdate} from './autoUpdate';
 export {platform} from './platform';
 export {getOverflowAncestors} from './utils/getOverflowAncestors';
+export {getRectRelativeToOffsetParent} from './utils/getRectRelativeToOffsetParent';
 export {
   arrow,
   autoPlacement,


atomiks:
> We need to set position: fixed for initial positioning so elements in React portals will not appear off screen

Is this only to prevent scroll jumps due to focus? You can use an `isPositioned`-like value to avoid this problem and stick to `absolute`. Here is an example: https://floating-ui.com/docs/react#effects

layershifter:
> Is this only to prevent scroll jumps due to focus? You can use an `isPositioned`-like value to avoid this problem and stick to `absolute`. Here is an example: https://floating-ui.com/docs/react#effects

Nope, check the linked issue it contains an explainer of what is happening 😉

layershifter:
> So, is it possible to remove that CSS property on the portal container, to prevent measurements from creating potential scrollbars?

Unfortunately, it's not an option currently 😿

layershifter:
@atomiks is there a chance to get this merged? Or maybe you have some suggestions to a case that I described above?

atomiks:
Well, I'm hesitant to export new APIs if they don't need to be, especially internal ones that I might want to change

I think you can use `position: absolute; top: 0;` on the wrapper instead

layershifter:
> Well, I'm hesitant to export new APIs if they don't need to be, especially internal ones that I might want to change
> 
> I think you can use `position: absolute; top: 0; left: 0; width: 100%;` on the wrapper instead

Thanks for suggestion, haven't considered applying `left` & `top` 🤔 TIL 👍 

P.S. `right: 0` is also required to have a proper width of a container

Issue #2349, opened 2023-05-27
[React] Large values of FloatingArrow padding makes arrow and tooltip positions incorrect


EliteUser:
Hello!

Stuck with a problem with the tooltip arrow... again 😄 

Reproduction example - [reproduction](https://codesandbox.io/p/sandbox/tooltip-arrow-positioning-45hdkq?file=%2Fsrc%2FApp.tsx) 

As a continuation of [this](https://github.com/floating-ui/floating-ui/discussions/2307) discussion (which is unanswered 😢) - when padding value is large the arrow is not positioned correctly. 

E.g - for placement=left, if tooltip height is 36px, and padding value is also 36px - looks like this:

![image](https://github.com/floating-ui/floating-ui/assets/31800794/fd7cd23a-3882-4aaf-855b-2bf955b5612e)

But should look like this: 

![image](https://github.com/floating-ui/floating-ui/assets/31800794/aff508b5-0c31-4a0f-959d-0f165a3ef067)

The problem is that the padding is just a number that only takes into account one side, even though the documentation says - "stop 5px from the edges of the floating element". It doesn't take into account the placement of the tooltip, its size, or the size of the trigger element.

In my case - I want arrow to be centered on tooltip, if large padding value makes arrow "disconnected" from tooltip. So I think that some improvements in arrow middleware will help - like callback argument in offset mw, which can provide info about rects.

There are also a lot of other cases I tried to describe in the [discussion](https://github.com/floating-ui/floating-ui/discussions/2307), but I think the main thing is the padding

Any ideas how to fix it? 
Thx in advance!

atomiks:
What is the use case for such a large padding value? It's only really necessary to be as large as the `border-radius` causes it to overflow the corners. A value of `36` means it has to be _at least_ 36 pixels from one of the edges, causing that result

EliteUser:
I have following tooltip design

![image](https://github.com/floating-ui/floating-ui/assets/31800794/71d9ac13-19d0-45a6-95ac-152c6de150a8)

This is left-top placement. So here I have 5px border-radius and 14px offset from the top edge to the arrow (the padding I want to set). The arrow itself is 18x8px. So when tooltip will only have title (and in this case its height will be 36px - minimum height possible) the padding will be very large, even if it not-aligned-placement. I have these cases in reproduction. 
So by design, if tooltip is small, I want to center the arrow - set padding to another value, depending on the tooltip or trigger element size

atomiks:
[...]

PR #2359, opened 2023-06-02
feat: function options

### LINKED ISSUE #2353: Lack of type definition in react-native
Middleware in @floating-ui/react-native core has no type definition.

eg) ShiftOptions, OffsetOptions type

atomiks:
Part 1 of fixing #2349
Fixes #2353 

To match `offset`'s `Options` type, which allows you to derive option values from state, all middleware options can now accept a function that receives the state and returns the options:

```js
shift(options);
shift(state => options);
```


FezVrasta (approved):

atomiks:
Removed the reactive deps list option from `useFloating` and won't add any reactivity behavior (yet at least). The problem is that the deps array disables reactivity for non-function option middleware as well, so it becomes all-or-nothing.

The best option is likely a dep array for middleware, which makes it granular:

```js
offset(() => statefulValue, [statefulValue])
```

This works fine but adds some overhead for exports, and since this hasn't been a requested feature yet, I'm not going to add it right now.

Either way, there's two workarounds that already exist:

**Stateful workaround:**

```js
const offsetResult = offset(() => statefulValue);
const reactiveOffsetResult = {...offsetResult, options: [statefulValue]};
useFloating({
  middleware: [reactiveOffsetResult],
});
```

**Ref workaround (already on docs):**

```js
const statefulValueRef = useLatestRef(statefulValue);
offset(() => statefulValueRef.current);
useLayoutEffect(update, [statefulValue]);
```

I'll add the first one to the docs as well.

Issue #2047, opened 2022-12-19
Is there any way to get the event, in the onOpenChange function?


aromanchev:
The problem is that when onOpenChange is triggered, only the open/closed state of dropdown comes. Is there any way to get an event how dropdown opens(keyboard, mouse, etc) when onOpenChange is triggered?
```typescript
...
const { x, y, reference, floating, refs, strategy, context } = useFloating({
	...
	onOpenChange: (opened, event?) => {
          ...
	},
        ...
});
```

atomiks:
`context.dataRef.current.openEvent` contains the event that caused the floating element to open.

aromanchev:
@atomiks but if you try to open in first time you get undefined from context.dataRef.current.openEvent

atomiks:
When are you accessing it? You need to wait for the event to have fired for it to be written to the ref

alexandredepape:
Hi, i have the same issue, i'm having a debug point in my code:

```js
onOpenChange: (open) => {
            const event = context.dataRef.current.openEvent;
            debugger;
},
```

But when looking in the dev tools, the event is undefined even though the event got fired since the onOpenchange was called ?

PR #2362, opened 2023-06-04
feat(react): `onOpenChange` event param

atomiks:
There have been requests to access the event (https://github.com/floating-ui/floating-ui/issues/2047) inside `onOpenChange`. `dataRef.current.openEvent` does not currently get written before `onOpenChange` is called in all cases, and there's no `close` equivalent.

- `onOpenChange` is called without an event in `FloatingDelayGroup`, so it's optionally passed
-  `onOpenChange` isn't called if the user externally changes state, like `setIsOpen(false)`. In this case I think it makes sense to just react to whatever's needed inside the event where they did that anyway


alexandredepape:
what version of the package do I need ? I have     "@floating-ui/dom": "^1.4.3",
 and don't have this feature

PR #2373, opened 2023-06-12
feat(autoUpdate): `layoutShift` check

### LINKED ISSUE #2053: [DOM] Use `IntersectionObserver` in `autoUpdate`
In our usage of `floating-ui` we came across a scenario where the floating element's position isn't getting updated, because neither `ancestorScroll`, `ancestorResize`, nor `elementResize` would catch what's changing in the document. Enabling `animationFrame` fixes the issue, however we're concerned about the performance impact this would have.

Looking into the implementation of `autoUpdate` with `animationFrame: true`, I noticed that you're comparing the result of `getBoundingClientRect` on every frame.

I was curious if this approach ever came up on your radar: https://toruskit.com/blog/how-to-get-element-bounds-without-reflow/

Here's a proposal that I haven't tested yet, but just to illustrate what I think this would look like: https://github.com/idevelop/floating-ui/commit/7ab6b5dd356e94a3f2252a0ec78c8587d6e5ca5a

### LINKED ISSUE #2380: Floating UI in a sticky header doesn't stay with the reference element
**Describe the bug**

Floating UI in a sticky header doesn't stay with the reference element. `strategy: fixed` is a workaround, but that can cause scroll jank when the sticky element isn't stuck. Ideally i think this case should work out of the box? 

**To Reproduce**

https://codesandbox.io/s/floating-ui-react-dom-template-forked-4yvkfm?file=/src/App.js

1. Scroll the preview window

**Expected behavior**

Tooltip makes an effort to stay positioned to the reference element

**Screenshots**

**Context:**

**Additional context**

atomiks:
Closes #2053 
Fixes #2380


samthor:
@atomiks I'd appreciate a tiny link in a comment to my blog post on this approach. Glad it's helpful.

atomiks:
@samthor of course, thank you for the technique! I usually credit with a comment but wasn't sure the best way here with the lib's license. I haven't seen that technique used anywhere else so it'd be cool to amplify it for sure.

I reverse engineered your lib code down to a minimum and didn't see the blog post until now, and one thing I noticed after reading the post is I didn't add the `root` `ResizeObserver`, but it looks like it's necessary in certain cases? Is this when for example content is added, the reference rect relocates (and it will refresh), but then if it moves _again_ after, it won't be detected, due to the root size being stale?

packages/dom/src/autoUpdate.ts
```diff
@@ -25,13 +30,83 @@ export type Options = Partial<{
...
+      onMove();
+    }
+
+    if (!rect.width || !rect.height) {
+      return;
+    }
+
+    const insetTop = floor(top);
+    const insetLeft = floor(left);
+    const insetRight = floor(root.offsetWidth - (left + rect.width));
```
samthor, line 76: `root.offsetWidth` and height are going to change if the `documentElement` changes in size, e.g., because you resized the browser window.
That's why you need a `ResizeObserver` on it.
This would be different if you could use `calc` here, e.g., `calc(100% - ${left + rect.width})`, so that's basically taking the place of 100% there.

packages/dom/src/autoUpdate.ts
```diff
@@ -25,13 +30,84 @@ export type Options = Partial<{
...
+
+  /**
+   * Whether to update on every animation frame if necessary. Only use if you
+   * need to update the position in response to an animation using transforms.
    * @default false
    */
   animationFrame: boolean;
 }>;
 
+// https://samthor.au/2021/observing-dom/
```
samthor, line 47: Thanks, that's fine. I'm not trying to claim ownership of the technique, just nice to have some context for the next reader(s) :)

packages/dom/src/autoUpdate.ts
samthor: No, I don't think so. I don't quite understand, is this `observeMove` top level function already rerun on browser resize?

atomiks: There are 3 detection mechanisms already in place to perform an update, `observeMove` only needs to handle layout shift:

Each one is an event that calls the update function:

- `ancestorScroll` = scrolling containers and main body/window scroll
- `ancestorResize` = window resize
- `elementResize` = floating or reference element resize (`ResizeObserver`)
- `layoutShift` = this PR, was missing as the position needs to update if the reference moved on the screen not as the result of any of those other ones

samthor: Ah, you've distilled it to purely a shift, I understand now.

In my code I use it to also detect the other resize cases, hence my confusion.

Slapbox:
This is the most exciting new feature I've seen since `floating-ui` itself was announced! Thank you @samthor and @atomiks!

atomiks:
So there's an edge case with `IntersectionObserver` and that's if the reference element is inside a scrolling container and is partially clipped. If it's clipped then moves won't be detected in the clipped region in some cases. It does mostly fix the nested floating element with different scrolling container issues aside from that. It also seems to enter a loop under very specific conditions with nested scrolling containers, but doesn't freeze the page at least (works like rAF).

atomiks:
fyi this currently doesn't actually work in `v1.4.0` after I refactored out the 0 check lol 🤦 

The tests are passing because an update is occurring via the effect when the options change. I'll try to release a patch shortly.

PR #2384, opened 2023-06-18
fix(autoUpdate): `layoutShift` check

atomiks:
Tests passed for https://github.com/floating-ui/floating-ui/pull/2373 since the useEffect was causing an update when the options changed, but the last refactor prevented the option from actually having an effect.

I couldn't work out a way to prevent the loop without the check breaking afterwards in various ways, so I throttled it to 100ms while the reference is clipped/hidden from view to prevent the CPU from spiking

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -51,7 +51,7 @@ function observeMove(element: Element, onMove: () => void) {
     io = null;
   }
 
-  function refresh(skip = false) {
+  function refresh(skip = false, threshold = 1) {
     cleanup();
 
     const {left, top, width, height} = element.getBoundingClientRect();
@@ -71,16 +71,30 @@ function observeMove(element: Element, onMove: () => void) {
     const rootMargin = `${-insetTop}px ${-insetRight}px ${-insetBottom}px ${-insetLeft}px`;
 
     let isFirstUpdate = true;
+    let timeoutId: NodeJS.Timeout;
 
     io = new IntersectionObserver(
       (entries) => {
-        if (entries[0].intersectionRatio !== 1 && !isFirstUpdate) {
-          refresh();
+        clearTimeout(timeoutId);
+        const ratio = entries[0].intersectionRatio;
+
+        if (ratio !== threshold) {
+          if (!isFirstUpdate) {
+            return refresh();
+          }
+
+          if (ratio === 0) {
+            timeoutId = setTimeout(() => {
+              refresh(false, 1e-7);
+            }, 100);
+          } else {
+            refresh(false, ratio);
+          }
         }
 
         isFirstUpdate = false;
       },
-      {rootMargin, threshold: 1}
+      {rootMargin, threshold}
     );
 
     io.observe(element);
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -44,9 +44,12 @@ export type Options = Partial<{
 // https://samthor.au/2021/observing-dom/
 function observeMove(element: Element, onMove: () => void) {
   let io: IntersectionObserver | null = null;
+  let timeoutId: NodeJS.Timeout;
+
   const root = getDocumentElement(element);
 
   function cleanup() {
+    clearTimeout(timeoutId);
     io && io.disconnect();
     io = null;
   }
@@ -71,11 +74,9 @@ function observeMove(element: Element, onMove: () => void) {
     const rootMargin = `${-insetTop}px ${-insetRight}px ${-insetBottom}px ${-insetLeft}px`;
 
     let isFirstUpdate = true;
-    let timeoutId: NodeJS.Timeout;
 
     io = new IntersectionObserver(
       (entries) => {
-        clearTimeout(timeoutId);
         const ratio = entries[0].intersectionRatio;
 
         if (ratio !== threshold) {
--- a/packages/dom/test/functional/autoUpdate.test.ts
+++ b/packages/dom/test/functional/autoUpdate.test.ts
@@ -26,6 +26,7 @@ import {click} from './utils/click';
 ['none', 'move', 'insert', 'delete'].forEach((option) => {
   test(`layoutShift: ${option}`, async ({page}) => {
     await page.goto('http://localhost:1234/autoUpdate');
+    await click(page, `[data-testid="layoutShift-init"]`);
     await click(page, `[data-testid="layoutShift-${option}"]`);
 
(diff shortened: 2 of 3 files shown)


sgrishchenko:
Could you explain how behaves observer and what cause the cycle? I rolled back changes from this commit, ran test and they passed (except test with inline, it seems I have some problems with fonts, maybe because I run them on Ubuntu)

sgrishchenko:
Oh, I think I get the idea, when reference clipped or hidden by container, it can't have `intersectionRatio === 1`, so in that case you try to figure out what  `intersectionRatio` it actually have and refresh the threshold. But if reference is hidden, it have `intersectionRatio === 0`. So it still cause cycle, and while reference in that state, you throttle updates with 100ms interval. Also you wait while a small bite of reference is shown, and in that moment you hope that cycle will stop.
It is really clever :clap:.

sgrishchenko:
But I still have one question: do you really need hack with `isFirstUpdate`? I believe it was copied from [original article](https://samthor.au/2021/observing-dom/) to prevent cycle on observer initialization, but it seems now `ratio !== threshold` condition already prevents cycle on initialization.

sgrishchenko:
@samthor what do you think about removing `isFirstUpdate` in current implementation?

sgrishchenko:
@atomiks I tried to examine if `isFirstUpdate` does something useful or not, and it seems it does not 😊.
I will try to prove it. There are 3 branches of code (inside `ratio !== threshold` condition)
1) `!isFirstUpdate` -> refresh immediately
2) `ratio !== 0` -> refresh immediately
3) `ratio === 0` -> refresh with timeout
So here we can see that second update can happen only in 100ms interval when `ratio === 0` (or element is hidden). It can happen only if some part of reference is visible (or `ratio !== 0`). But as it was described above if `ratio !== 0` it leads to immediate refresh.
I also attached screencast + source code with example, that demonstrates this behavior. You may notice that in console there are some `!isFirstUpdate` triggers, but all of them are linked with `ratio !== threshold` and `ratio !== 0` conditions, so in that case observer will be re-created anyway.
[Screencast from 07-18-2023 10:51:58 AM.webm](https://github.com/floating-ui/floating-ui/assets/15995890/77658b6d-f19c-49ce-a3f3-764d3549cf08)
```
<style>
    .item {
        width: 50px;
        height: 50px;
        min-width: 50px;
        min-height: 50px;

        background: darkgreen;
        border: 1px solid black;
    }

    .highlight {
        position: absolute;

        outline: 1px solid coral;
    }

    .canvas {
        position: relative;
        width: 300px;
        height: 300px;
        overflow: hidden;
        background-color: bisque;
    }

    .canvas .item {
        position: absolute;
        min-width: 0;
        min-height: 0;
    }
</style>

<div class="canvas">
  <div class="item"></div>
</div>
<div>
  <label for="x">X:</label>
  <input id="x" type="number" value="0"/>
  <label for="y">Y:</label>
  <input id="y" type="number" value="0"/>
</div>
<div>
  <label for="width">Width:</label>
  <input id="width" type="number" value="50"/>
  <label for="height">Height:</label>
  <input id="height" type="number" value="50"/>
</div>

<script>
  function observeMove(element, onMove) {
    let io = null;
    let timeoutId;

    const root = window.document.documentElement;

    function cleanup() {
      clearTimeout(timeoutId);
      io && io.disconnect();
      io = null;
    }

    function refresh(skip = false, threshold = 1) {
      cleanup();

      const {left, top, width, height} = element.getBoundingClientRect();

      if (!skip) {
        onMove();
      }

      if (!width || !height) {
        return;
      }

      const insetTop = Math.floor(top);
      const insetRight = Math.floor(root.clientWidth - (left + width));
      const insetBottom = Math.floor(root.clientHeight - (top + height));
      const insetLeft = Math.floor(left);
      const rootMargin = `${-insetTop}px ${-insetRight}px ${-insetBottom}px ${-insetLeft}px`;
      console.log(rootMargin)

      const options = {
        rootMargin,
        threshold: Math.max(0, Math.min(1, threshold)) || 1,
      };

      let isFirstUpdate = true;

      function handleObserve(entries) {
        console.log("isFirstUpdate", isFirstUpdate)
        console.log("ratio", entries[0].intersectionRatio)
        console.log("threshold", threshold)
        const ratio = entries[0].intersectionRatio;

        if (ratio !== threshold) {
          if (!isFirstUpdate) {
            console.log("!isFirstUpdate")
            return refresh();
          }

          if (!ratio) {
            timeoutId = setTimeout(() => {
              refresh(false, 1e-7);
            }, 1000);
          } else {
            refresh(false, ratio);
          }
        }

        isFirstUpdate = false;
      }

      // Older browsers don't support a `document` as the root and will throw an
      // error.
      try {
        io = new IntersectionObserver(handleObserve, {
          ...options,
          // Handle <iframe>s
          root: root.ownerDocument,
        });
      } catch (e) {
        io = new IntersectionObserver(handleObserve, options);
      }

      io.observe(element);
    }

    refresh(true);

    return cleanup
  }

  function handlePositionChange(element, highlight) {
    const rect = element.getBoundingClientRect()

    const top = rect.top
    const left = rect.left
    const width = rect.width
    const height = rect.height

    highlight.style.top = top + 1
    highlight.style.left = left + 1
    highlight.style.width = width - 2
    highlight.style.height = height - 2
  }

  function init() {
    const movedItem = document.querySelector(".canvas .item")

    const highlight = document.createElement("div")
    highlight.classList.add("highlight")
    document.body.append(highlight)

    observeMove(movedItem, () => {
      handlePositionChange(movedItem, highlight)
    })

    document.getElementById("x").addEventListener("change", event => {
      movedItem.style.left = event.target.value
    })

    document.getElementById("y").addEventListener("change", event => {
      movedItem.style.top = event.target.value
    })

    document.getElementById("width").addEventListener("change", event => {
      movedItem.style.width = event.target.value
    })

    document.getElementById("height").addEventListener("change", event => {
      movedItem.style.height = event.target.value
    })
  }

  init()
</script>
```

sgrishchenko:
I could make PR with removing of `isFirstUpdate` if you don't mind.

sgrishchenko:
Oh, I missed one more scenario: `ratio === 0` does not always mean that element is hidden, it actually can be the case, when position of reference is changed radically. If I remove `isFirstUpdate` check, position of element will be updated only after 100ms timeout. In attached video I commented out `isFirstUpdate` check and increased timeout to demonstrate the effect. It seems it wold be better to keep this flag to provide more responsive positioning of floating element.
 [Screencast from 07-18-2023 11:36:59 AM.webm](https://github.com/floating-ui/floating-ui/assets/15995890/8a007765-9607-41ff-8f35-ea5594293625)

PR #2387, opened 2023-06-20
fix(useMergeRefs): refs could be undefined

shlroland:
The ref can also be undefined, and in this case. And it is equivalent to null. Therefore, the double equals (==) can also filter it.


atomiks:
As far as I know the React TS types require you to type the element in a ref as `null` not `undefined`

shlroland:
> As far as I know the React TS types require you to type the element in a ref as `null` not `undefined`

You are right. However, some implementations of third-party libraries do not comply with this requirement, and the provided ref may indeed be undefined. In such cases, the only way to eliminate the type error is to use type assertions.

PR #2390, opened 2023-06-20
fix(autoUpdate): avoid potential `IntersectionObserver` crash

### LINKED ISSUE #2389: Uncaught RangeError: Failed to construct 'IntersectionObserver': Threshold values must be numbers between 0 and 1
**Describe the bug**
image

antd-mobile popover import [floating-ui](https://github.com/floating-ui/floating-ui)

image
antd-mobile-5.31.1  code

image
too much error about ':Uncaught RangeError: Failed to construct 'IntersectionObserver': Threshold values must be numbers between 0 and 1' generated here

atomiks:
Fixes #2389

I don't know how this can happen [given the spec says `intersectionRatio` is always between `0.0` and `1.0`](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/intersectionRatio#value)  - maybe some kind of floating point error? - but to avoid any issues with bugged implementations, clamp the value between `0` and `1` and fallback to `1` if it's some other falsy value

@HuanEragon not sure if you can test this, but if you edit the file locally with this change the error shouldn't happen

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -2,7 +2,7 @@ import type {FloatingElement, ReferenceElement} from './types';
 import {getBoundingClientRect} from './utils/getBoundingClientRect';
 import {getDocumentElement} from './utils/getDocumentElement';
 import {getOverflowAncestors} from './utils/getOverflowAncestors';
-import {floor} from './utils/math';
+import {floor, max, min} from './utils/math';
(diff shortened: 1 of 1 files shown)


zombieJ:
thx for quick fix. When will this release?

atomiks:
Since it can't be replicated according to the issue, I will merge this - pretty certain it will prevent the crash

PR #2393, opened 2023-06-26
added option to choose between mouse events and pointer events

glromeo:
I am using useHover with buttons and other elements that can be disabled but I still want to be able to use float-ui to show tooltips for them.
This PR is aimed at making configurable the choice of whether to use the mouse events or the pointer events (which are not subject to limitations when the reference is disabled).
If you like this PR I will invest time to cover these changes with tests


glromeo:
doh! I just saw the discussion... ok I guess this doesn't make sense

PR #2397, opened 2023-06-28
fix(useListNavigation): skip hidden items

mihkeleidast:
Skips hidden elements from list navigation. Not sure if checking `offsetParent` is the best solution, feel free to suggest some alternative.

## Background

I am building this combobox with a tree popup. When a group is collapsed, the items inside are still rendered in react (because we need to get the label/id data back to the parent component), but hidden via CSS. When they are hidden, they should be skipped by list navigation.

image

### DIFF
--- a/packages/react/test/setupTests.ts
+++ b/packages/react/test/setupTests.ts
@@ -17,28 +17,37 @@ function isNullOrUndefined(a: any) {
 
 // From https://github.com/jsdom/jsdom/issues/1261#issuecomment-512217225
 Object.defineProperty(HTMLElement.prototype, 'offsetParent', {
-    get() {
-        // eslint-disable-next-line
-        let element = this;
-        while (!isNullOrUndefined(element) &&
-                    (isNullOrUndefined(element.style) ||
-                    isNullOrUndefined(element.style.display) ||
-                    element.style.display.toLowerCase() !== 'none')) {
-            element = element.parentNode;
-        }
+  get() {
+    // eslint-disable-next-line
+    let element = this;
+    while (
+      !isNullOrUndefined(element) &&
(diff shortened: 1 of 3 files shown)


packages/react/test/setupTests.ts
```diff
@@ -10,3 +10,35 @@ jest
...
+}
+
+// From https://github.com/jsdom/jsdom/issues/1261#issuecomment-512217225
+Object.defineProperty(HTMLElement.prototype, 'offsetParent', {
+    get() {
+        // eslint-disable-next-line
+        let element = this;
+        while (!isNullOrUndefined(element) &&
+                    (isNullOrUndefined(element.style) ||
+                    isNullOrUndefined(element.style.display) ||
```
atomiks, line 25: Prettier on this?

atomiks:
One edge case is that if the list item is `position: fixed` then it'll be marked as hidden even if it's not actually... I expect that to be extremely rare, but theoretically possible. Maybe should check if `offsetWidth`/height is `0` instead? Might be an edge case there...

mihkeleidast:
> One edge case is that if the list item is `position: fixed` then it'll be marked as hidden even if it's not actually... I expect that to be extremely rare, but theoretically possible. Maybe should check if `offsetWidth`/height is `0` instead? Might be an edge case there...

Hmm, could we punt this change for when such usecase presents itself? Checking `offsetWidth/offsetHeight` would, for example, bring in edge cases with `visibility: hidden;` elements (as those are hidden but still take up space). Not sure we need to choose one or the other edge case to fix right now, both should be rare, especially in the list navigation context.

atomiks:
Yeah agree it's fine enough for now

PR #2398, opened 2023-06-28
fix(FloatingList): allow ReactNode as label type

mihkeleidast:
One quite common thing in the React world is to have a `<Translate key="some.key" />` component which eventually renders a string. But if the hook/refs array does not allow `ReactNode`, it needs to be worked around. This will allow passing such component to the composition-based option directly & avoids type casting and other shenanigans.


atomiks:
`<Translate key="some.key" />` renders a JSX element though which the typeahead hook can't accept - why not send the `props.children` type which is a string

mihkeleidast:
Huh, that's actually a good point, I had not yet arrived at the typeahead intergration. I guess then the string is better after all, translate libs usually also provide a hook/function that returns a string which can be used. Would've been nice if it was more dynamic, but I totally understand the typeahead point.

PR #2405, opened 2023-06-30
chore: dep upgrades

atomiks:
- Upgrade most dev packages to latest, Rollup 2 to Rollup 3
- Move from Parcel to Vite
- Move from Jest to Vitest


packages/vue/test/index.test.ts
```diff
@@ -608,7 +609,7 @@ describe('useFloating', () => {
   });
 });
 
-describe('arrow', () => {
+describe.skip('arrow', () => {
   function setup(options?: Omit<ArrowOptions, 'element'>) {
     const reference = ref<ReferenceElement | null>(null);
     const floating = ref<FloatingElement | null>(null);
```
atomiks, line 615: @lozinsky I'm not sure why this test is failing after upgrading to Vitest. Any ideas?

```
TypeError: The provided value is not of type 'Element'.
 ❯ Object.exports.convert ../../../node_modules/jsdom/lib/jsdom/living/generated/Element.js:26:9
 ❯ Window.getComputedStyle ../../../node_modules/jsdom/lib/jsdom/browser/Window.js:820:19
 ❯ Module.getComputedStyle ../../dom/src/utils/getComputedStyle.ts:4:29
      2| 
      3| export function getComputedStyle(element: Element) {
      4|   return getWindow(element).getComputedStyle(element);
       |                             ^
      5| }
      6| 
 ❯ Module.getCssDimensions ../../dom/src/utils/getCssDimensions.ts:8:15
 ❯ Object.getDimensions ../../dom/src/utils/getDimensions.ts:6:10
 ❯ fn ../../core/src/middleware/arrow.ts:48:44
 ❯ Module.computePosition ../../core/src/computePosition.ts:45:15

This error originated in "index.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "allows to use with component type arrow". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- This was the last recorded test before the error was thrown, if error originated after test finished its execution.
```

atomiks: Nevermind — it seems to be due to the export collision of `arrow` in the index file, so it's picking up the `dom`, not `vue`-wrapped `arrow`. Fixed it by doing explicit exports like `react-dom`.

lozinsky: > Fixed it by doing explicit exports like react-dom.

Cool! 👍🏻

lozinsky (approved):

PR #2452, opened 2023-07-18
[core] optimize getRectRelativeToOffsetParent function

lio-mengxiang:
Remove duplicate code.

The return value of isHTMLElement(offsetParent) has already been retrieved, so there is no need to compute it again


atomiks:
Thank you!

PR #2476, opened 2023-07-31
feat: use vite build system instead

### LINKED ISSUE #2472: Generate `.d.mts` types for imports
**Describe the bug**

Upgrading @floating-ui/dom from version 1.2.9 to 1.50 will give following errors:

```
src/index.ts:6:42 - error TS7031: Binding element 'x' implicitly has an 'any' type.

6 computePosition(button, tooltip).then(({ x, y }) => {
                                           ~

src/index.ts:6:45 - error TS7031: Binding element 'y' implicitly has an 'any' type.

6 computePosition(button, tooltip).then(({ x, y }) => {
```

**To Reproduce**
https://codesandbox.io/p/sandbox/fancy-moon-t5p4ps?file=%2Fpackage.json%3A12%2C1-12%2C31

1. Run `yarn build`
2. See error

**Expected behavior**
No errors.

**Screenshots**

image

**Context:**
 - Version 1.5.0

fourdim:
This closes #2472.

image

Please review and test it carefully as I may miss something important.

### DIFF
--- a/.github/workflows/release-floating-ui.yml
+++ b/.github/workflows/release-floating-ui.yml
@@ -21,7 +21,6 @@ jobs:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/setup
       - run: npm run build
-      - run: npm run build:typescript
       - run: npm run lint-packages
       - run: npx rollingversions publish
         env:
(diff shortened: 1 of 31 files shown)


atomiks:
Thank you for the PR - a few things:
- We can't change the names of the files generated as it's breaking. So need to keep the browser files and `.mjs`
- Need to keep the ESM import map (`import`), so `.d.mts` files have to also be generated as well
- The non `.min` files shouldn't be minified

Issue #2456, opened 2023-07-20
@floating-ui/vue "MaybeReadonlyRef" => MaybeRefOrGetter


VividLemon:
I apologize for neglecting the bug report template. I'm on my phone making this and things are a bit tough.

The instance if the issue that I'm seeing is

```ts
useFloating(ref, ref, {
strategy: () => props.strategy
})
```
This synatx is available in Vue 3.3. The only way to make strategy reactive correctly is to create an intermediate ref like so (assuming the composable accepts reactivity, which I'm sure you are correctly doing this)
```ts
useFloating(ref, ref, {
strategy: toRef(props, 'strategy')
})
```

This isn't necessarily a bad thing, but can have performance benefits by not creating intermediate refs. In addition, the getter syntax is more straightforward. It would be nice if this was supported globally. I'm not necessarily sure if "Read-only" is required as a typing 

I suppose this is a feature request more than a bug report. I could submit a PR in a few days if needed

VividLemon:
I am planning on building a PR sometime next week. So no worries

atomiks:
@lozinsky can you take a look at this and the PR?

VividLemon:
Requires a breaking change. `toValue` is not supported by Vue 2 or @vue/composition-api

atomiks:
Ok, we can revisit once Vue 2 is officially EOL.

VividLemon:
> Ok, we can revisit once Vue 2 is officially EOL.

Perhaps a BREAKING_CHANGE label may help it so this issue doesn't get lost in the void

PR #2478, opened 2023-08-01
feat: leverage vue 3.3 toValue to normalize property getters

VividLemon:
Fixes https://github.com/floating-ui/floating-ui/issues/2456

In the older way, you were forced to do this syntax when working with reactive structures
```ts
strategy: toRef(props, 'strategy'),
```
However, this can be simplified to 
```ts
strategy: () => props.strategy,
```

This isn't quite a perfect solution. It appears vue-demi's type definitions aren't perfectly up to date with Vue 3.3. As 
```ts
const middlewareOption = computed(() => toValue(options.middleware));
```
Could be changed to
```ts
const middlewareOption = readonly(toRef(options, 'middleware'));
// Possibly 
const middlewareOption = toRef(() => options.middleware);
```
The latter is preferred as it's type definition is already that of readonly. The first option needs to be wrapped in a readonly as it allows for two way data binding

This would prove more performant, but TS complains. I suspect it's due to vue-demi not updating their TS.

### DIFF
--- a/packages/vue/src/useFloating.ts
+++ b/packages/vue/src/useFloating.ts
@@ -1,18 +1,18 @@
-import type {
-  FloatingElement,
-  MiddlewareData,
-  ReferenceElement,
+import {  
+  type FloatingElement,
+  type MiddlewareData,
+  type ReferenceElement,
+  computePosition,
 } from '@floating-ui/dom';
-import {computePosition} from '@floating-ui/dom';
-import type {Ref} from 'vue-demi';
 import {
+  type Ref,
   computed,
   getCurrentScope,
   onScopeDispose,
   ref,
   shallowReadonly,
   shallowRef,
-  unref,
+  toValue,
   watch,
 } from 'vue-demi';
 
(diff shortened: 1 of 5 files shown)


VividLemon:
For the life of me, I cannot figure out what that issue is with npm. :disappointed: 

Perhaps move to pnpm?

I mean, say the word and I'll do it lol

packages/vue/src/useFloating.ts
```diff
@@ -1,18 +1,18 @@
...
 import {
+  type Ref,
   computed,
   getCurrentScope,
   onScopeDispose,
   ref,
   shallowReadonly,
   shallowRef,
-  unref,
+  toValue,
```
lozinsky, line 15: Are you sure that `vue-demi` provides these APIs for Vue 2.7?

VividLemon: > Are you sure that `vue-demi` provides these APIs for Vue 2.7?

The issue is not that vue-demi does not provide apis for this, but it's more likely that @vue/composition-api does not, and vue@2 likely does not provide these exports. 

As such, this is likely not going to work.

In my opinion, this is such a little issue that it doesn't warrant any action, but in my opinion, with the Vue 2 end of life approaching soon, this library should probably think about just dropping vue-demi and switch to pure Vue 3 support.

Closing

PR #2484, opened 2023-08-07
fix(FloatingArrow): allow generic reference type context

### LINKED ISSUE #2483: [React] FloatingArrow does not honor specified `context` typing
**Describe the bug**
The `<FloatingArrow />` does not infer the `context` typing like it is done in other components; see:
- [The `FloatingArrow` context](https://github.com/floating-ui/floating-ui/blob/94ba5b7c7069c272c967e4eef6ba09a793880855/packages/react/src/components/FloatingArrow.tsx#L8C12-L8C27)
- [The `FloatingFocusManager` context](https://github.com/floating-ui/floating-ui/blob/94ba5b7c7069c272c967e4eef6ba09a793880855/packages/react/src/components/FloatingFocusManager.tsx#L50C21-L50C21)

When passing anything more specific than just `Element` to `useFloating()`, the [the `context` typing](https://github.com/floating-ui/floating-ui/blob/94ba5b7c7069c272c967e4eef6ba09a793880855/packages/react/src/types.ts#L123C1-L123C1) is improved, but that leads to a error with the context typing in FloatingArrow which did not gain that improvement.

The error looks like:
> Type 'Element' is not assignable to type 'HTMLButtonElement'

**To Reproduce**
1. Go to the example in [the FloatingArrow documentation](https://floating-ui.com/docs/FloatingArrow#scale-transforms) ([here is the CodeSandbox link](https://codesandbox.io/s/floating-ui-react-scale-transform-origin-qv0t1c?file=/src/App.tsx))
2. Scroll down to the call to `useFloating()`
3. Update it to by typed for the HTML button (i.e., `useFloating<HTMLButtonElement>(...)`)
4. See error: "Type 'Element' is not assignable to type 'HTMLButtonElement'"

**Current Workaround**
Currently, to appease TypeScript, I do the following: `context={context as unknown as FloatingContext}`.

**Expected behavior**
TypeScript should not complain (probably has similar behaviour to `FloatingFocusManager`).

**Context:**
 - _OS_: macOS Ventura v13.5 (22G74)
 - _Browser_: Google Chrome v115
 - _Version_: 115.0.5790.114 (Official Build) (arm64)

**Additional context**
Beware that `FloatingArrow` uses `React.forwardRef()`, so passing generic typing may be non-trivial.

atomiks:
Fixes #2483

@mattbrazza I tried a few things with generics but it seems hard with `forwardRef` :\  omitting the refs from the type fixes the problem and the component doesn't need that part of the context anyway

### DIFF
--- a/packages/react/index.test-d.tsx
+++ b/packages/react/index.test-d.tsx
@@ -85,9 +85,12 @@ function App() {
   const ref2 = useRef<HTMLDivElement>(null);
   const ref = useMergeRefs([ref1, ref2, arrowRef, null]);
 
+  const {context: contextGeneric} = useFloating<HTMLDivElement>();
+
   return (
     <div ref={ref} style={floatingStyles}>
       <FloatingArrow context={context} />
+      <FloatingArrow context={contextGeneric} />
       <FloatingArrow
         ref={arrowRef}
         context={context}
--- a/packages/react/src/components/FloatingArrow.tsx
+++ b/packages/react/src/components/FloatingArrow.tsx
@@ -5,7 +5,7 @@ import {useId} from '../hooks/useId';
 import type {Alignment, FloatingContext, Side} from '../types';
 
 export interface FloatingArrowProps extends React.SVGAttributes<SVGSVGElement> {
-  context: FloatingContext;
+  context: Omit<FloatingContext, 'refs'>;
 
   /**
    * Width of the arrow.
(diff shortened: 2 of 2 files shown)


mattbrazza:
_Follow-up_: I tried this one-liner fix real fast in my project and it seems that our TypeScript is happy and builds.

Are there any other steps I need to take to help with this P.R.?

atomiks:
Sorry forgot about this, no there's nothing else. I've now also added `refs: any` in case of literal passing of the context object prop.

mattbrazza:
I just updated the package, and as expected things are working great. Thanks again for the help!

PR #2492, opened 2023-08-18
fix: replace includes with indexOf to support old browser

guaqiu5:
replace includes with indexOf to support old browser

### DIFF
--- a/packages/core/src/middleware/offset.ts
+++ b/packages/core/src/middleware/offset.ts
@@ -45,7 +45,7 @@ export async function convertValueToCoords(
   const side = getSide(placement);
   const alignment = getAlignment(placement);
   const isVertical = getSideAxis(placement) === 'y';
-  const mainAxisMulti = ['left', 'top'].includes(side) ? -1 : 1;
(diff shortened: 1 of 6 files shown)


FezVrasta:
We don't support older browsers, making this change wouldn't make Floating UI compatible.

guaqiu5:
> We don't support older browsers, making this change wouldn't make Floating UI compatible.
 but it will cause the js error rate of the website to increase😫

FezVrasta:
The library is not going to work nonetheless, having an error makes that even clearer. We suggest to include polyfills for unsupported functionalities when deemed appropriate

guaqiu5:
image
This is the jserror stack information for users of low version browsers
Low version of the browser errors occurred in case of Array some & includes methods, my guess is that the "packages/utils/dom/SRC/index" isContainingBlock function cause it.

piecyk:
@guaqiu5 something is not right, as here we see import from core-js, basic you can configure what browser do you suport and it will automatically pick the polyfills

PR #2524, opened 2023-08-31
Modifier type: Make properties optional

sydneyjodon-wk:
Hello 👋 I'm not sure if PRs are still accepted into v2, but I'm wondering if these properties on `Modifier` can be made optional?

- `enabled`
- `phase`
- `fn`

Based on examples in the docs, it seems like they shouldn't be required. Here's [another example](https://codesandbox.io/s/popper-example-modifier-type-v9hrjq?file=/Demo.tsx) where it doesn't seem like those properties are required to be set to use the Modifier type. Is there a reason they are set to required?

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -134,11 +134,11 @@ export type ModifierArguments<Options: Obj> = {
 };
 export type Modifier<Name, Options: Obj> = {|
   name: Name,
-  enabled: boolean,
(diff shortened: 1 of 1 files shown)


atomiks:
Thank you for the PR and issue, but we're currently not accepting features or API changes to Popper any longer. Bug fixes are still welcome though.

PR #2525, opened 2023-09-01
chore: use changesets

atomiks:
Will add the CI step after manually trying it out first.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
       "devDependencies": {
         "@babel/preset-env": "^7.22.5",
         "@babel/preset-typescript": "^7.22.5",
+        "@changesets/cli": "^2.26.2",
         "@playwright/test": "^1.35.1",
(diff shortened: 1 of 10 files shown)


FezVrasta:
Why the change?

atomiks:
RollingVersions:
- is effectively unmaintained, having last released in February 2022. 
- is too limited in that you can't release "special" versions like pre-releases.
- is too limited in its changelog in that you can't place features under the Features heading while keeping it a patch version (as semver says to, which allows it to be updated when used transitively).
- web UI is frustrating and buggy to use, as I just tried to remove an entry from "Features" but it's not clearing the input field, so now the PR is broken basically.

All in all, Changesets is much better maintained and far more popular, and you have full control over versioning based on other libraries that use it. While Rolling Versions is simplistic, it's not powerful enough to do what we need to do right now. If the problems I listed above are fixed I'll gladly switch back to it, but right now it's not suitable for this library.

PR #2535, opened 2023-09-08
getOverflowAncestors: traverse into iframe parent

### LINKED ISSUE #2518: getOverflowAncestors should traverse through parent iframes
**Describe the bug**
Since #2043 the `getBoundingClientRect` DOM util function traverses all the parent iframes and adds their offsets, so that the computed offset is always relative to the top-level document.

But it turns out that one piece is missing for full seamless iframe support. The `autoUpdate` function has `ancestorResize` and `ancestorScroll` options. They cause that all ancestor scrolling elements get `resize` and `scroll` listeners and that position is recomputed when the reference and floating element get scrolled relative to each other.

But the `getOverflowAncestors` DOM util function doesn't traverse through parent iframes. It stops when it reaches the current frame's document element. If the iframe is inside a scrolling container:
```
<div id="floating">popover</div>
<div id="scrolling" style="overflow: auto">
  <iframe>
    <div id="reference">anchor</div>
  </iframe>
</div>
```
Then when the `#scrolling` contents scroll, then `autoUpdate` doesn't catch that.

**To Reproduce**
I don't have a minimal repro example yet. I'm interested in working on a fix, together with tests and all.

Cc @ciampo

jsnajdr:
Fixes #2518.

When `getOverflowAncestors` reaches the `window` object, it checks whether there is `window.frameElement` and continues traversing up into the iframe's parent.

TODO before this is shippable:
- add tests
- verify usage in `getClippingElementAncestors`. Do we also want to traverse into iframe's parent there?

### DIFF
--- a/packages/utils/dom/src/index.ts
+++ b/packages/utils/dom/src/index.ts
@@ -168,7 +168,8 @@ export function getOverflowAncestors(
     return list.concat(
       win,
       win.visualViewport || [],
-      isOverflowElement(scrollableAncestor) ? scrollableAncestor : []
+      isOverflowElement(scrollableAncestor) ? scrollableAncestor : [],
+      win.frameElement ? getOverflowAncestors(win.frameElement) : [],
     );
   }
 
--- a/packages/utils/test/getOverflowAncestors.test.ts
+++ b/packages/utils/test/getOverflowAncestors.test.ts
@@ -65,3 +65,20 @@ test('does treat display: inline-block elements as overflow ancestors', () => {
     window,
   ]);
 });
+
+test('returns overflow ancestors in iframe parents', () => {
+  const scroll = document.createElement('div');
+  scroll.style.overflow = 'scroll';
+  const iframe = document.createElement('iframe');
+  const test = document.createElement('div');
+
+  document.body.append(scroll);
+  scroll.append(iframe);
+  iframe.contentDocument!.body.append(test);
+
+  expect(getOverflowAncestors(test)).toEqual([
+    iframe.contentWindow,
+    scroll,
+    window,
+  ]);
+});
--- a/packages/utils/test/getOverflowAncestors.test.ts
+++ b/packages/utils/test/getOverflowAncestors.test.ts
@@ -74,7 +74,8 @@ test('returns overflow ancestors in iframe parents', () => {
 
   document.body.append(scroll);
   scroll.append(iframe);
-  iframe.contentDocument!.body.append(test);
+  expect(iframe.contentDocument).not.toBeNull();
+  iframe.contentDocument?.body.append(test);
 
   expect(getOverflowAncestors(test)).toEqual([
     iframe.contentWindow,
--- a/.changeset/late-falcons-sell.md
+++ b/.changeset/late-falcons-sell.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/dom": patch
+---
+
+fix: traverse into iframe parents when finding overflow ancestors


packages/utils/dom/src/index.ts
```diff
@@ -168,7 +168,8 @@ export function getOverflowAncestors(
     return list.concat(
       win,
       win.visualViewport || [],
-      isOverflowElement(scrollableAncestor) ? scrollableAncestor : []
+      isOverflowElement(scrollableAncestor) ? scrollableAncestor : [],
+      win.frameElement ? getOverflowAncestors(win.frameElement) : [],
```
tyxla, line 172: This could use a unit test - there is already an [existing test suite](https://github.com/floating-ui/floating-ui/blob/master/packages/utils/test/getOverflowAncestors.test.ts) for `getOverflowAncestors()`. 

We might need to configure [loading of subresources](https://github.com/jsdom/jsdom#loading-subresources) because `jsdom` by default won't load any subresources like iframes, and for an element inside the iframe, `node.ownerDocument.defaultView` will be equal to the parent `window` object.

packages/utils/test/getOverflowAncestors.test.ts
```diff
@@ -65,3 +65,20 @@ test('does treat display: inline-block elements as overflow ancestors', () => {
...
+
+test('returns overflow ancestors in iframe parents', () => {
+  const scroll = document.createElement('div');
+  scroll.style.overflow = 'scroll';
+  const iframe = document.createElement('iframe');
+  const test = document.createElement('div');
+
+  document.body.append(scroll);
+  scroll.append(iframe);
+  iframe.contentDocument!.body.append(test);
```
tyxla, line 77: We may need a `no-non-null-assertion` ignore here.

packages/utils/dom/src/index.ts
tyxla: Nice, great to hear that it wasn't a problem 👍

packages/utils/test/getOverflowAncestors.test.ts
```diff
@@ -65,3 +65,20 @@ test('does treat display: inline-block elements as overflow ancestors', () => {
     window,
   ]);
 });
+
+test('returns overflow ancestors in iframe parents', () => {
```
tyxla, line 69: Thanks for adding the test 🙌

tyxla (approved):
This looks great to me 👍 🚀 

Looking forward to getting this landed so we can remove the extra custom listeners in [#54286](https://github.com/WordPress/gutenberg/pull/54286)

packages/utils/test/getOverflowAncestors.test.ts
jsnajdr: `iframe.contentDocument` can be `null`, but by appending the element into the `document.body` structure, we are "attaching" it and it gets loaded. Then `contentDocument` and `contentWindow` get a reliable non-null value. But this certaintly can't be expressed in types.

atomiks: This looks good, but to prevent warnings appearing here, can you appease the checker?

jsnajdr: Warning fixed :+1:

jsnajdr:
Thanks @atomiks for merging the patch so quickly :+1: There might be, however, something wrong with the release. The `getOverflowAncestors` function is part of the `@floating-ui/utils` package, but only the `@floating-ui/dom` got released.

atomiks:
oh 🤦 I forgot that function was moved out into the `utils` package, only `utils` needed to be released. I'll fix it in a second.

PR #2560, opened 2023-09-16
fix(FloatingFocusManager): treat untrapped combobox reference as non-modal guards

atomiks:


packages/react/src/hooks/useFocus.ts
```diff
@@ -124,6 +124,8 @@ export function useFocus<RT extends ReferenceType = ReferenceType>(
           blockFocusRef.current = false;
           const relatedTarget = event.relatedTarget;
 
+          if (!relatedTarget) return;
```
okonet, line 127: This actually breaks the `useFocus` inside an iframe. I just stumbled upon it while experimenting with #1717 in the Storybook.

atomiks: Thank you for flagging. I haven't released a new version yet, so it would be good to handle the iframe case. The main issue is it causes the floating element to close when pressing on the browser chrome but shouldn't in various cases.

atomiks: One example is the Twitter search omnibox. When the input loses focus pressing the browser Chrome, it doesn't close. Another example is Ariakit's combobox, which also doesn't close in that scenario. It causes the floating element to flicker open/close when focus is moved out of the browser, which doesn't really look great.

atomiks: I'll revert this for the time being anyway until a better solution is found

okonet: Hmm, maybe this could become an option? I think most of the time it should close the floating element but I agree there are some corner cases.

diegohaz: If there's an option, I'd suggest setting the default to keep the popup visible even when interacting outside the window or frame. Users should always be able to hide it on the window's `blur` event, anyway.

Here are a few more reasons:

-  When managing multiple windows on the same screen or across multiple displays, you frequently shift focus between them. It's frustrating to lose the view state of other windows while they're still visible.
-  When using third-party apps to capture screenshots, clicking on the app's "Capture screenshot" button can cause the web document to lose focus. If the popup is hidden, it won't be included in the screenshot.
-  Some magnification apps also cause the document to lose focus when they're activated.

atomiks: Thank you for the input @diegohaz! @okonet what do you think of leaving the code as-is, and using a window blur listener to close it as a workaround instead?

okonet: Yeah if that solves the issue I won’t mind of course.

Issue #2552, opened 2023-09-14
Weird behaviour when using in Storybook


okonet:
**Describe the bug**
In Storybook, the content of the story (and docs) is displayed in an iframe. In my case, it looks like the floating element coordinates are affected by the outer `document` outside of the iframe that has the code running.

**To Reproduce**
I have tried creating a minimal reproducible case with the Stackblitz imitating the iframe behaviour but there it works as expected.

**Expected behavior**
The floating element should not be affected by the second `document` layout.

**Screenshots**

https://github.com/floating-ui/floating-ui/assets/11071/5affcbed-bd57-438a-82a4-0396cbd8c2ca

**Context:**
 - OS: MacOS
 - Browser: Chrome (but same in Safari)
 - Version: 116.0.5845.140

atomiks:
If you install `@floating-ui/utils@0.1.1` does the issue go away? Wonder if https://github.com/floating-ui/floating-ui/pull/2535 potentially caused an issue.

atomiks:
Just confirmed this is a regression due to that. There were no tests to catch this but can be added easily. I will fix it soon. But overriding `@floating-ui/utils@0.1.2` to use `0.1.1` will fix it for now.

okonet:
Thanks a ton for confirming and clarifying! I've tried applying the workaround but I still have the issue.

atomiks:
Try uninstalling and reinstalling. It will pick up the latest utils package which has the fix (0.1.3)

okonet:
Tried the latest version and it's the same :(

okonet:
I noticed that `@floating-ui/react` wasn't updated. Not sure that's the issue but that's what I'm using.

atomiks:
I forgot to mention `@floating-ui/dom` also needs to be latest. (But still, uninstalling and re-installing `@floating-ui/react` will pick up all the latest transitives, assuming that's the only Floating UI package installed.)

okonet:
Sorry for a late reply. I managed to installed the fix by removing and re-adding the `@floating-ui/react` as a dependency. I can confirm that it fixes the problem with the position. 

Thanks a lot for your help and hard work!

dgonzalezr:
I'm having the same issue with Storybook while using @floating-ui/dom@1.5.3
https://develop--6408a8161bf55e5999ae828b.chromatic.com/?path=/story/components-select--default
Installing the latest @floating-ui/utils does not fix the problem on my side unfortunately 😞

[...]

PR #2561, opened 2023-09-17
fix(getOverflowAncestors): handle traverseIframes correctly when there are clipping ancestors in the inner frame

kokovtsev:
Fixing an edge case where the #2553 solution is not working. When the element has clipping ancestor within the iframe, the `traverseIframe` parameter was not passed down into the recursive call


atomiks:
Thank you! 🙏 

I suppose that would explain why it didn't fully fix this https://github.com/floating-ui/floating-ui/issues/2552

kokovtsev:
Thanks for your prompt reply and the great library overall!

PR #2581, opened 2023-09-30
feat(useFocus): `visibleOnly` option

### LINKED ISSUE #2574: Restrict useFocus to only elements that has `:focus-visible` pseudo class applied
I'd like to make sure that the floating element is only triggered by elements that has [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) to provide a better UX for  keyboard and pointer users.

I would assume that in order to detect this, it would be sufficient to add a check

```javascript
event.target.matches(':focus-visible')
```

in this event listener:

https://github.com/floating-ui/floating-ui/blob/07d8e853c81925316bf712270f9533962bc19fda/packages/react/src/hooks/useFocus.ts#L106-L122

If that's correct, I could draft a PR to implement this. 

WDYT?

atomiks:
This replaces `keyboardOnly` since it better does what should be intended (matching `:focus-visible` selector).

Closes #2574

### DIFF
--- a/.changeset/dirty-phones-pay.md
+++ b/.changeset/dirty-phones-pay.md
@@ -1,0 +1,6 @@
+---
+'@floating-ui/react': minor
+---
+
+feat(useFocus): replace `keyboardOnly` option with `visibleOnly` (matches
+:focus-visible CSS selector)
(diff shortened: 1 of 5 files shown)


okonet (approved):
Lgtm

PR #2583, opened 2023-10-01
fix(useFocus): avoid blur close only on browser page blur

### LINKED ISSUE #2575: Losing focus with `useFocus` not closing the floating element inside an iframe
This actually breaks the `useFocus` inside an iframe. I just stumbled upon it while experimenting with #1717 in the Storybook.

_Originally posted by @okonet in https://github.com/floating-ui/floating-ui/pull/2560#discussion_r1335693237_

atomiks:
Fixes #2575 

This amends `activeElement` function to traverse (allowed) iframe parent documents to find the active element. If it's `null` or the `body`, then it's assumed focus left the page into the chrome/elsewhere, so the blur should be ignored. This should fix local iframe issues while ensuring the floating element doesn't close on blur in other situations unwantedly

### DIFF
--- a/.changeset/six-boxes-decide.md
+++ b/.changeset/six-boxes-decide.md
@@ -1,6 +1,5 @@
 ---
 '@floating-ui/react': patch
-'@floating-ui/utils': patch
 ---
 
-fix(activeElement): traverse iframes
+fix(useFocus): close on blur only if focus remains in document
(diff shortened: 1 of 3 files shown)


okonet (approved):

okonet:
Thanks so much!

PR #2595, opened 2023-10-09
Call out that transition styles must be nested in floating

nfiacco:
Otherwise the rendering won't work correctly.


atomiks:
Thank you; can you use the Notice component to place this in? And maybe wording (with an example) similar to here: https://floating-ui.com/docs/useFloating#transform

atomiks:
0475c02d8129d364e0073b19e66f27d50e8631ad

Issue #2594, opened 2023-10-07
[React] Add "capture" option to useDismiss


YanZisuka:
**Is your feature request related to a problem? Please describe.**
When a modal dialog appears, we stop propagating events in the modal overlay because we don't want it to interact with the UI in the background.

In this case, if the modal's child element has a component that uses useDismiss, the dismiss event will be interrupted before it reaches the document and `closeOnPressOutside` will not work, so I want to use capture to detect the click.

**Describe the solution you'd like**
We can check the implementation at [airbnb/react-outside-click-handler
](https://github.com/airbnb/react-outside-click-handler/blob/master/src/OutsideClickHandler.jsx#L29)

so, in [useDismiss.ts](https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/hooks/useDismiss.ts).
Briefly, consider the following code:

```tsx
export interface UseDismissProps {
  enabled?: boolean;
  escapeKey?: boolean;
  ...
  capture?: boolean | {escapeKey?: boolean; outsidePress?: boolean}; // new line
}

...

const doc = getDocument(floating);
    escapeKey && doc.addEventListener('keydown', closeOnEscapeKeyDown, {capture: escapeKeyCapture});
    outsidePress &&
      doc.addEventListener(outsidePressEvent, closeOnPressOutside, {capture: outsidePressCapture});
```

Thank you for the nice library and dedication. :)

**Describe alternatives you've considered**
N/A

**Additional context**
[Capture phase events in React's official docs](https://react.dev/learn/responding-to-events#capture-phase-events)

atomiks:
There's an open PR already: https://github.com/floating-ui/floating-ui/pull/2345

But it has a problem, because capture phase is used to detect if "outside press" originated inside the React tree (i.e., in a nested portaled floating element) even if it's outside of the DOM tree.

YanZisuka:
Thanks for your reply!

Well, I don't think I fully understand the problem yet.

But I think the behavior you're talking about is consistent with the design intent, since React doesn't want programmers to think much about the actual DOM.

atomiks:
React's event system bubbles through portaled React elements, which allows us to determine if a nested floating element (by the React tree) has been pressed, even when it's not a child of the parent in the DOM. This prevents it from closing unexpectedly.

The technique used to do this uses capture phase so it runs before the bubble phase. If we switch the main handler to capture too, then this fires afterwards and breaks the logic.

[...]

PR #2597, opened 2023-10-13
fix(useDismiss): add capture phase option for useDismiss

YanZisuka:
fixes
- #2345 
- #2594 

In the document capture phase, a main handler is added to `event.target`, and during the "AT_TARGET" phase, the handler is executed to operate the logic that determines `insideReactTree`.

### DIFF
--- a/packages/react/src/hooks/useDismiss.ts
+++ b/packages/react/src/hooks/useDismiss.ts
@@ -66,6 +66,7 @@ export interface UseDismissProps {
   outsidePressEvent?: 'pointerdown' | 'mousedown' | 'click';
   ancestorScroll?: boolean;
   bubbles?: boolean | {escapeKey?: boolean; outsidePress?: boolean};
+  capture?: boolean;
 }
 
 /**
@@ -94,6 +95,7 @@ export function useDismiss<RT extends ReferenceType = ReferenceType>(
     referencePressEvent = 'pointerdown',
     ancestorScroll = false,
     bubbles,
+    capture = true,
   } = props;
 
   const tree = useFloatingTree();
@@ -282,6 +284,14 @@ export function useDismiss<RT extends ReferenceType = ReferenceType>(
     onOpenChange(false, event);
   });
 
+  const lazyCloseOnPressOutside = useEffectEvent((event: MouseEvent) => {
+    event.target?.addEventListener(
+      outsidePressEvent,
+      () => closeOnPressOutside(event),
+      {once: true}
+    );
+  });
+
   React.useEffect(() => {
     if (!open || !enabled) {
       return;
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/useDismiss.ts
```diff
@@ -66,6 +66,7 @@ export interface UseDismissProps {
   outsidePressEvent?: 'pointerdown' | 'mousedown' | 'click';
   ancestorScroll?: boolean;
   bubbles?: boolean | {escapeKey?: boolean; outsidePress?: boolean};
+  capture?: boolean;
```
atomiks, line 69: I think this should just be the default with no option. We can add it later if it proves to be necessary.

I'm also not sure about the "once" listener technique, maybe it's better to always remove it to prevent leaks. And for minification, just use `true` rather than the options object.

YanZisuka: > I'm also not sure about the "once" listener technique, maybe it's better to always remove it to prevent leaks.

I'm confused as to whether this line is exactly what I modified, but I reflected it. [ccc9370](https://github.com/floating-ui/floating-ui/pull/2597/commits/ccc9370ae9a4062819bc0955a8cdbbc27b8bfab0)

thank you for review :)

atomiks: Ok I see, only thing is `capture` would need to be scoped to the `outsidePress` option specifically i.e. `outsidePressCapture`

packages/react/src/hooks/useDismiss.ts
```diff
@@ -282,6 +284,14 @@ export function useDismiss<RT extends ReferenceType = ReferenceType>(
     onOpenChange(false, event);
   });
 
+  const closeOnPressOutsideCapture = useEffectEvent((event: MouseEvent) => {
+    const callback = () => {
+      closeOnPressOutside(event);
+      event.target?.removeEventListener(outsidePressEvent, callback);
+    };
```
atomiks, line 291: Might need to use the `getTarget` util to support shadow DOM

packages/react/src/hooks/useDismiss.ts
```diff
@@ -63,6 +63,7 @@ export interface UseDismissProps {
   referencePress?: boolean;
   referencePressEvent?: 'pointerdown' | 'mousedown' | 'click';
   outsidePress?: boolean | ((event: MouseEvent) => boolean);
+  outsidePressCapture?: boolean;
```
atomiks, line 66: Actually, thinking about it, I think this might need to be the same type as `bubbles`, which is `boolean | {outsidePress: boolean, escapeKey: boolean}`. This allows using capture phase for `escapeKey` if necessary as well to prevent too many props and be forward-looking? 

The `normalizeBubblesProp` could be changed to accomodate that.

YanZisuka: I agree. Would it be better to just rename normalizeBubblesProp? Or would it be better to create normalizeCaptureProp?

atomiks: Yeah since they do the same thing, can just be `normalizeProp` to reuse it for both

PR #2600, opened 2023-10-14
fix(useFocus): improve visibleOnly detection

atomiks:


packages/react/src/hooks/useFocus.ts
```diff
@@ -38,7 +40,7 @@ export function useFocus<RT extends ReferenceType = ReferenceType>(
 
   const blockFocusRef = React.useRef(false);
   const timeoutRef = React.useRef<number>();
-  const keyboardModalityRef = React.useRef(false);
+  const keyboardModalityRef = React.useRef(true);
```
henrinormak, line 43: @atomiks could you elaborate as to why this change to the initial value? Asking as I have stumbled upon a test scenario, which prior to 0.26.1 passed - the test is in essence checking for https://github.com/floating-ui/floating-ui/issues/2574, as we had that issue surface in our code some time ago.

In short, currently it seems that as this by default is true, and L108 triggering in tests (event appearing as virtual), then this is not correctly lowered to false and thus the linked issue regresses. If I remove either L108 and the virtual event check, or change this default to false, I get back the previous behaviour.

henrinormak: I did not open an issue as maybe I'm misunderstanding something, but if you agree that the initial value should not be true, I can also open a PR.

atomiks: I believe it was necessary in terms of being a good fallback with touch screen readers. It's better to ensure it'll be shown in case `:focus-visible` checks end up failing (i.e., assume it's `true`). This might already be an open issue (https://github.com/floating-ui/floating-ui/issues/2666) already though?

henrinormak: Indeed, thanks for pointing that out!

PR #2621, opened 2023-11-07
feat(useRole): add label role

atomiks:


packages/react/src/hooks/useRole.ts
```diff
@@ -30,16 +31,19 @@ export function useRole<RT extends ReferenceType = ReferenceType>(
...
-    if (!enabled) {
-      return {};
-    }
+    const floatingProps = {
+      id: floatingId,
+      ...(role !== 'label' && {role}),
+    };
 
-    if (role === 'tooltip') {
+    if (role === 'tooltip' || role === 'label') {
```
ri12888, line 41: I'm not sure `label` is a valid ARIA role defined in [the WAI-ARIA 1.3 spec](
https://www.w3.org/TR/wai-aria-1.3/#role_definitions), nor does there seem to be a mechanism for custom ARIA roles. Would it be better to allow users to disable and override the `labelledby/describedby` attrs?

atomiks: It's not a native role (`AriaRole`) but is typed as `ComponentRole` like `select` or `combobox`, adding the attributes in a way that fits the role

PR #2623, opened 2023-11-08
refactor(findNonDisabledIndex): optimize loop.

YonLJ:


packages/react/src/utils/composite.ts
```diff
@@ -51,17 +51,23 @@ export function findNonDisabledIndex(
...
+      return element == null ||
+        element.hasAttribute('disabled') ||
+        element.getAttribute('aria-disabled') === 'true';
+    };
+
   let index = startingIndex;
+  amount = decrement ? -amount : amount;
   do {
-    index = index + (decrement ? -amount : amount);
+    index = index + amount;
```
atomiks, line 66: Can be `index += decrement ? -amount : amount`

PR #2631, opened 2023-11-19
Clarify the purpose of react-dom and react in getting-started.mdx

sam-hosseini:
Hello Federico and atomiks! 👋🏼😊

Thank you very much for having made popper.js/floating-ui and maintaining them all these years 🙌🏼

I wanted to integrate floating-ui for positioning only (as I use other libraries for interactions) and started reading the documentation on the "Getting Started" page. I thought it'd be better if the Getting Started page already had the positioning-only-package install command, as their difference isn't clear until the user goes to the /docs/react page.

I used the same language found on /docs/react page for consistency.

### DIFF
--- a/website/pages/docs/getting-started.mdx
+++ b/website/pages/docs/getting-started.mdx
@@ -64,22 +64,28 @@ npm install @floating-ui/dom
 
 ### React
 
-Use with [React](/docs/react) or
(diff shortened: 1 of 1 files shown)


atomiks:
Thank you. I changed the presentation a bit to default to the full package first.

Issue #2099, opened 2023-01-08
[React] Make `useRole` more useful


atomiks:
Right now, `useRole` accepts the `role` of the floating element and using that, adds props to the floating element and reference elements to add screen reader support. 

## Problem

- Roles aren't specified for `item`s so they need to be manually added if needed.
- `listbox` shouldn't necessarily add `role="combobox"` to the reference element, e.g. for a select button rather than an input — sometimes there are bugs like with VoiceOver in Safari, so you need to add `role={undefined}` after the prop getters spread.
- Some additional props must be added based on the type of component which is more involved than is known from the role of the floating element, e.g. `aria-autocomplete`.
- There are some nuances around tooltips with `role="tooltip"`, e.g. using `aria-labelledby` instead of `aria-describedby` and omitting that role entirely for certain types of tooltips which has been discussed here previously.

## Solution

Configuration could be based on the component rather than the base floating role, such as:

- `type: 'tooltip'`
- `type: 'label'`
- `type: 'autocomplete'`
- `type: 'select'`
- `type: 'multiselect'`
- etc

`role` option could continue to exist for uncommon components to still enable custom configs where the base props that add `aria-expanded`/`aria-controls` are very useful, but this `type` option should help simplify things and ensure good defaults in the majority of use cases

ayushi0809:
HI @atomiks  can I work on this feature?

atomiks:
Sure, [this is the hook](https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/hooks/useRole.ts)

PR #2633, opened 2023-11-23
feat(useRole): Extended useRole Hook to Support Additional Role Types

tutizaraz:
Hey folks,

I've been tinkering with the `useRole` hook and added some cool stuff based on requirements [here](https://github.com/floating-ui/floating-ui/issues/2099). Now, we can specify more role types like `tooltip`, `label`, `autocomplete`, `select`, and `multiselect`. This should help components play nicer with screen readers.

I've also added some checks to make sure we're using valid roles and types. If something's off, it'll throw an error.

I'm no expert in screen readers or accessibility, but I think these changes will make a difference. They should make the `useRole` hook more flexible and easier to use, especially for those tricky cases where we need custom configurations.

Looking forward to hearing your thoughts!

### DIFF
--- a/packages/react/src/hooks/useRole.ts
+++ b/packages/react/src/hooks/useRole.ts
@@ -3,6 +3,8 @@ import * as React from 'react';
 import type {ElementProps, FloatingContext, ReferenceType} from '../types';
 import {useId} from './useId';
 
+type TypeKeys = 'tooltip' | 'label' | 'autocomplete' | 'select' | 'multiselect';
+
 export interface UseRoleProps {
   enabled?: boolean;
   role?:
@@ -14,7 +16,12 @@ export interface UseRoleProps {
     | 'listbox'
     | 'grid'
     | 'tree';
+  type?: TypeKeys;
 }
+
+type FloatingProps = {
+  [key: string]: string | boolean | undefined;
+};
 
 /**
  * Adds base screen reader props to the reference and floating elements for a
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/useRole.ts
```diff
@@ -26,18 +33,64 @@ export function useRole<RT extends ReferenceType = ReferenceType>(
...
+  }
+
+  if (
+    type &&
+    !['tooltip', 'label', 'autocomplete', 'select', 'multiselect'].includes(
+      type
+    )
+  ) {
+    throw new Error(`Invalid type: ${type}`);
+  }
```
atomiks, line 63: These shouldn't be necessary with TypeScript, runtime checking is only added if the types can't check it properly

atomiks (commented):
It seems the `type` prop as implemented here doesn't result in following ARIA guidelines correctly, and there's an issue with conflicting combinations. For example, `role: "dialog"` and `type: "tooltip"` doesn't make sense and the resulting attributes are not correct.

tutizaraz:
> It seems the `type` prop as implemented here doesn't result in following ARIA guidelines correctly, and there's an issue with conflicting combinations. For example, `role: "dialog"` and `type: "tooltip"` doesn't make sense and the resulting attributes are not correct.

Thank you for the feedback. Indeed, I did somewhat fail with this solution. I was working on it quite late and didn't take into account some important aspects. But now, I want to ask and add some code to find out if I'm moving in the right direction.

```js
  switch (type) {
      case 'tooltip':
        return {
          reference: {
            'aria-describedby': open ? floatingId : undefined,
          },
          floating: {
            ...floatingProps,
            role: role || 'tooltip',
          },
        };
      case 'label':
        return {
          reference: {
            'aria-labelledby': open ? floatingId : undefined,
          },
          floating: floatingProps,
        };
      case 'autocomplete':
        return {
          reference: {
            'aria-autocomplete': 'both',
            'aria-controls': open ? floatingId : undefined,
            'aria-expanded': open ? 'true' : 'false',
          },
          floating: floatingProps,
        };
      case 'select':
      case 'multiselect':
        return {
          reference: {
            'aria-haspopup': 'listbox',
            'aria-expanded': open ? 'true' : 'false',
            ...(isMultiSelect && {'aria-multiselectable': 'true'}),
          },
          floating: {
            ...floatingProps,
            role: role || 'listbox',
          },
        };
      default:
        return {
          reference: {
            'aria-expanded': open ? 'true' : 'false',
            'aria-controls': open ? floatingId : undefined,
            ...(role === 'listbox' && {role: 'combobox'}),
            ...(role === 'menu' && {id: referenceId}),
          },
          floating: {
            ...floatingProps,
            role,
            ...(role === 'menu' && {'aria-labelledby': referenceId}),
          },
        };
    }
```

@atomiks

atomiks:
One of the problems I wanted to solve was `item` roles for the `type`. For example, a `select` type would do:

```js
const referenceProps = {
  role: 'combobox',
  'aria-expanded': open,
  'aria-haspopup': 'listbox',
};

const floatingProps = {
  role: 'listbox',
};

const itemProps = {
  role: 'option',
  // Here's the problem - we don't have access to these:
  'aria-selected': index === selectedIndex,
};
```

This suggests it needs to be a function that accepts the data at spread time:

```js
{...getItemProps({ index, activeIndex, selectedIndex })}
```

e.g.

```js
const itemProps = ({index, selectedIndex}) => ({
  role: 'option',
  'aria-selected': index === selectedIndex,
});
```

`useInteractions` and the types probably needs to be adjusted to support this functionality

atomiks:
Superseded by https://github.com/floating-ui/floating-ui/pull/2658

Issue #2649, opened 2023-12-05
Floating UI Devtools


bsunderhus:
**Is your feature request related to a problem? Please describe.**
 
Debugging positioning issues is time consuming! There are also a lot of concepts to understand to know how positioning might be wrong, and it's quite hard to grasp how the way one is using Floating UI actually affects positioning itself.
 
**Describe the solution you'd like**
 
[Devtools for Floating UI positioned elements](https://chromewebstore.google.com/detail/floating-ui-devtools/ninlhpbnkjidaokbmgebblaehpokdmgb?hl=en) to  ease debugging without needing to use code breakpoints in Floating UI. Some datapoints to track for these devtools that come to my mind:
 
1. target element
2. floating element
3. middleware data
4. what caused a reset?!
7. original/flip position
 
**Additional context**
 
Hey there @atomiks @FezVrasta , I'm reaching out on behalf of [Fluent UI](https://github.com/microsoft/fluentui) library, we use Floating UI as one of our foundation dependencies for positioning components, such as [Menu](https://react.fluentui.dev/?path=/docs/components-menu-menu--default), [Tooltip](https://react.fluentui.dev/?path=/docs/components-tooltip--default) and [Popover](https://react.fluentui.dev/?path=/docs/components-popover--default) and other components.
 
It came to our attention a pattern of multiple partners and collaborators whom would contact us with problems/issues related to Fluent UI positioning components, and by the end of the day, with some investigation, most of those problems comes down to a lack of information between the person using our components and how does their use cases might actually influence on the appearance/positioning of those components, the number one strategy to debug those problems would involve using break points on the floating UI code to figure out how the configuration provided actually ends up impacting the element's styling.
 
We're most interested into collaborating with the creation of tools to mitigate this problem, as it seems to us that this is a complex issue and it'll not be easily solved by simply documenting countless edge cases. I'd love to have some feedback from you as authors of Floating UI on ideas and visions you might have for some sort of devtool that could help the community to better understand and deal with positioning problems.

As a first idea we took the liberty to develop [floating-ui-devtools](https://github.com/bsunderhus/floating-ui-devtools), but keep in mind that it's still quite beta.

bsunderhus:
> Love the extension!

I'm glad to hear that 🤩

We're still investigating what is the best data to be showed by the devtools panel.
At the moment I'd say those are the main next todos:

1. Show data history (at the moment the panel displays the last available value provided through the middleware chain)
[...]

PR #2653, opened 2023-12-06
feature(devtools): migrate from local repo to floating ui repo

bsunderhus:
Proposal: migrate devtools from personal repo (https://github.com/bsunderhus/floating-ui-devtools) to floating ui repo

Follow up of: https://github.com/floating-ui/floating-ui/issues/2649


extension/.storybook/decorators.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { FluentProvider, webDarkTheme, webLightTheme } from '@fluentui/react-components';
```
FezVrasta, line 2: Could you help me understand why we want to tie this to Fluent UI?

bsunderhus: We don't, I'll move it to the Fluent example, thx for the catch!

bsunderhus:
I'll keep this as draft for now as I still need to clean up some things

1. add proper readmes and contribution guides
2. properly integrate with the way things work in floating ui monorepo

bsunderhus:
Sorry upfront for the huge PR! Here's a sum up:

1. creates an `extension` folder containing a private package to build the chrome extension
2. creates a `packages/devtools` folder containinig everything related to the devtools middleware package exposing the minimal amount of information.
3. update tsconfig files to exclude `packages/devtools` from it's type verifications, as that package verification is not compatible
4. add estlint rule to avoid the usage of `chrome` global namespace

The extension is shipping by default with support to raw usages of floating ui and also supporting the FluentUI library use case. The idea here is that component libraries (such as fluent) will be using floating ui internally, and they might be interested in adding their own custom panel to ensure their clients have the support they need.

I've added a session in the REAMDE file of the extension folder explaining how you can add your own custom library middleware data (would be nice to have other major libraries implementing their own custom panel here, like Material UI)


## Next Steps

1. The `SidePanel` used to add extra helper functionalities is currently implemented using Fluent, this leaks FluentUI to the main bundle, ideally we should use a custom UI there instead of FluentUI
2. rename `floating-ui-devtools` to `@floating-ui/devtools` to follow other packages
3. follow up on https://github.com/floating-ui/floating-ui/issues/2649 todos

bsunderhus:
> Thanks this looks great! Any plan to add a CI step to publish the extension as well?

Yeah, that would be the next step, together with renaming the package from `floating-ui-devtools` to `@floating-ui/devtools` I can do this in this PR also, I'll update later on today

bsunderhus:
Ok, I've added the building process of `@floating-ui/devtools` to the `build` script which will be executed by the `release` CI if I understood correctly @FezVrasta, also change the name of the package and I guess this will trigger a release as this package has never been published before. Is this enough to include the package in the release pipeline?

FezVrasta:
I'll need @atomiks review here as I'm not informed on how the floating-ui new repo architecture works but what I see looks solid to me.

Regarding the build process, I hoped we could have a GitHub Actions workflow that directly publishes to the store any new version. Where I work we use `chrome-webstore-upload-cli` to make it happen

bsunderhus:
> I'll need @atomiks review here as I'm not informed on how the floating-ui new repo architecture works but what I see looks solid to me.
> 
> Regarding the build process, I hoped we could have a GitHub Actions workflow that directly publishes to the store any new version. Where I work we use `chrome-webstore-upload-cli` to make it happen

I'll need to register tokens on the Github Actions Pipeline for that, and I can't do that at the moment as I'm not an internal contributor to the repository with this permissions. How about we leave this automation for a separate PR?

atomiks:
Thank you for this PR! I'm sorry I didn't get to look at it until now. Taking a look now...

atomiks:
This is a big PR as you mentioned, but from briefly testing the extension and the local devtools package it appears to work as expected. It's probably going to be hard for me to personally contribute to the extension, but as long as you have the ownership of it (PRs etc) then it seems fine to add to this repo. 

Thank you!

FezVrasta:
Are we ready to merge or there's some work left? Just let me know and I'll hit the button

bsunderhus:
Thx for the review @atomiks! I'll be following up with the contribution to maintain the extension for sure. And let me know if we can create a token in the pipeline to automate the publishing of the extension, in the meantime I'll keep the publications manually.

This is ready to be merged @FezVrasta , once it's merged I'll update https://github.com/floating-ui/floating-ui/issues/2649 with next steps!

FezVrasta:
Thanks, if you want to send me the token by email you can reach me at federico.zivolo (AT) gmail

karthik15981:
image

This change is breaking in Web Pack 4..

bsunderhus:
> image
> 
> This change is breaking in Web Pack 4..

@karthik15981 I'd guess the problem here is that your webpack configuration does not support [Nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) operator?!


@FezVrasta @atomiks , what is the ES target for floating-ui, by the configuration files I'd guess it's ES2020, no?! should I strip away nullish coalescing operators from the distribution?

karthik15981:
@bsunderhus Yes, it's the nullish coalescing operator that's breaking in Web Pack 4..

atomiks:
The other packages do transpile the nullish coalescing already; I think the `devtools` package needs a `babel.config.js` file for Rollup to read from, then it should transpile it. (Need to verify if that's actually what causes it to be transpiled tho)

bsunderhus:
The devtools package is bundling with vite actually. I can create a PR to migrate it to the same rollup configuration as the other packages for the sake of following a pattern

atomiks:
Another thing is, the dist files should be `.js` for cjs and fake esm, and `.mjs` for node esm, since `.cjs` isn't supported in Create React App or something...

> The devtools package is bundling with vite actually. I can create a PR to migrate it to the same rollup configuration as the other packages for the sake of following a pattern

Yeah, that would also solve the above issue.

PR #2667, opened 2023-12-19
bugfix(devtools): migrates from vite to rollup

bsunderhus:
Follow up on https://github.com/floating-ui/floating-ui/pull/2653#issuecomment-1861456768 ,

## Modifications

1. removes `vite` in favor of a `rollup` configuration which follows the same pattern provided by the other packages in the monorepo
2. ensures the compilation uses `@babel/preset-env` to properly remove nullish coalescing operators https://github.com/floating-ui/floating-ui/pull/2653#issuecomment-1861331610
3. opt-in into local tsconfig system
4. properly declare emitted files on `package.json` follow same pattern as provided by the other packages in the monorepo

## Problems with this approach

I couldn't find a proper way to link dependency between the `packages/devtools` and `extension` packages, I ended up having to duplicate code from `extension` into `packages/devtools/src/extension` this is not ideal, but it would require a proper reference between projects using [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) and a proper transpilation of types, my suggestions to solve this would be:

1. have a monorepo orchestrator (like [turborepo](https://turbo.build/repo)) to ensure the chain of dependencies is respected while transpiling
2. have tsconfig dedicated to each package and linking between each other through [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html)
3. **EXTRA** would be nice to have the type generation at build time to avoid a separate configuration just for it.


bsunderhus:
> ## Problems with this approach

> I couldn't find a proper way to link dependency between the `packages/devtools` and `extension` packages, I ended up having to duplicate code from `extension` into `packages/devtools/src/extension` this is not ideal, but it would require a proper reference between projects using [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) and a proper transpilation of types, my suggestions to solve this would be:

> 1. have a monorepo orchestrator (like [turborepo](https://turbo.build/repo)) to ensure the chain of dependencies is respected while transpiling
> 2. have tsconfig dedicated to each package and linking between each other through [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html)
> 3. **EXTRA** would be nice to have the type generation at build time to avoid a separate configuration just for it.


I'd be willing to follow up on this with another PR, wdyt @atomiks ?

atomiks:
@bsunderhus that sounds like it would be great so feel free - would be good if https://github.com/floating-ui/floating-ui/issues/2472 could also be fixed (`.d.mts` types)

bsunderhus:
> @bsunderhus that sounds like it would be great so feel free - would be good if #2472 could also be fixed (`.d.mts` types)

Great, I can check on that also!

completely forgot to bump a patch version in the devtools in this PR🙈, I'll create a separate one for that

PR #2677, opened 2023-12-28
chore: migrates to turborepo

bsunderhus:
This PR (is a follow up of https://github.com/floating-ui/floating-ui/pull/2667#issuecomment-1862258711) migrates the whole monorepo structure by relying on [Turbo Repo](https://turbo.build/repo) to declare dependencies between packages tasks.

Here are the main changes:

 1. Creates a `config` folder containing a definition for:
     * rollup (`defineRollupConfig`)
     * vite (`defineViteConfig`)
     * binary file to build types (`build-api`) (it generates `.d.ts` and `.d.mts` files, Fixes https://github.com/floating-ui/floating-ui/issues/2472) based on [API extractor](https://api-extractor.com/)
     * `tsconfig.base.json` (a base typescript configuration that all configurations should extend from)
     * `api-extractor.base.json` (a base api-extractor configuration)
     * `env.d.ts` with globals (only global is `__DEV__`)
 2. Migrates the declaration of `tsconfig` dependencies to use [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html), each package has it's own typescript configuration
    * uses `moduleResolution: Bundler` to ensure we're using the latest module resolution and follow the `exports` provided by `package.json` properly
 3. Deletes all global declaration on `.config`
 4. Removes `resize-observer` (they're unnecessary once type conflicts are solved)
 5. Deletes patch on `vue` (patch becomes unnecessary once type conflicts between vue and react have been solved)
 6. All `rollup.config.mjs` files from each package now invoke `defineRollupConfig` to reuse the same base configuration (each package has it's own perks, but by the end of the day there should be a single configuration to build them all)
 7. All `vite.config.mts` files from each package now invoke `defineViteConfig` to reuse the same base configuration
 8. `extension`  and `config` has been added as part of the workspace to ensure `npm install` will install all their dependencies too (no dependency conflict is generated by this)
 9. Flattens all files into `src` folders (in `utils` for example, files were scattered between `src`, `dom` and `react` folders, with the new `defineRollupConfig` we can have all the files in `src` and define different outputs for each of the separate entries, producing `floating-ui.utils`, `floating-ui.utils.dom` and `floating-ui.utils.react` files in a flat structure.
 10. uses [pnpm](https://pnpm.io/) to resolve dependencies instead of npm


bsunderhus:
Sorry upfront again for the huuuuge PR @atomiks , this is as we talked, an attempt on reorganizing a bit the monorepo structure in a way where the type declarations are generated in the distribution folder along. Let me know what you think, I'll keep this as a Draft for now so that we can discuss how viable this modifications are, I don't expect a full review on everything as it's a huge modification, and I'm willing to break this into smaller PRs if required 😁, but this is basically a gist of what I mentioned of possible ways to solve the typings dependencies issues https://github.com/floating-ui/floating-ui/pull/2667#issuecomment-1862258711

bsunderhus:
TODO: fix playwright, seems like configuration file should be moved to `dom` package in this structure

atomiks:
Thanks!

> The biggest challenge now is how npm handles dependency resolutions, I noticed most type conflicts in this repo is basically because some global types are fighting each other. I'd suggest migrating either for [pnpm](https://pnpm.io/) or [yarn](https://yarnpkg.com/) as both of those package manager have more reliable tools for resolving dependency resolution conflicts

Let's go with pnpm if possible 👍

packages/core/src/index.ts
```diff
@@ -1,11 +1,54 @@
...
-export {arrow} from './middleware/arrow';
-export {autoPlacement} from './middleware/autoPlacement';
-export {flip} from './middleware/flip';
-export {hide} from './middleware/hide';
-export {inline} from './middleware/inline';
-export {offset} from './middleware/offset';
-export {limitShift, shift} from './middleware/shift';
-export {size} from './middleware/size';
+export {type DetectOverflowOptions, detectOverflow} from './detectOverflow';
+export {type ArrowOptions, arrow} from './middleware/arrow';
```
atomiks, line 3: `{type X}` syntax is TS 4.5+, but this library only requires a minimum of 4.1

atomiks:
Got an error with this PR with `npm run build`, and it told me to delete `node_modules` + `package-lock.json`:

> @floating-ui/utils:build: > @floating-ui/utils@0.1.6 build
@floating-ui/utils:build: > rollup -c
@floating-ui/utils:build: 
@floating-ui/utils:build: /Users/sites/floating-ui/node_modules/rollup/dist/native.js:87
@floating-ui/utils:build:               throw new Error(
@floating-ui/utils:build:               ^
@floating-ui/utils:build: 
@floating-ui/utils:build: Error: Cannot find module @rollup/rollup-darwin-arm64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.

When I did that it worked mostly, but then got this error:

> extension:build: [vite]: Rollup failed to resolve import "@fluentui/react-provider" from "/Users/sites/floating-ui/extension/node_modules/@fluentui/react-components/lib/index.js".
extension:build: This is most likely unintended because it can break your application at runtime.
extension:build: If you do want to externalize this module explicitly add it to
extension:build: `build.rollupOptions.external`
extension:build: error during build:

bsunderhus:
> Got an error with this PR with `npm run build`, and it told me to delete `node_modules` + `package-lock.json`:
> 
> > @floating-ui/utils:build: > @floating-ui/utils@0.1.6 build
> > @floating-ui/utils:build: > rollup -c
> > @floating-ui/utils:build:
> > @floating-ui/utils:build: /Users/sites/floating-ui/node_modules/rollup/dist/native.js:87
> > @floating-ui/utils:build:               throw new Error(
> > @floating-ui/utils:build:               ^
> > @floating-ui/utils:build:
> > @floating-ui/utils:build: Error: Cannot find module @rollup/rollup-darwin-arm64. npm has a bug related to optional dependencies ([npm/cli#4828](https://github.com/npm/cli/issues/4828)). Please try `npm i` again after removing both package-lock.json and node_modules directory.
> 
> When I did that it worked mostly, but then got this error:
> 
> > extension:build: [vite]: Rollup failed to resolve import "@fluentui/react-provider" from "/Users/sites/floating-ui/extension/node_modules/@fluentui/react-components/lib/index.js".
> > extension:build: This is most likely unintended because it can break your application at runtime.
> > extension:build: If you do want to externalize this module explicitly add it to
> > extension:build: `build.rollupOptions.external`
> > extension:build: error during build:

What version of npm are you using?

I upgraded the lockfile to version 3 https://github.com/floating-ui/floating-ui/pull/2677/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519 , I think it requires npm v10, anyway. I'll include in this PR the migration to pnpm also, this will solve this

packages/core/src/index.ts
bsunderhus: It doesn't matter, type exports use the classic export syntax. but I'll see to use the old syntax, perhaps also add eslint rule to forbid this syntax then

bsunderhus: Ok, I just added support for https://typescript-eslint.io/rules/consistent-type-exports/ in the eslint config, also modified it a little bit to follow the new tsconfig paths

bsunderhus:
Just confirmed on [changeset documentation](https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish), seems like `changeset publish` should work fine with `pnpm`

atomiks:
> Flattens all files into src folders (in utils for example, files were scattered between src, dom and react folders, with the new defineRollupConfig we can have all the files in src and define different outputs for each of the separate entries, producing floating-ui.utils, floating-ui.utils.dom and floating-ui.utils.react files in a flat structure.

Unfortunately this was actually intentional to prevent breaking old bundlers. It uses Node's directory resolution so it's still required to exist for bundlers that don't support `exports`

i.e. `@floating-ui/utils/dom` resolves the nested `package.json`

atomiks:
> [chore: patch bump every package](https://github.com/floating-ui/floating-ui/pull/2677/commits/b8e7a685a6e5ffdb6726b919a409e6e7a49f5f69)

Also, shouldn't a changeset just be added for each package?
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. If these changes should result in a version bump, you need to add a changeset.

bsunderhus:
> > [chore: patch bump every package](https://github.com/floating-ui/floating-ui/pull/2677/commits/b8e7a685a6e5ffdb6726b919a409e6e7a49f5f69)
> 
> Also, shouldn't a changeset just be added for each package?
> 
> > Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. If these changes should result in a version bump, you need to add a changeset.

I see, I'm actually new to change set! I'll add one following the [documentation](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

bsunderhus:
> > Flattens all files into src folders (in utils for example, files were scattered between src, dom and react folders, with the new defineRollupConfig we can have all the files in src and define different outputs for each of the separate entries, producing floating-ui.utils, floating-ui.utils.dom and floating-ui.utils.react files in a flat structure.
> 
> Unfortunately this was actually intentional to prevent breaking old bundlers. It uses Node's directory resolution so it's still required to exist for bundlers that don't support `exports`
> 
> i.e. `@floating-ui/utils/dom` resolves the nested `package.json`

Gotcha 👍🏻, I'll convert this back to a draft and look into resolving the structure.

bsunderhus:
> > > Flattens all files into src folders (in utils for example, files were scattered between src, dom and react folders, with the new defineRollupConfig we can have all the files in src and define different outputs for each of the separate entries, producing floating-ui.utils, floating-ui.utils.dom and floating-ui.utils.react files in a flat structure.
> > 
> > 
> > Unfortunately this was actually intentional to prevent breaking old bundlers. It uses Node's directory resolution so it's still required to exist for bundlers that don't support `exports`
> > i.e. `@floating-ui/utils/dom` resolves the nested `package.json`
> 
> Gotcha 👍🏻, I'll convert this back to a draft and look into resolving the structure.

As an alternative solution, I'm keeping the structure flat and I'm adding a `prepack` command to copy the necessary files from `dist` into its respective folders, it looks like this one you run `prepack`
image

atomiks:
Seems to be mostly there, but I'm getting `publint` errors about missing `.d.ts` for the nested packages?

Seem to get an error with `prepack` for the `react` and `utils` packages:

```
Error: 
    Missing either:
    ./dist/floating-ui.utils.dom.umd.js
    ./dist/floating-ui.utils.dom.esm.js
    ./dist/floating-ui.utils.dom.d.ts
    Did you forget to build?
```

The `react` package also has `dist/index.{hash}.{js/css/html}` files for some reason

bsunderhus:
> Seems to be mostly there, but I'm getting `publint` errors about missing `.d.ts` for the nested packages?
> 
> Seem to get an error with `prepack` for the `react` and `utils` packages:
> 
> ```
> Error: 
>     Missing either:
>     ./dist/floating-ui.utils.dom.umd.js
>     ./dist/floating-ui.utils.dom.esm.js
>     ./dist/floating-ui.utils.dom.d.ts
>     Did you forget to build?
> ```

Yeah, I missed to link the dependency between `publint` and `prepack` command, now running `pnpm run publint` should invoke `prepack` and `prepack` should then invoke both `build` and `build:api`

Two things worth noting:

1. **Always**  invoke the commands using `turbo`, otherwise the commands won't follow dependency chain,
To make things easier the root of the root package has invocations to turbo as scripts, so if you wanna run publint, run `pnpm run publint` from the root
2. There's a difference between `pnpm run publint --filter @floating-ui/core` and `pnpm --filter @floating-ui/core run publint`. The first one will use `turbo` and invoke publint chain of dependants commands only related to the core package while the second one will only invoke the command `publint` from the core package without running the chain of dependants commands



> The `react` package also has `dist/index.{hash}.{js/css/html}` files for some reason

That's odd, I don't see any `dist/index.{hash}.{js/css/html}`, where's that? 👀

atomiks:
It looks like Vite is building some files and placing them in `dist` upon build:

> extension:build: vite v4.5.1 building for production...
extension:build: transforming...
extension:build: ✓ 1669 modules transformed.
extension:build: rendering chunks...
extension:build: computing gzip size...
extension:build: dist/index.html                                  0.56 kB │ gzip:  0.32 kB
extension:build: dist/assets/index-660304c7.css                   1.59 kB │ gzip:  0.55 kB
extension:build: dist/assets/FloatingUIMiddleware-422be482.css    1.65 kB │ gzip:  0.61 kB
extension:build: dist/assets/FloatingUIMiddleware-abf140b2.js     1.73 kB │ gzip:  0.75 kB
extension:build: dist/assets/FluentUIMiddleware-0f0133f8.js       4.18 kB │ gzip:  1.48 kB
extension:build: dist/assets/main-c0ea7fb6.js                   117.51 kB │ gzip: 34.60 kB
extension:build: dist/assets/index-2a60272c.js                  308.60 kB │ gzip: 92.75 kB
extension:build: ✓ built in 3.07s

When I run in the root:

```
pnpm run build
pnpm run build:api
pnpm run prepack
```

I still get:

```
@floating-ui/utils:prepack:     ./dist/floating-ui.utils.dom.umd.js
@floating-ui/utils:prepack:     ./dist/floating-ui.utils.dom.esm.js
@floating-ui/utils:prepack:     ./dist/floating-ui.utils.dom.d.ts
@floating-ui/utils:prepack:     Did you forget to build?
```

bsunderhus:
I think you might have some cache problem with. To ignore cache just add [`--force`](https://turbo.build/repo/docs/reference/command-line-reference/run#--force), have you tried this?!

Can you try running:

```bash
# clean first
pnpm run clean
# execute prepack from root without cache
pnpm run prepack --force
```

atomiks:
Ran that, but it didn't seem to help :\ 

I guess if it works in CI for releases then it doesn't really matter...

```
@floating-ui/dom:build:api: 
@floating-ui/dom:build:api: api-extractor 7.39.0  - https://api-extractor.com/
@floating-ui/dom:build:api: 
@floating-ui/utils:prepack: file:///Users/sites/floating-ui/config/scripts/prepack.mjs:51
@floating-ui/utils:prepack:     throw new Error(`
@floating-ui/utils:prepack:           ^
@floating-ui/utils:prepack: 
@floating-ui/utils:prepack: Error: 
@floating-ui/utils:prepack:     Missing either:
@floating-ui/utils:prepack:     ./dist/floating-ui.utils.dom.umd.js
@floating-ui/utils:prepack:     ./dist/floating-ui.utils.dom.esm.js
@floating-ui/utils:prepack:     ./dist/floating-ui.utils.dom.d.ts
@floating-ui/utils:prepack:     Did you forget to build?
@floating-ui/utils:prepack:     
@floating-ui/utils:prepack:     at assertExportFilesExists (file:///Users/sites/floating-ui/config/scripts/prepack.mjs:51:11)
@floating-ui/utils:prepack:     at async file:///Users/sites/floating-ui/config/scripts/prepack.mjs:22:3
@floating-ui/utils:prepack: 
@floating-ui/utils:prepack: Node.js v20.10.0
@floating-ui/utils:prepack:  ELIFECYCLE  Command failed with exit code 1.
```

bsunderhus:
> @floating-ui/utils:prepack: Node.js v20.10.0

I've been running local and in the CI with Node v18 btw @atomiks, I'll try later on with v20. this might be the problem?! Let's not merge this until this is figured, I'll convert back to draft and investigate more

atomiks:
Also tried on `Node v18.19.0` and it has the same issue. Alright 👍

bsunderhus:
Well, I'm clueless of what might be happening then 😅, maybe you can give me more logs?! Or some other info

atomiks:
Only the type file `./dist/floating-ui.utils.dom.d.ts` seems to be missing for the nested packages, rather than the actual modules. Maybe something wrong with the API extraction there?

bsunderhus:
Thx for the full log!

I can see that the API extractor ran only for `api-extractor.json` and ignored `api-extractor.dom.json` and `api-extractor.react.json` on the utils `build:api` command, are you sure it's only missing `./dist/floating-ui.utils.dom.d.ts`? should be missing also `./dist/floating-ui.utils.react.d.ts` according to this log 🤔

So the `build:api` command on `@floating-ui/utils` is `build-api --tsc tsconfig.lib.json --aec api-extractor{,.dom,.react}.json`, I'm guessing that there's something going on wrong in your environment with this glob: `api-extractor{,.dom,.react}.json`?! And it cannot locate the respective 3 API extractor configuration files.

Ok, lets try to do some test cases:

1. try modifying the command to `build-api --tsc tsconfig.lib.json --aec api-extractor.dom.json` just to verify if that will generate only the `./dist/floating-ui.utils.dom.d.ts`
2. try modifying the command to `build-api --tsc tsconfig.lib.json --aec api-extractor.react.json` just to verify if that will generate only the `./dist/floating-ui.utils.react.d.ts`

both test cases should work to confirm we got a problem going on with the glob `api-extractor{,.dom,.react}.json` in your environment. 

In the meantime I'm going to look for alternatives for that glob pattern

bsunderhus:
Ok, I just added support for passing multiple arguments if you want to try it instead of using globs.

You can modify

```bash
build-api --tsc tsconfig.lib.json --aec api-extractor{,.dom,.react}.json
```

To

```bash
build-api --tsc tsconfig.lib.json --aec api-extractor.json --aec api-extractor.dom.json --aec api-extractor.react.json
```

on `utils/package.json` `build:api` command and it should work the same but without using globs

atomiks:
I can confirm that works 👍 (I added it to the `react` package too, as you mentioned)

bsunderhus:
Great! I've updated the respective `package.json` files to stop using glob then 👍🏼

> Would be nice if you could investigate why glob is not working in your environment though, the glob syntax is way nicer 👀

I guess the PR is ready for review once again 😁.

I verified distribution files and I can confirm that there's no changes being produced, so this PR is mainly only restructuring the repository without making any actual modifications to the dist files, I'd suggest you go through each `dist` file and verify them to ensure everything is in order 🫡.

I'm also bumping a patch version for every single package, as they all have new `.d.mts` files being exported

atomiks:
Thank you 🙏 

However I think there's one remaining issue. The nested packages don't have a `package.json` upon prepack, so the resolution of the packages doesn't actually work for old bundlers.

For e.g. `utils/dom`:

```
dist/
  floating-ui.utils.dom.umd.js
  floating-ui.utils.dom.esm.js
  floating-ui.utils.dom.d.ts
package.json
```

The existing one:

Screenshot 2024-01-03 at 10 47 18 am

atomiks:
Wait, I ran the wrong command lol. Looks fine actually. Going to merge now

atomiks:
@bsunderhus I think there may be an issue with the wrapped DOM types and `index.test-d.ts` files. The `// @ts-expect-error` comments seem to be incorrect now, as e.g. the `boundary` type is `any` now. It seems the `typecheck` command is now missing (no `tsc` check on the whole repo)

bsunderhus:
> @bsunderhus I think there may be an issue with the wrapped DOM types and `index.test-d.ts` files. The `// @ts-expect-error` comments seem to be incorrect now, as e.g. the `boundary` type is `any` now. It seems the `typecheck` command is now missing (no `tsc` check on the whole repo)

Gotcha, I'll investigate and follow up with another PR

bsunderhus:
I'll  be adding here a TODO list with what I believe it would be a nice follow up to further improve the structure.

atomiks:
@bsunderhus it looks like this PR has caused a plethora of `act` warnings to start appearing in the tests. It looks like `Composite.test.tsx` had an issue before, but the other tests were okay. Now it seems all of them have issues.

Possibly an issue with a dependency?

This is used after floating element renders to avoid `act` warnings, but maybe it no longer works in some new version somewhere?:
```js
await act(async () => {});
```

atomiks:
Opened https://github.com/floating-ui/floating-ui/pull/2699

PR #2679, opened 2023-12-29
fix(FloatingArrow): remove placement gap

Sangminnn:
First, i appreciate your good library :)

when i use FloatingArrow for using custom tooltip in my enterprise service, i found a gap between the arrow and the tooltip body.

Actually, It looked the same when I checked in the codesandbox in the official document. 
(https://codesandbox.io/s/floating-ui-react-scale-transform-origin-qv0t1c?file=/src/App.tsx)

You need to zoom in to find it and I'll attach the image file. :)

when i test Approximately 0.5 px was required to reduce the spacing.

If it's your intent, you can ignore the issue.

thank you 👍 

placement

### DIFF
--- a/packages/react/src/components/FloatingArrow.tsx
+++ b/packages/react/src/components/FloatingArrow.tsx
@@ -111,6 +111,8 @@ export const FloatingArrow = React.forwardRef(function FloatingArrow(
   const arrowX = arrow?.x != null ? staticOffset || arrow.x : '';
   const arrowY = arrow?.y != null ? staticOffset || arrow.y : '';
 
+  const placementGap = 0.5;
+
   const dValue =
     d ||
     'M0,0' +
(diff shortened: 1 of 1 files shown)


atomiks:
Unfortunately the rendering is dependent on the device/display. In the example on my computer, I don't see a 0.5px gap. When adding this code, transparency overlaps become a problem:

Screenshot 2023-12-30 at 12 27 42 pm

I don't know of a way to robustly handle this issue in a way that avoids such transparency issues. However, this workaround should work for this case:

```js
<FloatingArrow style={{ transform: 'translateY(-0.5px)' }} />
```

Sangminnn:
Thank you for your comment :)

Actually, i didn't check the transparent issue.

i found a solution about this problem( double opacity in transparency case ) for a few days, but i didn't find that.

Also, I thought about how to use tricks using js, but I thought it was over-engineering.

so, i close that PR.

Thank you :) 👍

PR #2687, opened 2024-01-03
chore: adds typecheck command to verify types of every package

bsunderhus:
Follow up on https://github.com/floating-ui/floating-ui/pull/2677#issuecomment-1874861287

1. adds `typecheck` command to run `tsc -b` for every single package
2. move `*-d.ts{,x}` files to `test` folders to be included in types verification
3. modifies CI gh action to use `pnpm run typecheck` instead of `pnpm run build:api`
4. redeclares middlewares from `core` at `dom` [with more specific types](https://github.com/floating-ui/floating-ui/pull/2687/files#diff-df833d72712902a450cc9e36309797cbe2109bd425ada294461da6229cce1882) (this was mistakenly removed by https://github.com/floating-ui/floating-ui/pull/2677/files#diff-b5f898db2c0f33da655f0da10d2aa08eb272bd34c63b177adb3a2b38157de73f)

## Remaining problems

Seems like types for the `extension` package stories are having some conflict problems between multiple vite versions (v4 and v5), for now I've omitted stories from the typechecking, I'll investigate more later on

### DIFF
--- a/.github/actions/typescript/action.yml
+++ b/.github/actions/typescript/action.yml
@@ -5,5 +5,5 @@ runs:
   steps:
     - run: pnpm run lint
       shell: sh
-    - run: pnpm run build:api
+    - run: pnpm run typecheck
       shell: sh
--- a/extension/package.json
+++ b/extension/package.json
@@ -9,7 +9,8 @@
     "clean": "rimraf dist out-tsc",
     "storybook": "storybook dev",
     "build:storybook": "storybook build",
-    "lint": "eslint src"
+    "lint": "eslint src",
+    "typecheck": "tsc -b tsconfig.lib.json"
   },
   "dependencies": {
     "@floating-ui/devtools": "workspace:*",
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "publint": "turbo publint",
     "test": "turbo test",
     "prepack": "turbo prepack",
+    "typecheck": "turbo typecheck",
     "release": "turbo publint && changeset publish"
   },
   "author": "",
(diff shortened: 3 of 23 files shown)


atomiks:
`dom/index.test-d.ts` hasn't been moved, that's the one with the type errors that don't seem to go away when importing from `src` 🤔

bsunderhus:
> `dom/index.test-d.ts` hasn't been moved, that's the one with the type errors that don't seem to go away when importing from `src` 🤔
> 
> Specifically, [these declarations](https://github.com/floating-ui/floating-ui/pull/2677/files#diff-b5f898db2c0f33da655f0da10d2aa08eb272bd34c63b177adb3a2b38157de73fL162-L187) are necessary since the DOM types need to be specified
> 
> Screenshot 2024-01-03 at 8 19 21 pm

Oh I see, so the types are redefined from `@floating-ui/core` you're not just re-exporting it! my bad 🙈, I'll go over the redefinitions and check one by one

bsunderhus:
Is there any other type redeclaration I deleted by mistkake? @atomiks or just the ones from `@floating-ui/dom`?

atomiks:
It should only be the `dom` package, which wraps `core`'s "any" types for `react-dom`, `react` and `vue`. `react-native` has "any" types since it uses `core`

atomiks:
@bsunderhus i'm curious, is the built `out-tsc` dir meant to be published to npm?

bsunderhus:
It's not. It's only required for api extractor to evaluate types, the types thar are meant to be in the distribution will be under `dist` folder

atomiks:
Alright, I noticed they're in the `"files"` array for the `dom` package using `npm publish --dry-run`. Think it was accidentally missed

PR #2690, opened 2024-01-03
chore: update dependencies version

bsunderhus:
This PR is the result of running `pnpm update -r` which bumps all dependencies version.

Additionally:

1. fix inline imports that weren't caught before
2. adds `.eslintignore` to `@floating-ui/utils` to ignore `dom` folder
3. adds `.eslintignore` to `@floating-ui/react` to ignore `utils` folder


bsunderhus:
This is just a small cleanup after the migration, to ensure all dependencies versions are up to date without any major bumps, of course

PR #2676, opened 2023-12-27
feat(useListNavigation, Composite): support grid navigation over items with variable sizes

JavierM42:
`useListNavigation` and `Composite` both assume that grids will be uniform (all items span 1 row and 1 column). We came across the need for items with variable sizes and decided to implement it.

If it's a welcome feature, we'd be happy to see it added to the library.

With the proposed changes, `useListNavigation` and `Composite` accept would two new optional props:
  - `itemSizes`: an array of `{ width: number; height: number }`, which represent the row and column spans of each item (defaults to all items being 1x1).
  - `dense`: a boolean flag indicating if the grid follows [CSS grid-auto-flow](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-flow) `dense` placement rules or not.
  
The navigation logic is quite complex but I hope it's commented and tested with enough clarity.
  
## Demo video:

https://github.com/floating-ui/floating-ui/assets/32175149/e8b356e2-1ffe-4ced-88f3-7afedb0c77c1

---

Huge thanks to @dantevvp who originally needed this feature and then helped so much to implement it.

Issue #2666, opened 2023-12-18
useFocus opens incorrectly on click/pointer in react-testing-library for visibleOnly


cjp-psl:
When executing inside react @testing-library (jest) with *useFocus* hook, float becomes incorrectly visible on .click() or .pointer() user actions.

This works as expected in browser.

**Steps to reproduce the behavior:**

* Create floating ui react component usage with only *useFocus* hook and button
* Create jest test using @testing-library, and trigger button click via testing library user actions 
* Check if float has been opened in test

Couldn't get a codesandbox for @testing-library running unfortunately due to recent codesandbox changes, but have created an isolated sample 
[FocusTestBug.test.tsx.txt](https://github.com/floating-ui/floating-ui/files/13706565/FocusTestBug.test.tsx.txt)

**Expected behaviour:**
Float is not shown in jest execution context.

**Investigation:**

I've tracked this to a couple of linked causes:

1. the target.match(':focus-visible') call (ln 129) throws an exception, as jsdom does not support that selector (and doesn't seem like they will anytime soon from [this](https://github.com/jsdom/jsdom/issues/3426) issues activity)

2. the pointer event generated by jsdom in @testing-library has no width or height prop set, so passes the *isVirtualPointerEvent* check in *pointerDown* (ln 112)
That in turn prevents keyboardModality from being set to false, which leads to visibleOnly fallback behaviour failing to prevent open event trigger in *onFocus*

I appreciate this isn't strictly speaking a floating-ui problem as browser behaviour is correct, but as component testing via jest and @testing-library is pretty common I wonder if this could be resolved by adjusting the *isVirtualPointerEvent* to return false if the user agent contains jsdom (like with similar android/safari checks already). The fallback behaviour would then work as expected.

For reference navigator.userAgent in jest execution reports for me as 
"Mozilla/5.0 (win32) AppleWebKit/537.36 (KHTML, like Gecko) jsdom/20.0.3"

atomiks:
> the pointer event generated by jsdom in https://github.com/testing-library has no width or height prop set, so passes the isVirtualPointerEvent check in pointerDown (ln 112)
That in turn prevents keyboardModality from being set to false, which leads to visibleOnly fallback behaviour failing to prevent open event trigger in onFocus

Interestingly, in latest Chrome on Mac, it seems like `isVirtualPointerEvent` sometimes incorrectly returns `true` when clicking repeatedly. Doesn't seem to happen in Firefox or Safari. 

[...]

PR #2696, opened 2024-01-05
Fix initial keyboard focus on Grid navigation

JavierM42:
As @atomiks pointed out [here](https://github.com/floating-ui/floating-ui/pull/2676#issuecomment-1877993338), initial focus with arrow keys was broken on grid navigations after the variable sizes PR.

Turns out `getCellIndexOfCorner` was called with `index` `-1` and it wasn't prepared for that case. Easy fix.

---

However, when I tried to add a test so this doesn't happen again, I found some inconsistent behavior (possibly a bug). 
I added `<style>*:focus { border: 1px solid red !important }</style>` to the page before recording this video of the `Grid` component on commit `359e5e15` (before the variable sizes were introduced).

https://github.com/floating-ui/floating-ui/assets/32175149/2cd066c7-df33-45c0-80f2-7de965a4d1b1

You can see that when spam-clicking button, initial focus usually lands on the list, but sometimes it ends up in the first element. No idea why.

---

Because of this issue, I could only add a test for arrow down and not this one for arrow up:

```
test('ArrowUp focuses last item', async () => {
  render(<Grid />);

  fireEvent.click(screen.getByRole('button'));
  expect(screen.queryByRole('menu')).toBeInTheDocument();
  fireEvent.keyDown(document, {key: 'ArrowUp'});
  expect(screen.getAllByRole('option')[47]).toHaveFocus();
  cleanup();
});
```

It's worth noting that in the test environment, I ran a repeated test of `500` occurrences and focus always landed on the first element (which makes the test always fail).

### DIFF
--- a/packages/react/src/utils/composite.ts
+++ b/packages/react/src/utils/composite.ts
@@ -275,7 +275,7 @@ export function buildCellMap(sizes: Dimensions[], cols: number, dense: boolean) 
   return [...cellMap];
 }
 
-/** Gets cell index of an item's corner */
+/** Gets cell index of an item's corner or -1 when index is -1. */
 export function getCellIndexOfCorner(
   index: number,
   sizes: Dimensions[],
@@ -283,6 +283,8 @@ export function getCellIndexOfCorner(
   cols: number,
   corner: 'topLeft' | 'topRight' | 'bottomLeft' | 'bottomRight'
 ) {
+  if (index === -1) return -1;
+
   const firstCellIndex = cellMap.indexOf(index);
 
   switch (corner) {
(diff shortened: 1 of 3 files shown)


atomiks:
> You can see that when spam-clicking button, initial focus usually lands on the list, but sometimes it ends up in the first element. No idea why.

It's due to this function. It seems to be a regression in Chrome where `.pressure` sometimes returns `0` instead of `>0` (`0.5`) - but when I updated the browser the issue went away and I can't reproduce it any longer.

That said, this always returns `true` in testing envs, so that would be why it can't be tested.  This issue describes it: https://github.com/floating-ui/floating-ui/issues/2666
https://github.com/floating-ui/floating-ui/blob/672e4585b4e9ff231893f9bdaecaac5b5ed36536/packages/react/src/utils.ts#L90-L104

PR #2711, opened 2024-01-10
chore: add biome configuration file to react and utils workspace

bsunderhus:
Running the command `pnpm run lint` after `pnpm run prepack` will fail as biome is not properly filtering the distribution folders from `packages/utils` and `packages/react`

1. adds biome configuration file to those packages (react, and utils) including their own specific ignore patterns adn extending the main configuration file

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -17,7 +17,12 @@
       "./**/*.json",
       "./**/*.jsonc"
     ],
-    "ignore": ["./**/node_modules", "./**/dist", "./extension/dist"],
+    "ignore": [
+      "./**/node_modules",
+      "./**/dist",
+      "./**/out-tsc",
+      "./extension/dist"
+    ],
     "ignoreUnknown": true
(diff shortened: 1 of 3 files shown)


atomiks:
Thanks! Yeah I noticed that and wasn't sure how to fix (I thought it was fine in CI since the step runs after, but I guess release doesn't?)

bsunderhus:
I got a question though, is this adoption of `biome` a migration to stop using Eslint?! and if so, should we remove eslint configurations from the repo? also, why are you moving from Eslint?!

atomiks:
Yes, it formats and lints simultaneously and is faster/simple to use. It doesn't format CSS or MDX yet though, so I kept Prettier for those ones on the website. Forgot to remove the eslint config file and possibly some comments in certain places after I rebased with the new version I guess, although they don't harm anything I think...

Issue #2665, opened 2023-12-18
[Devtools] show data history


bsunderhus:
#  Show data history

At the moment the panel displays the last available value provided through the middleware chain.

<img width="209" alt="image" src="https://github.com/floating-ui/floating-ui/assets/5483269/bf8e46ea-36f6-46b7-8676-6ab541562c15">

Ideally the panel should display all history of data that has been passed through the middleware chain.

## Actual proposal

[MiddlewareData](https://github.com/floating-ui/floating-ui/blob/e6a42f56be26957c23f451ff829f101ce113fb55/packages/devtools/src/types.ts#L15) should be updated to support multiple serialized data, something like:

```diff
export type MiddlewareMetadata = {
  type: 'middleware';
-   serializedData: MiddlewareData;
+  serializedData: MiddlewareData[];
  references: References;
};
```

PR #2712, opened 2024-01-10
feature(devtools): show data history

bsunderhus:
### root

1. adds `noRestrictedGlobals` to `biome` configuration to restrict the usage of `chrome` global

### @floating-ui/devtools

1. bumps version into a minor
2. ⚠️ BREAKING CHANGE ⚠️ [`Middleware.serializedData`](https://github.com/floating-ui/floating-ui/pull/2712/files#diff-6016dfad79c454ca8329dabb58252bb2f5e89a2f8ff1231cdb1cd7347d1c767dR19) becomes an array of objects instead of an object
3. serialized data injected by `devtools` method will now accumulate a "history" of all serialized data by keeping pushing new data into an array instead of modifying the data reference
4. `devtools` method invokes [`window.postMessage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to communicate with the `extension` about modifications on the serialized data

### extension

1.  bumps version
2. adds `vite-bundle-visualizer` dependency to help visualizing bundles
3. stops using `@fluentui/react-icons` in favor of `react-feather` (which is what is used by floating ui website apparently?!)
4. adds [content script](https://developer.chrome.com/docs/extensions/develop/concepts/content-scripts) to ensure communication between the main inspected window and the extension itself (this allow to know whenever a serialized data has been modified)
5. restructures components to stop using `fluentui` packages in the extension (except for the fluentui panel itself)
6. adds postcss to allow css with css nesting 
7. adss css theming variables
8. splits `devtools` context into `devtools` and `serializedData` to make it easier to reuse on stories examples
9. updates `FloatingUIMiddleware` and `FluentUIMiddleware` to show multiple data history (at the moment this mean there's a [input:range](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) element that allows the user to chose which of the serialized data they want to look at)
image

Fixes https://github.com/floating-ui/floating-ui/issues/2665


bsunderhus:
@atomiks, let me know after running the release pipeline for this! the extension release should be done right after to be sure we have a compatible version available at the chrome extension store

atomiks:
@bsunderhus the GitHub release is sync'd with the npm package release. There will be a Version Packages PR opened (or rather updating the current one) to be merged first

Issue #2691, opened 2024-01-03
Vue HeadlessUI + FloatingUI: Failed to execute 'getComputedStyle' on 'Window'


pbrzosko:
**Describe the bug** 

When using Vue version, I get the following errors:

```
TypeError: Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.
    at getComputedStyle (floating-ui.dom.umd.js:87:31)
    at isOverflowElement (floating-ui.dom.umd.js:55:9)
    at getOverflowAncestors (floating-ui.dom.umd.js:138:57)
    at autoUpdate (floating-ui.dom.umd.js:655:121)
    at attach (floating-ui.vue.umd.js:141:39)
    at callWithErrorHandling (vue.esm-browser.js:1663:22)
    at callWithAsyncErrorHandling (vue.esm-browser.js:1673:21)
    at job (vue.esm-browser.js:3085:17)
    at resetScheduling (vue.esm-browser.js:594:38)
    at triggerEffects (vue.esm-browser.js:638:5)
```

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'defaultView')
    at getDPR (floating-ui.vue.umd.js:38:39)
    at roundByDPR (floating-ui.vue.umd.js:43:17)
    at eval (floating-ui.vue.umd.js:93:20)
    at ReactiveEffect.eval [as fn] (vue.esm-browser.js:1257:48)
    at ReactiveEffect.run (vue.esm-browser.js:507:25)
    at get value [as value] (vue.esm-browser.js:1266:67)
    at triggerComputed (vue.esm-browser.js:528:21)
    at get dirty [as dirty] (vue.esm-browser.js:481:21)
    at instance.update (vue.esm-browser.js:6694:24)
    at callWithErrorHandling (vue.esm-browser.js:1663:36)
```

**To Reproduce**

https://stackblitz.com/edit/vue-tzedys?file=src%2Fcomponents%2FHelloWorld.vue

1. Run the project
2. Inspect console errors

**Expected behavior** 

Project should run without issues.

**Screenshots** 

![image](https://github.com/floating-ui/floating-ui/assets/67122459/c3014a68-dcb4-41f5-b153-1e47472d676f)

**Context:**

- OS: iOS
- Browser chrome
- Version 120

**Additional context**

[...]

PR #2697, opened 2024-01-05
Does not throw when component type reference or floating renders nothing

lozinsky:
Fixes #2691

PR #2741, opened 2024-01-16
fix(vue): change `isComponentPublicInstance` implementation

lozinsky:
https://github.com/floating-ui/floating-ui/issues/2691#issuecomment-1893590794

### DIFF
--- a/.changeset/large-donuts-beam.md
+++ b/.changeset/large-donuts-beam.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/vue": patch
+---
+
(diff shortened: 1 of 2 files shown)


atomiks:
Thank you! Do you have any idea for this one? https://github.com/floating-ui/floating-ui/pull/2697#issuecomment-1887693294 Might be able to fix it as well...

lozinsky:
I think it might be the same problem.

Muratovnik:
Yep. Broblem is fixed. Thank you!

PR #2747, opened 2024-01-17
fix: CSS `:top-layer` elements inside containing blocks

### LINKED ISSUE #1842: [DOM] Support for :top-layer content, e.g. <dialog> and [popup]

atomiks:
Re-do of https://github.com/floating-ui/floating-ui/pull/2351

Fixes #1842

@Westbrook I took most of your tests here. One thing I didn't understand in your latest version is the checks for `isOverTransforms` — the `isContainingBlock` function already implements all those checks 🤔  I guess "over containing blocks" is more accurate than "over transforms", and I think just calling it `topLayer` is enough.


packages/dom/src/middleware/top-layer.ts
```diff
@@ -0,0 +1,92 @@
...
+        diffCoords.x = rect.x + parseFloat(marginInlineStart);
+        diffCoords.y = rect.y + parseFloat(marginBlockStart);
+      }
+    }
+
+    if (onTopLayer && topLayerIsFloating) {
+      return {
+        x: x + diffCoords.x,
+        y: y + diffCoords.y,
+        data: diffCoords,
```
atomiks, line 74: @Westbrook do you actually use this data?

Westbrook: I don't use them, beyond debugging. I mimicked this structure from one of the other plugins (I thought).

Westbrook:
I originally didn't use `isContainingBlock` because something about it wasn't working. We've gotten closer to the same over time, so if you're seeing it work as the same method that would likely be fine.

packages/dom/src/middleware/topLayer.ts
```diff
@@ -0,0 +1,87 @@
...
+  getComputedStyle,
+  getContainingBlock,
+  isContainingBlock,
+  isElement,
+} from '@floating-ui/utils/dom';
+import type {Middleware} from '../types';
+import {unwrapElement} from '../utils/unwrapElement';
+
+const ancestorQueryEventName = '__fluiaq__';
+const topLayerSelectors = [':popover-open', ':open', ':modal'] as const;
```
Westbrook, line 11: I'm fairly certain that `:open` never made it out of Canary releases of browsers in this context, if you're looking for cycles to trim.

Westbrook (commented):
I'll try to test this version with `isContainingBlock` against our test suite in case I captured the issue we had seen with it anywhere.

Neophen:
Amazing work!

atomiks:
So it turns out the collision detection just using the middleware isn't enough :\ it requires passing checks inside some of the default platform methods. This adds +0.25kB minzipped to the bundle, while the external middleware is now only +0.15-0.2 kB on top of that, so separating it doesn't really provide much benefit...

atomiks:
Refactored from middleware to platform method to include collision detection checks in one single export. This also prevents inflating bundle size by default (aside from slight inflation from a few extra checks internally to handle this new condition):

```js
import {platform, topLayer} from '@floating-ui/dom';

computePosition(reference, floating, {
  platform: {
    ...platform,
    topLayer,
  }
});
```

atomiks:
@Westbrook I think I've reduced it to minimum necessary code now. I found some of the checks were not relevant, but that might only be because the tests you wrote last year weren't covering the new edge cases you found. In that case it'd be good to add them if necessary. That said, collision detection wasn't working at all properly with just a middleware, I'm not sure if you were experiencing that issue... it detected the collision too early rather than too late, so that might be why it didn't seem to be an issue since that's not a poor UX issue necessarily.

To test the `topLayer` platform method export, you'll need to build the `dom` package locally and replace the file in your app since it deeply integrates with other parts of the code now.

atomiks:
I've removed the event listener logic and have integrated it directly as a simple fix. The tests all pass. 

It appears to actually be simpler than it seemed to handle... and I could've added this a long time ago if I knew it was only a few lines 😅 . Of course, maybe some extra logic is necessary to handle the other edge cases you found later on, and if so, a PR adding new failing tests for those would be good.

Since if you were trying to add a middleware to counteract this, it will be a minor version.

Issue #2753, opened 2024-01-18
Ignore `website/` dir for `build` command


atomiks:
After the website was added to pnpm workspaces, it runs its own build command which seems to cause issues with `Update Visual Snapshots` action. The website should be ignored for builds.

@bsunderhus

bsunderhus:
I see some possibilities here:

1. change the name of the script from `build` to something more specific, like `build:website` (this is nice as it'll stop the build of website on every single run of  `pnpm run build` command, we could even do the same to stop the build of the `extension` workspace as it's also not part of the library per se...)
2. seems like the `Update Visual Snapshots` runs `playwright:update` we could just filter the `website` workspace on that specific script, `turbo playwright --filter=!website -- -u` (this works but it isn't so nice, as it would only exclude website from that specific command)

atomiks:
@bsunderhus turns out it was a stray `test.only()` I left instead 🤦   but at least this makes the builds faster by not doing unneeded work I guess

PR #2754, opened 2024-01-18
chore(meta): renames website build command to avoid conflict

bsunderhus:
Fixes https://github.com/floating-ui/floating-ui/issues/2753

1. renames `build` command from `/website` and `/extension` to `build:website`


atomiks:
Will update the build command in Netlify

PR #2773, opened 2024-01-29
Do not close the floating element when focus moves inside shadow roots

sventschui:
This fixes a bug where floating elements are closed when the focus is moved from the reference element to any other element inside the reference or tooltip while the whole thing is rendered inside a shadow root. The issue is that the `relatedTarget` property of the focusOut event points to the shadow root instead of the actually focused element.

### DIFF
--- a/packages/react/src/hooks/useFocus.ts
+++ b/packages/react/src/hooks/useFocus.ts
@@ -170,9 +170,13 @@ export function useFocus<RT extends ReferenceType = ReferenceType>(
             // When focusing the reference element (e.g. regular click), then
             // clicking into the floating element, prevent it from hiding.
             // Note: it must be focusable, e.g. `tabindex="-1"`.
+            // We can not rely on relatedTarget to point to the correct element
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/useFocus.ts
```diff
@@ -170,9 +170,13 @@ export function useFocus<RT extends ReferenceType = ReferenceType>(
...
             // clicking into the floating element, prevent it from hiding.
             // Note: it must be focusable, e.g. `tabindex="-1"`.
+            // We can not rely on relatedTarget to point to the correct element
+            // as it will only point to the shadow host of the newly focused element
+            // and not the element that actually has received focus if it is located
+            // inside a shadow root.
             if (
-              contains(refs.floating.current, relatedTarget) ||
-              contains(domReference, relatedTarget) ||
+              contains(refs.floating.current, activeEl) ||
```
atomiks, line 178: I'm assuming `activeEl` is the same as `relatedTarget` here due to the timeout wrapper right?

PR #2785, opened 2024-02-09
Calculate reference element offset relative to offsetParent iframe

jsnajdr:
This PR fixes a positioning bug when there are nested iframes like this:
```js
<html>
  <iframe>
    <div>floating</div>
    <iframe>
      <div>reference</div>
    </iframe>
  </iframe>
</html>
```
The difference from the scenarios that are already tested in the visual iframe tests is:
1. the `floating` and `reference` elements are in different iframes, the `reference` is one level deeper
2. both `floating` and `reference` are inside an iframe, as opposed to being in the top-level document.

Basically, take the existing "Outside" scenario and embed the page in an iframe. Hence the "Outside Embedded" name.

What happens is that the bounding rect of the `reference` embedded has offsets from both parent iframes. But because the outer iframe is an `offsetParent`, its offset should not be applied because it's common for both the `floating` and `reference` element. The `floating` element will be positioned relative to the outer iframe, not relative to the top-level document.

**How to test:**
Look at the "Outside Nested" visual test without the `getBoundingClientRect` patch applied. The positions of the two elements will be wildly different. After the patch, they will be correctly aligned.

**Known problems:**
The `shift({ boundary })` middleware still doesn't work very well. If the boundary is the `rootIFrame` instead of `nestedIFrame`, the offsets will be calculated incorrectly. I suspect this is very closely related to the fix in #2553.


FezVrasta (approved):

atomiks:
Thank you. Several functional tests are failing with the change:
```
    8 failed
      test/functional/iframe.test.ts:22:7 › [inside] 900,900 correctly positioned on bottom with clipping detection 
      test/functional/iframe.test.ts:44:7 › [virtual] 900,900 correctly positioned on bottom with clipping detection 
      test/functional/iframe.test.ts:22:7 › [inside] 1090,900 correctly positioned on bottom with clipping detection 
      test/functional/iframe.test.ts:44:7 › [virtual] 1090,900 correctly positioned on bottom with clipping detection 
      test/functional/iframe.test.ts:22:7 › [inside] 665,900 correctly positioned on bottom with clipping detection 
      test/functional/iframe.test.ts:44:7 › [virtual] 665,900 correctly positioned on bottom with clipping detection 
      test/functional/iframe.test.ts:22:7 › [inside] 865,665 correctly positioned on bottom with clipping detection 
      test/functional/iframe.test.ts:44:7 › [virtual] 865,665 correctly positioned on bottom with clipping detection 
```

Also, would be good to add new tests specifically for this problem as well

jsnajdr:
> would be good to add new tests specifically for this problem as well

I added a new functional test for the `outside-embedded` scenario. But I need help with creating the actual Linux snapshots. I can do only Darwin.

> Several functional tests are failing with the change:

I don't understand why the tests are failing here. When I look at the visual diffs in the artifact file, I don't see any differences. For example, for this failing test:
```
test/functional/iframe.test.ts:22:7 › [inside] 900,900 correctly positioned on bottom with clipping detection
```
the corresponding visual diff looks like this:
inside-900-900-diff
There are some one-pixel differences, but I'd say the test should be passing.

atomiks:
Hmm I see, they seem to be false positives. Might be related to the new `win` check which creates a 1px offset somewhere, but is correct regardless.

To update and add the new snapshot: `Actions` -> `Update Visual Snapshots [workflow on left]` -> `Run workflow [dropdown on right]` -> `Choose this branch` -> `Run workflow [button]`

jsnajdr:
> Run workflow [dropdown on right]

I don't see this option, probably because of insufficient permissions. I can only browse past runs.

atomiks:
I think you need to do it in your own fork, not this repo

jsnajdr:
> I think you need to do it in your own fork, not this repo

In my fork repo there are no actions available to run. According to docs I should be able to _request_ a workflow run on the PR and you'd need to approve it, but I can't find that either. Could you please run the snapshots update workflow for me?

atomiks:
Your branch doesn't appear in this repo to run the workflow with. Other external contributors have been able to update snapshots, such as https://github.com/floating-ui/floating-ui/pull/2561. I'm not sure why the workflow wouldn't appear on your fork

jsnajdr:
I found it: the `master` branch of my fork was many commits behind upstream. After syncing the fork the actions appeared :+1: The snapshot update is now running.

PR #2794, opened 2024-02-13
fix(vue): useFloating types

philippevezina:
Since `useFloating` uses `unwrapElement` which calls `$el` on a Vue component, `useFloating`'s types should accept Vue components as `reference` and `floating`.

Since `T` extended `ReferenceElement` from `@floating-ui/dom`, it would not allow Vue components as `reference`. The same goes for `floating` which used `FloatingElement`.

The `VueReferenceElement` and `VueFloatingElement` types were added to support the Vue supported types.


philippevezina:
After a second glance, it seems like this wasn't an issue.

PR #2798, opened 2024-02-18
fix NaN error when measure params are undefined

ehxxn:


atomiks:
This doesn't actually address the root cause though and will still result in broken positioning. 

Isn't it this Android issue? 

Screenshot 2024-02-22 at 11 08 57 PM

Issue #2791, opened 2024-02-12
Detect `event.target.matches(':hover')` in `useFocus onBlur` handler


aiktb:
**Describe the bug**

I am using `Listbox` component of `headlessui`, in simple words, when I click on `Listbox.Button` which is used to open `Listbox.Options`, the focus will be moved to `Listbox.Options` programmatically, but At this time, the mouse is still on `Listbox.Button`, which will trigger the `onBlur` event defined by `useFocus`, causing the ToolTip to disappear. As the mouse continues to move on `Listbox.Button`, the ToolTip will reappear until the mouse leaves the ToolTip.

This is weird, I hope to be able to add logic similar to the following to `useFocus`:

```ts
onBlur(event) {
   if (event.target.matches(':hover')) {
     return;
   }
}
```

A new option for useFocus can be introduced to solve this issue. If you are willing to provide information on how to name and default value for this option, I would be happy to open a PR for this.

**To Reproduce**

**Screenshots** 

https://github.com/floating-ui/floating-ui/assets/104723527/962add92-e877-49f2-9c5c-cef67673cc2c

**Expected behavior**

**Screenshots** 

https://github.com/floating-ui/floating-ui/assets/104723527/17e38e72-7a5f-43b2-ac32-ede9aa4ed226

**Context:**

- OS: Windows 11
- Browser: Chrome@121.0.6167.161
- Version "@floating-ui/react": "^0.26.7",

atomiks:
Are there cases where you wouldn't want this? Would be better to simply not have an option at all if possible.

aiktb:
> Are there cases where you wouldn't want this? Would be better to simply not have an option at all if possible.

No, you're right, but I don't know if that would break anything.

PR #2809, opened 2024-03-03
fix(useFocus): focus movement causes floating elements in the hover s…

aiktb:
Reproduction:
```bash
git clone https://github.com/aiktb/test-floating-ui.git
cd test-floating-ui
pnpm i
pnpm link --global @floating-ui/react
pnpm dev
```

Close https://github.com/floating-ui/floating-ui/issues/2791


kokovtsev:
Please note that
1. This fix solves the problem of "blur while hovered" but does not fix the opposite "mouseleave while focused" case
2. This `:hover` check should not be done if `useHover` is not used

PR #2826, opened 2024-03-22
feat(FloatingFocusManager): support returning focus to the first focusable descendent of the reference

adammiller:
Updates `<FloatingFocusManager />` to support returning focus to the first focusable descendant of the reference node, when the reference node itself is not focusable. 

This would be beneficial for utilizing patterns like [Popover](https://github.com/floating-ui/floating-ui/blob/master/packages/react/test/visual/components/Popover.tsx) with reference components like a wrapped input (`<div><input /></div>`). 

Related discussion: https://github.com/floating-ui/floating-ui/discussions/2825

### DIFF
--- a/.changeset/cyan-insects-know.md
+++ b/.changeset/cyan-insects-know.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+Enhances the return focus functionality of FloatingFocusManager to return focus to the first focusable descendant of the reference element, if the reference element itself is not focusable
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -39,11 +39,11 @@ function addPreviouslyFocusedElement(element: Element | null) {
     (el) => el.isConnected,
   );
   let targetEl = element;
-  if(targetEl && getNodeName(targetEl) !== 'body' && !isTabbable(targetEl)) {
-	// If the element itself is not tabbable, try to find the first tabbable node inside it.
-    targetEl = tabbable(targetEl, getTabbableOptions())[0];
-  }
   if (targetEl && getNodeName(targetEl) !== 'body') {
+    if (!isTabbable(targetEl, getTabbableOptions())) {
+      // If the element itself is not tabbable, try to find the first tabbable node inside it.
+      targetEl = tabbable(targetEl, getTabbableOptions())[0];
+    }
     previouslyFocusedElements.push(targetEl);
     if (previouslyFocusedElements.length > LIST_LIMIT) {
       previouslyFocusedElements = previouslyFocusedElements.slice(-LIST_LIMIT);
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -54,6 +54,21 @@ function getPreviouslyFocusedElement() {
     .find((el) => el.isConnected);
 }
 
+const focusableElementsQuery = 'a, button:not(:disabled), input:not(:disabled), textarea:not(:disabled), select:not(:disabled), details, [tabindex]:not([tabindex="-1"])';
+
+function isElementFocusable(node: Element) {
+	if (!(node instanceof HTMLElement) || !node.isConnected) {
+		return false;
+	}
+	return node.matches(focusableElementsQuery);
+}
+
+function getFirstFocusableElement(node: Element) {
+	const focusable = Array.from(node.querySelectorAll(focusableElementsQuery))
+	  .filter(isElementFocusable);
+	return focusable[0];
+}
+
 const VisuallyHiddenDismiss = React.forwardRef(function VisuallyHiddenDismiss(
   props: React.ButtonHTMLAttributes<HTMLButtonElement>,
   ref: React.Ref<HTMLButtonElement>,
@@ -482,8 +497,12 @@ export function FloatingFocusManager<RT extends ReferenceType = ReferenceType>(
         addPreviouslyFocusedElement(refs.domReference.current);
       }
 
-      const returnElement = getPreviouslyFocusedElement();
+      let returnElement = getPreviouslyFocusedElement();
 
+	  // if the returnElement is not focusable, attempt to find a focusable element within it's tree
+	  if (returnElement && !isElementFocusable(returnElement)) {
+		returnElement = getFirstFocusableElement(returnElement);
+	  }
       if (
         returnFocusRef.current &&
         !preventReturnFocusRef.current &&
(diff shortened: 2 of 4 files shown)


atomiks:
The solution should actually be simpler. 

In `onOpenChange`, there's this check:

```js
if (reason === 'escape-key' && refs.domReference.current) {
  addPreviouslyFocusedElement(refs.domReference.current);
}
``` 

Since we use `tabbable` as a dependency, you can do:

```js
import {isTabbable} from 'tabbable';

if (
  reason === "escape-key" &&
  refs.domReference.current &&
  isTabbable(refs.domReference.current)
) {
  addPreviouslyFocusedElement(refs.domReference.current);
}
```

Now it will use the actual previously focused element rather than the non-focusable/tabbable reference element.

Furthermore, could you format the code afterwards with Biome? Thank you!

adammiller:
@atomiks made the suggested changes. I had to add a mock for tabbable to the test config, as it started failing some tests when I started checking `isTabbable`.

packages/react/test/unit/setupTests.ts
```diff
@@ -56,3 +56,19 @@ Object.defineProperty(HTMLElement.prototype, 'offsetParent', {
...
+	const isFocusableMock: typeof lib['isFocusable'] = (node, options) => lib.isFocusable(node, { ...options, displayCheck: 'none' });
+	const isTabbableMock: typeof lib['isTabbable'] = (node, options) => lib.isTabbable(node, { ...options, displayCheck: 'none' });
+	return {
+		...lib,
+		tabbable: tabbableMock,
+		focusable: focusableMock,
+		isFocusable: isFocusableMock,
+		isTabbable: isTabbableMock
+	};
+});
```
atomiks, line 74: This whole mock isn't necessary if you use `getTabbableOptions()` (already imported in `FloatingFocusManager`):

`isTabbable(node, getTabbableOptions())`

packages/react/src/components/FloatingFocusManager.tsx
```diff
@@ -38,9 +38,13 @@ function addPreviouslyFocusedElement(element: Element | null) {
...
   );
-
-  if (element && getNodeName(element) !== 'body') {
-    previouslyFocusedElements.push(element);
+  let targetEl = element;
+  if(targetEl && getNodeName(targetEl) !== 'body' && !isTabbable(targetEl)) {
+	// If the element itself is not tabbable, try to find the first tabbable node inside it.
+    targetEl = tabbable(targetEl, getTabbableOptions())[0];
+  }
+  if (targetEl && getNodeName(targetEl) !== 'body') {
```
atomiks, line 46: Extract `targetEl && getNodeName(targetEl) !== 'body'` into a common variable to reuse here

.changeset/cyan-insects-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+Enhances the return focus functionality of FloatingFocusManager to return focus to the first focusable descendant of the reference element, if the reference element itself is not focusable
```
atomiks, line 5:
```suggestion
fix(FloatingFocusManager): return focus to the first tabbable descendant of the reference element if the reference element itself is not focusable
```

atomiks (commented):
Run `npx biome format packages --write` in the root after you've made your changes

adammiller:
> Run `npx biome format packages --write` in the root after you've made your changes

This changes a lot of other files - i've only committed the changes to the files that I'm modifying here to keep the PR focused. Should this be added to scripts and to the readme for contributions? I didn't see any info on how to format the code anywhere.

packages/react/src/components/FloatingFocusManager.tsx
adammiller: Rolling those into a single variable wouldn't prevent me from needing to null check targetEl each time. Restructured it a bit - see if you like that better

packages/react/src/components/FloatingFocusManager.tsx
```diff
@@ -38,9 +38,13 @@ function addPreviouslyFocusedElement(element: Element | null) {
...
     (el) => el.isConnected,
   );
-
-  if (element && getNodeName(element) !== 'body') {
-    previouslyFocusedElements.push(element);
+  let targetEl = element;
+  if (targetEl && getNodeName(targetEl) !== 'body') {
+    if (!isTabbable(targetEl, getTabbableOptions())) {
+      // If the element itself is not tabbable, try to find the first tabbable node inside it.
+      targetEl = tabbable(targetEl, getTabbableOptions())[0];
```
atomiks, line 45: I think this may cause a runtime error if it has no tabbable children (when it gets filtered above, later on)? I'd add a check to see if it's defined first.

atomiks:
> This changes a lot of other files - i've only committed the changes to the files that I'm modifying here to keep the PR focused. Should this be added to scripts and to the readme for contributions? I didn't see any info on how to format the code anywhere.

Yeah it should probably be done on precommit, previously the repo used Prettier which is widely used in the IDE on save itself so, wasn't really necessary

packages/react/src/components/FloatingFocusManager.tsx
adammiller: `tabbable` has a return type of `FocusableElement[]` - i added a fallback to `element` just in case and restructured it a bit.

atomiks: TS settings in this repo don't check for indexed access to save some size/verbosity, but it does have these risks

packages/react/src/components/FloatingFocusManager.tsx
```diff
@@ -38,12 +38,17 @@ function addPreviouslyFocusedElement(element: Element | null) {
...
+    return;
+  }
+  if (!isTabbable(targetEl, getTabbableOptions())) {
+    // If the element itself is not tabbable, try to find the first tabbable node inside it.
+    targetEl = tabbable(targetEl, getTabbableOptions())[0] ?? targetEl;
+  }
+  previouslyFocusedElements.push(targetEl);
+  if (previouslyFocusedElements.length > LIST_LIMIT) {
+    previouslyFocusedElements = previouslyFocusedElements.slice(-LIST_LIMIT);
   }
```
atomiks, line 52:
```suggestion
  let tabbableEl = element;
  if (!tabbableEl || getNodeName(targetEl) === 'body') return;
  if (!isTabbable(targetEl, getTabbableOptions())) {
    const tabbableChild = tabbable(targetEl, getTabbableOptions())[0];
    if (!tabbableChild) return;
    tabbableEl = tabbableChild;
  }
  previouslyFocusedElements.push(targetEl);
  if (previouslyFocusedElements.length > LIST_LIMIT) {
    previouslyFocusedElements = previouslyFocusedElements.slice(-LIST_LIMIT);
  }
```

atomiks: I think this is easier to follow. We don't add the element to the list if it's not tabbable at all here.

Issue #2863, opened 2024-04-16
Duplicate generated type aliases


atomiks:
When hovering over some module definitions in VS Code, you'll see a `_2` indicating a separate type was generated. This happens for `Derivable` and `React` sometimes:

- `Derivable_2`
- `React_2`

```js
(alias) const shift: (options?: ShiftOptions | Derivable_2<ShiftOptions>) => Middleware
```

```js
(alias) function useInteractions(propsList?: Array<ElementProps | void>): {
    getReferenceProps: (userProps?: React_2.HTMLProps<Element>) => Record<string, unknown>;
    getFloatingProps: (userProps?: React_2.HTMLProps<HTMLElement>) => Record<string, unknown>;
    getItemProps: (userProps?: Omit<React_2.HTMLProps<HTMLElement>, 'selected' | 'active'> & ExtendedUserProps) => Record<string, unknown>;
}
```

@bsunderhus maybe you know how this issue could happen with the type generation?

bsunderhus:
This seems to be related to https://github.com/microsoft/rushstack/issues/2479

atomiks:
Ok, I think I know how to fix `Derivable_2` (it needs to use the DOM, not Core type), but I guess we can manually patch `React_2` back to `React` after the types have been generated.

bsunderhus:
> Ok, I think I know how to fix `Derivable_2` (it needs to use the DOM, not Core type), but I guess we can manually patch `React_2` back to `React` after the types have been generated.

Is it necessary though? it still points to the same export which is the react package, and it's not like `React_2` is leaking....

atomiks:
In `Derivable`'s case it has 1 looser property type than it should. As for `React`, it's just unsightly yeah

bsunderhus:
please assign this issue to me @atomiks , I'll follow up with a PR ASAP

bsunderhus:
Ok, it seems like by upgrading `api-extractor` to `7.43.1` solves this issue. I'll create a PR

bsunderhus:
I can see that this strategy of adding a suffix is to avoid conflicts, in the case of `Derivable_2` is because there's a local definition of `Derivable` that makes sense to me. In the case of `React_2` for some reason API extractor is assuming there's a local definition of `React`?! that is weird

PR #2865, opened 2024-04-18
bugfix: replaces React_2 for React from type generation

bsunderhus:
Fixes https://github.com/floating-ui/floating-ui/issues/2863

1. manually replace `React_2` for `React` from type generation at `build-api` script
2. adds `build-api` script as global dependency on turbo configuration
3. bumps `api-extractor` to latest version
4. adds changeset for patching:
   * `@floating-ui/react`
   * `@floating-ui/react-dom`
   * `@floating-ui/react-native`

### DIFF
--- a/config/scripts/build-api.mjs
+++ b/config/scripts/build-api.mjs
@@ -40,6 +40,7 @@ echo(
 echo('');
 
 echo(chalk.cyan(`Running tsc (${chalk.greenBright(tscPaths.join(', '))})`));
+
 await Promise.all(tscPaths.map((tscPath) => $`npx tsc -b ${tscPath}`));
 
 echo(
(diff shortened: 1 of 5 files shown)


config/scripts/build-api.mjs
```diff
@@ -56,6 +58,14 @@ await Promise.all(
...
         configFile.projectFolder ?? '.',
       );
+      echo(
+        chalk.cyan(
+          `Replacing "React_2" for "React" from ${chalk.greenBright(path)}`,
+        ),
+      );
+      const dtsFile = await fs.readFile(path, 'utf-8');
+      await fs.writeFile(path, dtsFile.replace(/React_2/g, 'React'));
+
```
bsunderhus, line 68: Here we're manually replacing "React_2" for "React" on every `.d.ts` file generated

bsunderhus:
Just wanted to make it clear, this is not an ideal solution! we should subscribe to https://github.com/microsoft/rushstack/issues/2479 and once that is fixed on API extractor we can roll this PR back

PR #2889, opened 2024-05-09
Prevent `setIsMounted` Infinite Loop in React 18

louisscruz:
We're upgrading a React application to React 18, and we use `floating-ui`. In doing the upgrade, we encountered an infinite loop that occurs when some elements are hovered over. After doing some digging, we found that the root cause of the problem was:

```ts
function useDelayUnmount(open: boolean, durationMs: number): boolean {
  const [isMounted, setIsMounted] = React.useState(open);

  if (open && !isMounted) {
    setIsMounted(true); <- side effect executes during render
  }
```

We've validated that after wrapping the `setIsMounted` call in a `useLayoutEffect` (or, more accurately, a `useModernLayoutEffect`), the issue ceases to exist.

### DIFF
--- a/packages/react/src/hooks/useTransition.ts
+++ b/packages/react/src/hooks/useTransition.ts
@@ -22,9 +22,11 @@ function execWithArgsOrReturn<Value extends object | undefined, SidePlacement>(
 function useDelayUnmount(open: boolean, durationMs: number): boolean {
   const [isMounted, setIsMounted] = React.useState(open);
 
-  if (open && !isMounted) {
-    setIsMounted(true);
-  }
+  useModernLayoutEffect(() => {
+    if (open) {
+      setIsMounted(true);
+    }
+  }, [open]);
 
   React.useEffect(() => {
     if (!open) {


atomiks:
Calling a setter in render is a valid pattern if it's conditionally guarded, and this hook is used throughout the dev tests locally in this repo where React 18 has been used for 2 years now. 

Do you have a minimal reproduction?

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

louisscruz:
Yeah, we can try to make a minimal reproduction, but there's a chance we won't be able to replicate the exact sequence of events in our repo that lead to this. We use a combination of the following from `@floating-ui/react` in the area that encounters this problem:

```tsx
import {
  useFloating,
  autoUpdate,
  offset,
  flip,
  shift,
  arrow,
  useHover,
  useFocus,
  useDismiss,
  useRole,
  useInteractions,
  useTransitionStatus,
} from '@floating-ui/react';
```

If I had to guess, I'd say it might be some sequence of events where the `isMounted` state changes, which causes the `status` [state to change](https://github.com/louisscruz/floating-ui/blob/7ee1281398dedeb9e851712996d8aab1c4b4f256/packages/react/src/hooks/useTransition.ts#L80) (by the way, that _is_ in a `useLayoutEffect`), which is causing triggering an `open` change to state that we control a few levels higher. I wonder if we're hitting a case related to what React warns against:

> If you update another component’s state during a render, you’ll see an error.

In any case, this has been very difficult to debug, like React's documentation also mentions:

> Although this pattern is more efficient than an Effect, most components shouldn’t need it either. No matter how you do it, adjusting state based on props or other state makes your data flow more difficult to understand and debug.

So, yes, we can attempt a reproduction case, but I think that is going to be quite time consuming, so I really should ask: what's the rationale for updating state during a render like this in this specific case? Is there a reason `useModernLayoutEffect` is used in similar cases elsewhere in the`@floating-ui/react` but not here? Also, is there any advice for projects that experience this at the moment?

atomiks:
> Is there a reason useModernLayoutEffect is used in similar cases elsewhere in the@floating-ui/react but not here? 

It was some edge case with `FloatingPortal` needing it, but I'm not sure if it's necessary anymore. The hook can likely be simplified significantly, for example how I've written a new one here:

https://github.com/mui/base-ui/blob/3a0c40f846f4666ab61af0b72dec4f0b4dd46535/packages/mui-base/src/useTransitionStatus/useTransitionStatus.ts

I plan to rewrite Floating UI's similarly. However, this new one also does the "set state in render" pattern because it's more performant compared to using effects, as the latter commits multiple times.

louisscruz:
[Here is a reproduction](https://codesandbox.io/p/sandbox/floating-ui-infinite-loop-k3y38m?layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clw11q3d400063b6lxn9kv0mz%2522%252C%2522sizes%2522%253A%255B100%252C0%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clw11q3d400023b6lt2y0mkjm%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clw11q3d400033b6lb6qlj4no%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clw11q3d400053b6lxdjjujrr%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B50%252C50%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clw11q3d400023b6lt2y0mkjm%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clw11q3d300013b6lhb96y0bt%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252Fsrc%252Findex.tsx%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%252C%257B%2522id%2522%253A%2522clw1dc2ky00023b6lcrisu4nd%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522initialSelections%2522%253A%255B%257B%2522startLineNumber%2522%253A43%252C%2522startColumn%2522%253A58%252C%2522endLineNumber%2522%253A43%252C%2522endColumn%2522%253A58%257D%255D%252C%2522filepath%2522%253A%2522%252Fsrc%252FApp.tsx%2522%252C%2522state%2522%253A%2522IDLE%2522%257D%255D%252C%2522id%2522%253A%2522clw11q3d400023b6lt2y0mkjm%2522%252C%2522activeTabId%2522%253A%2522clw1dc2ky00023b6lcrisu4nd%2522%257D%252C%2522clw11q3d400053b6lxdjjujrr%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clw11q3d400043b6lir9chmpv%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522UNASSIGNED_PORT%2522%252C%2522port%2522%253A0%252C%2522path%2522%253A%2522%252F%2522%257D%255D%252C%2522id%2522%253A%2522clw11q3d400053b6lxdjjujrr%2522%252C%2522activeTabId%2522%253A%2522clw11q3d400043b6lir9chmpv%2522%257D%252C%2522clw11q3d400033b6lb6qlj4no%2522%253A%257B%2522tabs%2522%253A%255B%255D%252C%2522id%2522%253A%2522clw11q3d400033b6lb6qlj4no%2522%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Afalse%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D). The code sandbox includes three different examples of how this can fail, though there are other variations of failure modes that are also possible.

It appears that the combination of state setter calls in the renders with the use of unstable ref functions that also set state is the root cause. This is very unfortunate, because the stability of a ref isn't necessarily in the control of the client in all cases. Many third-party libraries use either ref merging or unstable refs internally, without any way for consumers applications to control the implementation. Also, [React does not consider unstable refs invalid](https://github.com/facebook/react/issues/27432#issuecomment-1769820109). It's also not clear how one can even reliably prevent refs from being unstable in the first place, as this [open, 4 year old issue](https://github.com/gregberge/react-merge-refs/issues/5) with a very popular ref merging package covers. And wrapping a callback ref in useMemo or useCallback to fix the problem goes against React's guidance:

* "**You should only rely on useCallback as a performance optimization**. If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useCallback back." – [documentation on useCallback](https://react.dev/reference/react/useCallback) (with their emphasis)
* "**You should only rely on useMemo as a performance optimization**. If your code doesn’t work without it, find the underlying problem and fix it first. Then you may add useMemo to improve performance." – [documentation on useMemo](https://react.dev/reference/react/useMemo) (with their emphasis)

It's really unclear how to reconcile all of this. Overall, this seems like a React-level issue. But, regardless, as a consumer of floating-ui who is just trying to upgrade to React 18 , it's unclear how to proceed. How would one go about continuing to use floating-ui if they encounter an issue with it being used simultaneously with some third party package that has refs of the kind discussed above?

atomiks:
Thanks for the repro. 

For migration, I think you should try to ensure all your callback refs are stable, since when they're not it leads to a lot of unnecessary renders anyway which hurts performance. While I don't think that justifies this issue existing, it at least highlights a potential perf issue that can be eradicated by some refactoring.

The idea that you can't rely on the stability of the values with memo still wouldn't make relying on the stability an issue in this case, since it would only discard the values every now and then, not every render which would lead to an infinite loop like it does right now.

I see several options to unblock you here:

- Ensure all callback refs are stable where `useTransitionStatus` is used. This will improve performance as well and is a good thing to refactor.
- Remove `useTransitionStatus` and forgo animations for now or use a different solution
- Use `patch-package` and apply your change for now, temporarily

I don't know why this doesn't cause the issue though? I'm going to need to keep investigating:

```js
function New() {
  const [el, setEl] = useState<HTMLElement | null>(null);
  const [open, setOpen] = useState(false);
  const [mounted, setMounted] = useState(false);

  if (open && !mounted) {
    setMounted(true);
  }

  const {refs, floatingStyles} = useFloating({
    open,
    onOpenChange: setOpen,
  });

  return (
    <div className="grid place-items-center border border-slate-400 rounded lg:w-[40rem] h-[20rem] mb-4">
      <div ref={refs.setReference} onClick={() => setOpen((o) => !o)}>
        <div ref={(node) => setEl(node)} />
        Reference
      </div>
      {open && (
        <div ref={refs.setFloating} style={floatingStyles}>
          Floating
        </div>
      )}
    </div>
  );
}
```

packages/react/src/hooks/useTransition.ts
```diff
@@ -22,9 +22,11 @@ function execWithArgsOrReturn<Value extends object | undefined, SidePlacement>(
...
-    setIsMounted(true);
-  }
+  useModernLayoutEffect(() => {
+    if (open) {
+      setIsMounted(true);
+    }
+  }, [open]);
 
   React.useEffect(() => {
     if (!open) {
```
atomiks, line 32: Actually it looks like this effect needs to be updated to:

```js
useEffect(() => {
  if (!open && isMounted) {
    const timeout = setTimeout(() => setIsMounted(false), durationMs);
    return () => clearTimeout(timeout);
  }
}, [open, isMounted, durationMs]);
```

The `isMounted` guard keeps the "set state in render" working without a loop.

atomiks:
Created https://github.com/floating-ui/floating-ui/pull/2897

PR #2896, opened 2024-05-14
fix(size): clip floating width/height to avoid overflowing outside viewport

### LINKED ISSUE #2895: [Core] `size` middleware not correctly clipping floating element width/height
**Describe the bug**

When the height of a floating element is greater than the boundary (the viewport by default), and if we set the `crossAxis` option to `true` for the `shift` middleware to allow the floating element to overlap the reference element,

the `size` middleware computes a different available height for the `top` placement than for the `bottom` placement, whereas we would expect it to return the same (to avoid overflow).

**To Reproduce**

https://codesandbox.io/p/sandbox/cranky-wood-k9pqcz

1. Go to CodeSandbox via the above link
2. Click the `Button` to open the popover, you can see the maximum height of the popover is `100px`
3. Click the `top` to change the `placement`, now you can see the maximum height of the popover is `160px`, we expect its maximum height to remain `100px`

**Expected behavior**

The `size` middleware should correctly clip the width/height of the floating element to prevent it from overflowing the viewport or boundary.

**Screenshots**

`placement = bottom`
image

`placement = top`
image

aicest:
To fix #2895.

### DIFF
--- a/.changeset/plenty-chefs-thank.md
+++ b/.changeset/plenty-chefs-thank.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/core": patch
+---
+
+fix(size): correctly constrain floating element to avoid overflowing outside viewport with `shift({crossAxis: true})`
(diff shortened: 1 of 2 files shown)


atomiks:
Thank you! I have noticed this issue before, and it looks like the fix was straightforward

PR #2899, opened 2024-05-14
fix(useTransitionStatus): set status to `unmounted`

atomiks:


louisscruz:
@atomiks Thank you for the fix. I've confirmed that this change fixed the issues we were encountering.

PR #2910, opened 2024-05-19
feat(FloatingFocusManager): `restoreFocus` prop

### LINKED ISSUE #2903: When the focused element of a `FloatingFocusManager` is removed from the dom focus escapes to the `body`
While building a Wizard like flow with `FloatingFocusManager` and the Dialog behavior in react, we realized that once the "move to the next screen" button is pressed the focus leaves the Dialog and escapes to the `body` element.

We were able to work around it, but wanted to bring the issue to your attention.

**To Reproduce**
[Code Sandbox](https://codesandbox.io/p/sandbox/loving-goodall-2gt6l7)
[Sandbox Output Page](https://2gt6l7.csb.app/)

1. Go to output page to avoid iframe pain.
1. Enable screen reader.
1. Click on 'Delete Balloon'
1. Click on 'Confirm'
1. Screen reader does not announce that newly titled modal, focus has moved to the `body` (this can also be seen by calling `document.activeElement` in the inspector.

**Expected behavior** 

When an element of a Modal is removed from the active Dom the focus should move to the `dialog`

There is a `useEffect` call which is commented out which shows the expected behavior.  

**Context:**

- OS: Verified on MacOS and Window.
- Browser Chrome, Firefox, Edge
- Version `"@floating-ui/react": "0.26.15"`

atomiks:
Closes #2903

Currently you need to manually restore focus to the nearest element if the focused element was removed (as done in the `Omnibox` local demo), or focus the floating element itself. 

This does it automatically via a boolean prop. I'm not sure if it should be configurable, with the option to focus the entire floating element itself instead; for example, if the whole dialog's content gets replaced with a new screen, that may be preferrable?

@haagmm-ef

### DIFF
--- a/.changeset/selfish-lizards-grow.md
+++ b/.changeset/selfish-lizards-grow.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+feat(FloatingFocusManager): `restoreFocus` prop
--- a/.changeset/selfish-lizards-grow.md
+++ b/.changeset/selfish-lizards-grow.md
@@ -1,5 +1,5 @@
 ---
-"@floating-ui/react": patch
+'@floating-ui/react': patch
 ---
 
-feat(FloatingFocusManager): `restoreFocus` prop
+feat(FloatingFocusManager): `restoreFocus` prop. This enables automatic restoration of focus to the nearest tabbable element if the element that currently has focus inside the floating element is removed from the DOM.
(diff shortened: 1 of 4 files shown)


haagmm-ef (approved):
Thank you for doing this.  I am trying to think of a scenario where you would not want focus to go back to the nearest element or if there are not any tappable up to the dialog.  

The biggest use case I guess is for everyone who already worked around this, where they would be fighting the feature.

PR #2920, opened 2024-05-29
chore: Update vite to ^5.0.13

egardner:


egardner:
👋 Hi there – I've opened a small PR to update `vite` to a slightly later version than what is currently being used.

I don't have much experience using `pnpm`, so please let me know if there is a way to minimize the delta in the lockfile.

atomiks:
Hi, unfortunately, the pnpm lockfile version is `6` (compatible with `pnpm` `v8`). When using pnpm v9, it blows away the lockfile versions of transitive packages entirely for some reason, and I don't really know why. So if you want to install a new Vite, you need to be using pnpm v8 until I get the deps fixed.

egardner:
> Hi, unfortunately, the pnpm lockfile version is `6` (compatible with `pnpm` `v8`). When using pnpm v9, it blows away the lockfile versions of transitive packages entirely for some reason, and I don't really know why. So if you want to install a new Vite, you need to be using pnpm v8 until I get the deps fixed.

Ok – I will downgrade `pnpm` and try this again – hopefully with cleaner results.

atomiks:
New PR seems like it'll work best

PR #2935, opened 2024-06-14
Export FloatingPortalProps

blueputty01:
This is an amazing project, thank you! 

I just noted that `FloatingFocusManagerProps`, `UseFloatingOptions` etc are all exported except for FloatingPortalProps. This PR proposes that.

Please let me know if this doesn't make sense!

### DIFF
--- a/.changeset/nine-flies-accept.md
+++ b/.changeset/nine-flies-accept.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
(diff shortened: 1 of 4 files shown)


atomiks:
Thanks, these also need to be exported from the main `types.ts` file

blueputty01:
> Thanks, these also need to be exported from the main `types.ts` file

Sounds good! I'll try to get this PR updated this weekend.

PR #2948, opened 2024-06-24
refactor: improve types and internal codebase consistency

atomiks:


teleskop150750:
hello. Is there a random `element.floating' missing here?

[link](https://github.com/floating-ui/floating-ui/blob/fe06ea2d4b5d4714602a0d1b7e8844d7b7dc21e5/packages/react/src/hooks/useDismiss.ts#L264)

teleskop150750:
@atomiks 

hello. Is there a random `element.floating' missing here?

[link](https://github.com/floating-ui/floating-ui/blob/fe06ea2d4b5d4714602a0d1b7e8844d7b7dc21e5/packages/react/src/hooks/useDismiss.ts#L264)

atomiks:
That variable is actually unnecessary after [this commit](https://github.com/floating-ui/floating-ui/commit/676f336b93852a89a1d5551725aee6b5827dcf23) since the `floating` is no longer used in the block. So it can just be removed (and likely why it was missed as it didn't error. Now it's always truthy since it refers to the prop getter statement below).

PR #2952, opened 2024-06-27
refactor: use `scrollX`/`scrollY` instead of deprecated `pageXOffset`/`pageYOffset`

Mini-ghost:
The `pageXOffset` and `pageYOffset` properties have been deprecated. In this PR, I have replaced `element.pageXOffset` with `element.scrollX` and `element.pageYOffset` with `element.scrollY`.

The `scrollX` and `scrollY` properties are supported by most modern browsers, with the only exception being Internet Explorer. However, according to the [Floating UI support list](https://floating-ui.com/docs/misc), our project no longer requires support for IE. This allows us to utilize these modern properties.

### DIFF
--- a/.changeset/chatty-pens-cough.md
+++ b/.changeset/chatty-pens-cough.md
@@ -1,0 +1,6 @@
+---
+"@floating-ui/react": patch
+"@floating-ui/utils": patch
+---
+
+refactor: use `scrollX`/`scrollY` instead of deprecated `pageXOffset`/`pageYOffset`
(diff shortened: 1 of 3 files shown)


atomiks:
I've known about this for a while but never bothered changing it because I felt it wasn't worth it in case there are some browser engines/old versions where they somehow _don't_ act as simple aliases. The functional tests all passed but they only test Chromium (for speed reasons) not FF/Safari. I guess we can just hope for the best :-)

Mini-ghost:
I also believe this is a minor adjustment. I initiated this PR because my editor flagged `pageXOffset` and `pageYOffset` while I was reading the source code.

I manually tested `scrollX` and `scrollY` in Safari (17.5) and Firefox (124.0), and they worked fine.

This is a very small change, and there might be some aspects I haven't fully considered. If you ultimately decide it's not suitable, please feel free to close this PR.

PR #2980, opened 2024-07-17
feat(utils): allow reuse of existing computed style in `isContainingBlock`

jcfranco:
This helps avoid recomputing styles when integrating `isContainingBlock` into logic that already has them ([example](https://github.com/jcfranco/composed-offset-position/blob/main/src/index.ts#L71)).

### DIFF
--- a/.changeset/early-emus-destroy.md
+++ b/.changeset/early-emus-destroy.md
@@ -1,5 +1,5 @@
 ---
-"@floating-ui/utils": minor
+"@floating-ui/utils": patch
 ---
 
 feat(getContainingBlock): allow `CSSStyleDeclaration` as an argument
--- a/.changeset/early-emus-destroy.md
+++ b/.changeset/early-emus-destroy.md
@@ -2,4 +2,4 @@
 "@floating-ui/utils": minor
 ---
 
-allow reuse of existing computed style in `isContainingBlock`
+feat(getContainingBlock): allow `CSSStyleDeclaration` as an argument
--- a/.changeset/early-emus-destroy.md
+++ b/.changeset/early-emus-destroy.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/utils": minor
+---
+
+allow reuse of existing computed style in `isContainingBlock`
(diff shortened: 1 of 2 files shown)


packages/utils/src/dom.ts
```diff
@@ -68,9 +68,11 @@ export function isTopLayer(element: Element): boolean {
   });
 }
 
-export function isContainingBlock(element: Element): boolean {
+export function isContainingBlock(element: Element): boolean;
+export function isContainingBlock(css: CSSStyleDeclaration): boolean;
```
atomiks, line 72: Why are these overloads necessary if the return type is always a boolean?

.changeset/early-emus-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@floating-ui/utils": minor
```
atomiks, line 2: Since `utils` is in `v0`, features are added to patches

packages/utils/src/dom.ts
jcfranco: IMO, it leads to cleaner code hints, but I can remove these if you're fine with `elementOrCss: Element | CSSStyleDeclaration)` as the arg + type.

Screenshot 2024-07-17 at 1 23 49 AM

vs

Screenshot 2024-07-17 at 1 23 34 AM

jcfranco:
@atomiks Thanks for the merge!

Considering your [earlier feedback](https://github.com/floating-ui/floating-ui/pull/2980/files/66a3268bb88dc7edcb72d2cfccf6362bed66706a#r1680530795), I think avoiding the function overload is a better approach. This would also follow TS's recommendation [to prefer union type params vs overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#:~:text=Always%20prefer%20parameters%20with%20union%20types%20instead%20of%20overloads%20when%20possible). Is it OK if I update the signature to:

```ts
export function isContainingBlock(elementOrCss: Element | CSSStyleDeclaration): boolean { /* ... * /}
```

?

PR #2982, opened 2024-07-17
fix: replace `left` with `insetInlineStart`

zaaakher:
Currently `floatingStyles` returned from `useFloating` is overwriting placement styles and preventing `-start` to truly adapt to RTL layouts.

See https://github.com/TypeCellOS/BlockNote/pull/922#issuecomment-2233768520

### DIFF
--- a/.changeset/silver-rings-double.md
+++ b/.changeset/silver-rings-double.md
@@ -1,0 +1,5 @@
+---
+'@floating-ui/react-dom': patch
+---
+
+- Fixes `floatingStyles` positioning to use `insetInlineStart` instead of `left`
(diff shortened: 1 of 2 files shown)


atomiks:
This won't work correctly. Looking at the issue you can disable the auto RTL logic in two ways:

- Putting `direction: ltr` styles on the floating element (you can then undo that with `direction: rtl` on a wrapper inside for its content if necessary)
- Overwriting the [`platform.isRTL`](https://floating-ui.com/docs/platform#isrtl) method to always return `false` (for this instance)

zaaakher:
@atomiks ah you're right. It's the `floatingStyles.transform` that's actually positioning it.

PR #2985, opened 2024-07-19
fix: focus always mounted nested lists with listNavigation

michaldudak:
When building a nested menu that's always mounted in the DOM (and hidden with CSS / `aria-hidden`), `useListNavigation` doesn't focus the first item when a submenu is open. See https://codesandbox.io/p/sandbox/focused-silence-m63rxt for an example.

This is because useListNavigation sets the initial focus only if the list hasn't been mounted on the previous render.

In this PR, I added the `previousOpenRef`, so the focusing logic also works for always mounted lists.

I modified the Menu visual test to include always-mounted branches. Let me know if you'd like to keep it the way it was, and I can create a test component directly in the unit test file.

### DIFF
--- a/.changeset/lazy-paws-cry.md
+++ b/.changeset/lazy-paws-cry.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+fix: focus always mounted nested lists with listNavigation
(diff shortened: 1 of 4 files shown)


atomiks:
I remember animations being an issue specifically without conditional rendering with this hook (i.e. `keepMounted`). I'll have to see how this works with transitions, unless you have already?

michaldudak:
I tested it with Base UI's Menu: https://669e0fc0bafb0900081ece78--base-ui.netlify.app/base-ui/react-menu/#nested-menu. It seems to be working well.
Pressing the left arrow key, then immediately the right arrow key when on a submenu correctly returns focus where is should be (which wasn't the case previously, as the parent menu item was still focused even though the submenu was open).

PR #2987, opened 2024-07-23
fix: test if frameElement is readable

chrishoyle:
**Describe the bug** 

When referencing window.frameElement inside of a cross-origin iframe in Safari, MSEdge, and IE, an error is thrown to the console: "Blocked a frame with origin: __ from accessing a frame with origin: __". Since window.frameElement is referred to inside of getBoundingClientRect, this can quickly accumulate a sizable amount of errors in the console. 

**Steps to reproduce the behavior:**

1. go to [codepen.io/jcfranco/pen/LYoOJGe?editors=1010](https://codepen.io/jcfranco/pen/LYoOJGe?editors=1010) in Safari
hover over tooltip buttons (hover over me)
2. See errors in console


atomiks:
Thank you for the fix!

PR #2996, opened 2024-08-06
Arrow: Proper placement when using shift() or autoPlacement()

baylorpaul:
Arrow: Add test components & solution for placement issues with the arrow() middleware when also used with the shift() or autoPlacement() middleware. Specifically, any "top" or "bottom" placement with shift() causes an issue. And for autoPlacement(), any "right" value in "allowedPlacements" causes an issue.

### DIFF
--- a/.changeset/lemon-baboons-search.md
+++ b/.changeset/lemon-baboons-search.md
@@ -1,6 +1,0 @@
----
-"@floating-ui/dom": patch
-"@floating-ui/utils": patch
----
-
-fix: test if `frameElement` is readable to avoid errors in Safari and MSEdge with cross-origin iframes
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @floating-ui/core
 
+## 1.6.6
+
+### Patch Changes
+
+- Update dependencies: `@floating-ui/utils@0.2.6`
+
 ## 1.6.5
 
 ### Patch Changes
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@floating-ui/core",
-  "version": "1.6.5",
+  "version": "1.6.6",
   "description": "Positioning library for floating elements: tooltips, popovers, dropdowns, and more",
   "publishConfig": {
     "access": "public"
--- a/packages/dom/CHANGELOG.md
+++ b/packages/dom/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @floating-ui/dom
 
+## 1.6.9
+
+### Patch Changes
+
+- fix: test if `frameElement` is readable to avoid errors in Safari and MSEdge with cross-origin iframes
+- Update dependencies: `@floating-ui/utils@0.2.6`
+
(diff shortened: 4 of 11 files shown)


atomiks:
Thanks for this - I also noticed this as an issue. I will try to test this out and get back to you.

atomiks:
I tested this out, but I'm not sure I fully understand the issue being fixed.

The issue I noticed was that the `staticOffset` could cause the arrow to not point to anything if the floating element was shifted enough. Is that what this is trying to fix? 

There seems to be a problem here in that the arrow can escape the floating element's boundaries instead of staying in view:

Screenshot 2024-08-24 at 5 55 00 PM

Above, the arrow is clipped on the left and doesn't shift itself to stay within the floating element's boundaries.

baylorpaul:
Thanks for looking at this. That's one aspect of one issue, yes, but not the main thing. There are two issues being resolved by this pull request, both related to arrow(). The first commit is just the test cases to demonstrate the issues. The 2nd commit resolves one issue, and the 3rd commit resolves the other. Try running the first commit by itself to see the issues.

1. When autoplacement() is used with arrow(): When the only placement option is "right", the arrow is moved over to the right too much. Check out this screenshot.
arrow-with-autoplacement-right

2. When shift() is used with arrow(): If the floating content is wide enough that a "shift" becomes necessary, when the "side" is "top" or "bottom", the arrow is misplaced in various ways. Check out this screenshot. The `staticOffset` is not the main thing being addressed, but comes into play. If that wasn't addressed, this pull request would introduce a different bug. The `staticOffset` is not relevant when the content becomes so wide that it is shifted, and using it when the content is shifted would also misplace the arrow, so in my commits, it is ignored when a shift is necessary.
arrow-with-shift-when-content-is-wide

Both of those issues should be resolved with the pull request.

atomiks:
1. This is not actually a bug, it's caused by placing `autoPlacement` _after_ `arrow` inside the middleware array. `autoPlacement` needs to be placed before `arrow`. This is not really explicitly documented since the ordering can be anything, though the doc mentions arrow should go "towards the end": https://floating-ui.com/docs/arrow#order
2.  This should still be valid though. However, this PR's implementation doesn't account for the issue I mentioned above, where it won't stay inside the floating element.

baylorpaul:
Gotcha - thanks! Yes, I did not realize ordering was important, and the arrow() needed to be at/near the end. But I see you have documented that.

With your latest changes, once I put arrow() at the end, I no longer have any issues.

Two other comments:
1. Do you think you could also apply the changes I submitted for FloatingArrow.mdx and the test changes from Arrow.tsx? In Arrow.tsx:42, one change to my submission would be necessary: the `...(middleware ?? [])` should be placed before the `arrow()`.
2. In FloatingArrow.tsx:157, you may want to do `transform: [rotation, transform].filter(t => !!t).join(' '),` instead. If both transforms exist, and the transforms don't have precisely one space between them, they won't apply.

atomiks:
Sure, you can edit your PR with just those changes and it could be merged

baylorpaul:
I've added the test components & documentation portion at https://github.com/floating-ui/floating-ui/pull/3023

PR #3016, opened 2024-08-21
fix(useHover): `blockPointerEvents` on `ownerDocument` instead of `document`

### LINKED ISSUE #3013: `blockPointerEvents` applies stray `pointer-events: none` to other documents/`<body>`
We hope to use [`blockPointerEvents`](https://floating-ui.com/docs/usehover#blockpointerevents) in our dropdown menu. Some of these dropdown menus are in the main window and some are in popout windows.

When we hover over menu items in the dropdown menu in the *popout window*, `pointer-events: none` is applied to the `<body>` in the popout window and removed correctly. But a stray `pointer-events: none` is also applied to the `<body>` in the *main window* that is never removed. So, users can no longer click anything in the main window anymore unless they refresh the page.

**To Reproduce**

[Sandbox](https://codesandbox.io/p/sandbox/autumn-cookies-q66468?workspaceId=3031d7ee-f21d-4739-ab65-daef77c0c8dd) that builds upon the [DropdownMenu example](https://codesandbox.io/s/admiring-lamport-5wt3yg?file=/src/DropdownMenu.tsx) in the [docs](https://floating-ui.com/docs/react-examples#other). I left the original `DropdownMenu.tsx` nearly the same except to add the [`root`](https://floating-ui.com/docs/floatingportal#root) to the `FloatingPortal`.

1. Click "Edit"
2. Hover over "Copy as".
  1. Notice a `pointer-events: none` added to the `<body>`
3. Hover out
  1. Notice that the `pointer-events` style is removed (✅)
4. Click "Toggle new window" to open a popout window
5. Click "Edit"
6. Hover over "Copy as".
  1. Notice a `pointer-events: none` added to the `<body>` of the popout window and the main document
7. Hover out
  2. Notice that only the `pointer-events` from the popout window's body is removed (✅). The `pointer-events` from the main document is never removed (❌).
7. Try clicking anything in the main document. No clicks would work any longer.

**Expected behavior** : No stray `pointer-events: none` should be added. If they are added for some reason, they should be removed when appropriate.

**Screenshots**

https://github.com/user-attachments/assets/0c05d79d-2d43-470a-8cdf-bb532748c7bf

**Context:**

- OS: Mac
- Browser: Edge
- Version: `@floating-ui/react`: 0.26.22 

**Probable cause**:

In the implementation for `blockPointerEvents`, there is a useLayoutEffect that adds `pointer-events: none` to the body.

https://github.com/floating-ui/floating-ui/blob/a8b7a88a73f780c59b89bb8c6e3db16ec589a43d/packages/react/src/hooks/useHover.ts#L374-L376

`getDocument` tries to return the `ownerDocument` but returns the `document` if `elements.floating` is `null`:

https://github.com/floating-ui/floating-ui/blob/a8b7a88a73f780c59b89bb8c6e3db16ec589a43d/packages/react/src/utils.ts#L150-L152

And `elements.floating` is set using the ref prop. E.g.:

```jsx
<div ref={refs.setFloating} />
```

Upon debugging, I could confirm that this ref function (i.e. setFloating) is called *after* the `useLayoutEffect`. Thus, the `useLayoutEffect` first runs with `elements.floating == null`. Consequently, `getDocument` returns `document` instead of `ownerDocument` leading to `pointer-events: none` being applied to the main window's body even when the dropdown menu is in the popout window.

After hovering out of the menu item trigger, `clearPointerEvents` is called.

https://github.com/floating-ui/floating-ui/blob/a8b7a88a73f780c59b89bb8c6e3db16ec589a43d/packages/react/src/hooks/useHover.ts#L208-L215

But this function always has `elements.floating != null`. Thus, only the `pointer-events: none` from the popout window is removed. The one from the main window's body is never removed.

r100-stack:
Fixes #3013.

As suggested by @atomiks in https://github.com/floating-ui/floating-ui/issues/3013#issuecomment-2299935268, I moved the code in `useHover` that adds `pointer-events: none` to the body inside the if-statement that verifies that `elements.floating` is set.

That way, `pointer-events: none` is always set on `ownerDocument` instead of `document`. This prevents `pointer-events: none` from being applied to incorrect `<body>` elements and thus making them unclickable until the page is refreshed.

I also removed the `pointer-events` in the effect's cleanup function.

I confirmed that this PR fixes the issue (#3013) in our app (by making this change in the node_modules).

### DIFF
--- a/packages/react/src/hooks/useHover.ts
+++ b/packages/react/src/hooks/useHover.ts
@@ -371,13 +371,14 @@ export function useHover(
       handleCloseRef.current?.__options.blockPointerEvents &&
       isHoverOpen()
     ) {
-      const body = getDocument(elements.floating).body;
-      body.setAttribute(safePolygonIdentifier, '');
-      body.style.pointerEvents = 'none';
       performedPointerEventsMutationRef.current = true;
       const floatingEl = elements.floating;
 
       if (isElement(elements.domReference) && floatingEl) {
+        const body = getDocument(elements.floating).body;
+        body.setAttribute(safePolygonIdentifier, '');
+        body.style.pointerEvents = 'none';
+
         const ref = elements.domReference as unknown as
           | HTMLElement
           | SVGSVGElement;
--- a/packages/react/src/hooks/useHover.ts
+++ b/packages/react/src/hooks/useHover.ts
@@ -377,7 +377,6 @@ export function useHover(
       if (isElement(elements.domReference) && floatingEl) {
         const body = getDocument(elements.floating).body;
         body.setAttribute(safePolygonIdentifier, '');
-        body.style.pointerEvents = 'none';
 
         const ref = elements.domReference as unknown as
           | HTMLElement
@@ -391,10 +390,12 @@ export function useHover(
           parentFloating.style.pointerEvents = '';
         }
 
+        body.style.pointerEvents = 'none';
         ref.style.pointerEvents = 'auto';
         floatingEl.style.pointerEvents = 'auto';
 
         return () => {
+          body.style.pointerEvents = '';
           ref.style.pointerEvents = '';
           floatingEl.style.pointerEvents = '';
         };
--- a/.changeset/sharp-cheetahs-whisper.md
+++ b/.changeset/sharp-cheetahs-whisper.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(useHover): `blockPointerEvents` no longer adds `pointer-events: none` to unintended `<body>` elements.


packages/react/src/hooks/useHover.ts
```diff
@@ -390,10 +390,12 @@ export function useHover(
           parentFloating.style.pointerEvents = '';
         }
 
+        body.style.pointerEvents = 'none';
         ref.style.pointerEvents = 'auto';
         floatingEl.style.pointerEvents = 'auto';
 
         return () => {
+          body.style.pointerEvents = '';
```
atomiks, line 398: Not sure if this should be added for nested elements? The original code didn't need it at least.

atomiks: I think it's fine actually, since the polygon gets generated anew for nested elements, this doesn't end up affecting them since the cleanup function is not called during traversal of them.

PR #3021, opened 2024-08-25
refactor: `floating-focus-out` -> `focus-out` reason

atomiks:
fyi  @oleksandr-danylchenko I'm changing this because it can close if the reference element lost focus instead (when tabbing backwards to the trigger, then back)


oleksandr-danylchenko (commented):
Thanks, it makes sense!

PR #3024, opened 2024-08-27
fix(offset): avoid NaN when mainAxis or crossAxis is undefined

inokawa:
The coords become NaN when we use `offset` middleware and pass undefined to `option.mainAxis` or `option.crossAxis`. This behavior can cause problem when the option was optional value, like `offset({ mainAxis: offsetOrUndefined })`.


atomiks:
Thank you

PR #3027, opened 2024-08-28
fix(size): fill viewport along an axis if shift is enabled on that axis

### LINKED ISSUE #3011: `size` with `shift` doesn't fill the viewport when resizing
### Discussed in https://github.com/floating-ui/floating-ui/discussions/3005

<div type='discussions-op-text'>

<sup>Originally posted by **madoci** August 14, 2024</sup>

**Describe the bug**

When using `shift` with `crossAxis` set to true, along with `size`, the `availableHeight` (or `availableWidth` if placement is left or right) doesn't increase to the actual available height when the floating element shifts over the reference.

**To Reproduce**

1. Go to https://codesandbox.io/s/modest-dream-dy5ns6?file=/src/index.js
2. Resize the window down until the tooltip hits the top and starts shrinking.
3. Resize the window up.
4. The tooltip moves down, but stays the same size. Its starts growing again when the top hits the bottom of the reference.

**Expected behavior**

When resizing up, the tooltip should regrow to fill the new available space.

**Screenshots**
Capture d’écran du 2024-08-14 15-36-35

**Context:**

- OS: Debian
- Browser: Chrome 127.0.6533.99
- Version: @floating-ui/dom 1.6.10

**Additional context**

I think the issue lies here : 

https://github.com/floating-ui/floating-ui/blob/a8b7a88a73f780c59b89bb8c6e3db16ec589a43d/packages/core/src/middleware/size.ts#L85-L95

The `availableHeight` or `availableWidth` is set to `maximumClippingHeight` when shifting, but only on the main shift axis.
In fact, this code seems overcomplicated to me. The `overflowAvailable<size>` is set before to the min between some value and `maximumClipping<size>`, so `min(overflowAvailable<size>, maximumClipping<size>)` is equivalent to `overflowAvailable<size>`. And since `available<size>` is set before to `overflowAvailable<size>`, the ternary condition essentially means `If alignment or noShift, do nothing. Otherwise, set to maximumClipping<size>`. So the condition could be put in the `if... else...`. It doesn't change much, but it will allow me to simplify my proposal.

The `available<size>` should be set to `maximumClipping<size>` if there is a shift along its axis, even if it's on the placement axis. Something like this would work : 
```js
if (state.middlewareData.shift?.x && !(alignment && isYAxis)) {
  availableHeight = maximumClippingHeight;
}
if (state.middlewareData.shift?.y && !(alignment && !isYAxis)) {
  availableHeight = maximumClippingHeight;
}
```

This works in my case, but I don't know the side effects it can have.

However, during testing on the codesandbox, I noticed that the behaviour also depends on the padding given to `shift` and `size`. The behaviour explained here shows when we use the same padding for both middleware. In the codesanbox, you can clik on the "Show other tooltips" button to display a tooltip where `size` has less padding than `shift`, and one where `size` has more padding than `shift`. With less padding, the tooltip resize almost correctly (it has some weird deferment when approaching its max height). With more padding, the tooltip shifts only when it cannot shrink anymore.

This appears to come from what `detectOverflow` returns. With the code above, all of these cases seems to be fixed, but I don't know if this is a bug of the `detectOverflow` function or if it works as intended.</div>

madoci:
Closes #3011 

A few notes on this PR : 
- To be able to know on which axis the shift middleware is enabled, I added `enabled` to the shift data, an object with `x` and `y` boolean properties, rather than the `mainAxis` and `crossAxis`. I found it easier to use as it can be indexed with an axis, and we don't have to remember that `shift.mainAxis` and `shift.crossAxis` are opposite to usual `mainAxis` and `crossAxis`.
- I added a resize CSS property to the Size test container and a ResizeObserver to be able to test the effect of resizing. It adds a resize handle to the container, hence why I had to regenerate the screenshots. I split the implementation and tests into two different commits to demonstrate that tests are still passing with the fix.


atomiks:
Thanks for your hard work on this. Looks really good

Issue #3017, opened 2024-08-21
isPositioned only becomes true after setting the open option to false


sebbayer:
**Describe the bug**
Looking at the documentation of [Vue Effects](https://floating-ui.com/docs/vue#effects) I understand that positioning is done in an async function and that the `useFloating` composable returns an `isPositioned` boolean ref that lets me know if the floating element has been positioned. I can also provide an `open` option that represents the open/close state of the floating element. So far, so good.

Judging from the linked example in the documentation, I would expect that the `isPositioned` ref would be set to `true` everytime the floating element has been opened, so I can for example focus elements inside it. I need to ensure the positioning is done first before focusing anything to prevent unwanted scrolling behaviour. But `isPositioned` only becomes true after the `open` option has been set to `false`. See my example. This way, I can never determine when the positioning is done after opening a dropdown.

The linked documentation says in the comment:
> Each time the floating element opens, we want to focus and scroll some element into view.

So either I am using `isPositioned` wrong or the library is not working as intended per the documentation.

**To Reproduce**

I forked the `@floating-ui/vue` demo: https://codesandbox.io/p/sandbox/floating-ui-vue-demo-forked-wrzxdx

1. Open the codesandbox link
2. Click on the button multiple times
3. See how `isPositioned` only becomes true after `open` is set to `false`.

**Expected behavior**
I would expect `isPositioned` to be set every time the `open` option updates, not only when it is `false`.

**Context:**

- OS: macOS Sonoma
- Browser Chrome 127
- Version @floating-ui/vue 1.1.3

oleksandr-danylchenko:
Seems like the sandbox is private:
<img width="589" alt="image" src="https://github.com/user-attachments/assets/08f9638c-a084-45d6-8f4b-0d0b9b2d544e">

sebbayer:
Sorry @oleksandr-danylchenko, it should be accessible now.

atomiks:
@lozinsky can you take a look here?

lozinsky:
The issue also applies to the `@floating-ui/react` package, because `useFloating` from `@floating-ui/vue` implements the same behaviour.

Demo: https://codesandbox.io/p/sandbox/floating-ui-react-demo-p5rxyp

[...]

PR #3042, opened 2024-09-10
fix(useFloating): avoid setting `isPositioned` to true when `open` is false

atomiks:
https://github.com/floating-ui/floating-ui/issues/3017#issuecomment-2330517330

### DIFF
--- a/.changeset/flat-pianos-cough.md
+++ b/.changeset/flat-pianos-cough.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react-dom": patch
+---
+
+fix(useFloating): avoid setting `isPositioned` to true when `open` is false
--- a/packages/react-dom/src/useFloating.ts
+++ b/packages/react-dom/src/useFloating.ts
@@ -75,6 +75,7 @@ export function useFloating<RT extends ReferenceType = ReferenceType>(
   const hasWhileElementsMounted = whileElementsMounted != null;
   const whileElementsMountedRef = useLatestRef(whileElementsMounted);
   const platformRef = useLatestRef(platform);
+  const openRef = useLatestRef(open);
 
   const update = React.useCallback(() => {
     if (!referenceRef.current || !floatingRef.current) {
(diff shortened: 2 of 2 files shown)


asso1985:
Are you sure that this did not break what is documented here? 

```
// Once `isOpen` flips to `true`, `isPositioned` will switch to `true`
// asynchronously. We can use an Effect to determine when it has
// been positioned.
useEffect(() => {
  if (isPositioned) {
    // ...
  }
}, [isPositioned]);
```

isPositioned is no longer changing

atomiks:
> isPositioned is no longer changing

It works locally just fine, which means you might be experiencing an edge case where you were relying on the old behavior in some scenario. The only thing I can think of is `update()` for positioning is being called before your `open` state changes, which sets a ref inside the layout effect, but I'm not sure when/how this can happen given `update()` is always called in a later effect.

Can you provide a reproduction where it fails?

atomiks:
Okay, I know where it can fail, but it seems to highlight improper usage.

It can happen when combining these usage scenarios:
1. Not using conditional rendering, so the floating element is always present when the component is mounted.
2. Using `whileElementsMounted: autoUpdate` (which isn't supported when using case 1 above), or not specifying `autoUpdate` at all.
3. If not using `autoUpdate` at all, also not calling `update()` when `open` changes to `true`. This is the minimum requirement for this new logic to work (and also for the positioning to work reliably in general).

I would say this isn’t technically a regression but rather a bug fix that caused code relying on the old behavior to unfortunately start breaking unexpectedly (something semver isn't capable of handling). However it actually highlights an issue in the consumer code because the positioning wasn’t implemented correctly before.

When not using conditional rendering, this is the correct way to use `autoUpdate`:

```js
useLayoutEffect(() => {
  if (open && elements.reference && elements.floating) {
    return autoUpdate(elements.reference, elements.floating, update);
  }
}, [elements.reference, elements.floating, open, update]);
```

Or at the very least, when not using `autoUpdate`:

```js
useLayoutEffect(() => {
  if (open) {
    update();
  }
}, [open, update]);
```

This not only makes `isPositioned` work as expected, but also fixes a bug with the positioning implementation as well.

asso1985:
@atomiks Thanks for clarifying.

When not using conditional rendering tho if open is set to true and then false, would you expect to have isPositioned as `false`?

atomiks:
Yes because even though it was already positioned and remains mounted, the position can become "stale" while it's closed. It's possible that the next open has a drastically different position where you'll need to wait for `isPositioned` once again to do `focus` or scrolling since the old position is now out of view. e.g. reference changed its location on the screen due to scrolling containers or something else.

asso1985:
ok this is weird to me and it was not like this before (if it was a bug or not whatever) 

To me it should return `isPositioned: true` even if the pos did not change from previous

PR #3043, opened 2024-09-11
fix(useFloating): avoid setting `isPositioned` to true when `open` is false

lozinsky:
Closes https://github.com/floating-ui/floating-ui/issues/3017

Sync with https://github.com/floating-ui/floating-ui/pull/3042


atomiks:
Does this close the whole issue, or was my comment and subsequent PR only a somewhat related bug?

lozinsky:
This should close #3017 and also sync vue behaviour with react behaviour. Your PR is about the same problem.

PR #3045, opened 2024-09-12
fix: handle html relative offset

atomiks:


FezVrasta (approved):

atomiks:
This snapshot: [packages/dom/test/functional/transform.test.ts-snapshots/html-linux.png](https://github.com/floating-ui/floating-ui/pull/3045/commits/0436d07775f335294eb3951c3bfbe71cdaa55e57#diff-c6309504096a563414e579740ad905651960c1ebc2f5de5741df5dac2fca9c54) was testing an already-broken case that wasn't actually fixed, so it's not a regression

PR #3049, opened 2024-09-14
fix(FloatingFocusManager): use fallback return focus element

### LINKED ISSUE #3029: FloatingFocusManager incorrect focus restore when the the reference unmounts before the floating element is closed
**Describe the bug**
when `FloatingFocusManager` unmounts and it has `returnFocus={true}` - if the reference element that opened it is no longer available in the DOM (because it is conditionally rendered) then focus moves to a random (the first?) focusable element.

in my case - I have a button that opens a menu and when one of the items inside of it is clicked - I render a different component which doesn't contain the button (i.e: the reference element) so it's no longer available in the DOM.
when this happens, the focus moves to another element in my page which then runs some problematic side effects (due to an `onFocus` event handler on the now focused element) which results in a buggy UX

This behavior felt incorrect to me, but if it's actually the intended behavior then I can simply pass `returnFocus={false}` wherever i need

**To Reproduce**
https://codesandbox.io/p/sandbox/amazing-vaughan-ywsm5s?workspaceId=6aebd151-ce60-4ebd-b309-b7331d6bb1ed

1. Instructions in codesandbox

**Expected behavior**
In case the reference element is not available anymore when returning the focus, then do nothing (i think browsers default to `document.body`)

**Context:**

- OS: macOS
- Browser: Chrome
- Version: 0.26.23

Thanks!

atomiks:
Closes #3029

### DIFF
--- a/.changeset/dull-bugs-rescue.md
+++ b/.changeset/dull-bugs-rescue.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(FloatingFocusManager): avoid returning focus to nearest tabbable element of the reference if it gets removed when the floating element closes to avoid unwanted focus effects of unrelated elements firing. Tab index context remains preserved if the floating element is portaled.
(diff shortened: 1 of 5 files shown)


roigiladi:
Thanks @atomiks! 👑

PR #3075, opened 2024-10-10
fix(useDismiss): handle IME keydown events on `Escape`

atomiks:
https://github.com/mui/material-ui/pull/39713

- `event.isComposing` can't be used in Safari due to the order of events.
- `event.which === 229` can't be used since event listeners are added on the floating element AND document and fires it twice, the second with `event.which === 0` and thus doesn't end up preventing the keypress.

A ref using `compositionstart`/`compositionend` events solves the above issues

### DIFF
--- a/.changeset/olive-dogs-look.md
+++ b/.changeset/olive-dogs-look.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(useDismiss): handle IME keydown events on Escape
--- a/packages/react/src/hooks/useDismiss.ts
+++ b/packages/react/src/hooks/useDismiss.ts
@@ -419,7 +419,11 @@ export function useDismiss(
           escapeKeyCapture ? closeOnEscapeKeyDownCapture : closeOnEscapeKeyDown,
           escapeKeyCapture,
         );
-        doc.removeEventListener('compositionstart', handleCompositionEnd, true);
+        doc.removeEventListener(
+          'compositionstart',
+          handleCompositionStart,
+          true,
+        );
         doc.removeEventListener('compositionend', handleCompositionEnd, true);
       }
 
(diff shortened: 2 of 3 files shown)


mj12albert:
@atomiks I've tested this locally and it works in Chrome/Firefox (latest) but not Safari (18.0.1), here's some screen recordings I made (Quicktime recordings don't show keypresses/keystrokes but you should be able to see that in Safari the IME and the floating elem always close at the same time)

### Safari 18.0.1

https://github.com/user-attachments/assets/2175fce0-f4d5-49ea-9fd2-90c08fd0e70c

### Chrome 129.0.6668.101

https://github.com/user-attachments/assets/281d336d-7d4f-4df6-a8a5-0de8c56d3711

### Firefox 131.0.2

https://github.com/user-attachments/assets/0c9ae0d5-52bc-416d-88a3-bc039fa4d1cd

PS: I made a test page [here](https://github.com/mj12albert/floating-ui/blob/fix/useDismiss-composition/website/pages/test.js) because I just couldn't install this branch as a dependency in CSB/Stackblitz 😓

mj12albert:
@atomiks I've tested the latest changes in Safari - once in a while, it fails the very first time the popup is opened and input is attempted (1:00 in the screen capture), but works the 2nd time you try to do it again and afterwards 

https://github.com/user-attachments/assets/4fcb0574-4ce5-4733-8ec6-ce07c6e6f581

atomiks:
It's better than nothing but ideally we'd find the cause of why it's failing on the first time too. 

I'm not sure why the timeout doesn't always work, unless the `keydown` fires one tick after `compositionend`? Or `compositionstart` has an issue, or something else?

atomiks:
Seems like Safari needs ~`5ms` in the `setTimeout` to work, not `0ms` or `1ms` 🙄 . With `5` it consistently seems to be prevented, but 0/1 it always closes...

PR #3076, opened 2024-10-10
fix(getClippingRect): allow passing DOMRect as boundary

acupofspirt:
This pr allows passing `DOMRect` objects a the [boundary](https://floating-ui.com/docs/detectOverflow#boundary)/[rootBoundary](https://floating-ui.com/docs/detectOverflow#rootboundary) option in `detectOverflow`.
While `DOMRect` is compatible with the expected `Rect` interface, its properties are not enumerable and can't be used with the spread operator.

In other words, this code will work without NaNs:
```ts
computePosition(referenceEl, floatingEl, {
  middleware: [
    shift({
      boundary: myContainer.getBoundingClientRect()
    })
  ]
})
```

### DIFF
--- a/.changeset/ten-bears-protect.md
+++ b/.changeset/ten-bears-protect.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/dom": patch
+---
+
+fix(getClippingRect): allow passing `DOMRect` as a `boundary`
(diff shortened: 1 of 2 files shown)


FezVrasta (commented):
I think Rect/DomRect also provide top/bottom/right/left? So we want to preserve them as well?

atomiks:
@FezVrasta the type it accepts is a `Rect`, so it doesn't need the other `ClientRect` properties. This looks fine

PR #3083, opened 2024-10-15
feat(FloatingFocusManager): Specify element to return focus to

michaldudak:
Made it possible to explicitly specify the element that will be focused after FloatingFocusManager closes. This reuses the same logic that's already in place to determine when to focus an element and simply adds an ability to focus the provided item.

### DIFF
--- a/packages/react/test/unit/FloatingFocusManager.test.tsx
+++ b/packages/react/test/unit/FloatingFocusManager.test.tsx
@@ -177,6 +177,29 @@ describe('returnFocus', () => {
     expect(screen.getByTestId('reference')).not.toHaveFocus();
   });
 
+  test('ref', async () => {
+    function Test() {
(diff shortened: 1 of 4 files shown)


packages/react/test/unit/FloatingFocusManager.test.tsx
```diff
@@ -177,6 +177,29 @@ describe('returnFocus', () => {
...
+          <input />
+          <input data-testid="focus-target" ref={ref} />
+          <input />
+          <App returnFocus={ref} />
+        </div>
+      );
+    }
+
+    render(<Test />);
+    screen.getByTestId('reference').focus();
```
atomiks, line 194: Focus calls don't need `act`?

michaldudak: Apparently, since the event handler is sync and doesn't trigger a rerender nor effects.

PR #3112, opened 2024-11-13
chore(.github): update github actions 'uses' to latest version

sukvvon:
image

* github actions is now using node version 20 but actions/checkout@v3, pnpm/action-setup@v2, actions/setup-nodes@v3 are using node verion 16 which is deprecated in github actions, so i update it all v4, you can see [GitHub Actions: Transitioning from Node 16 to Node 20](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/)

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -4,13 +4,12 @@ runs:
   using: composite
   steps:
     - name: Checkout
-      uses: actions/setup-node@v3
+      uses: actions/setup-node@v4
       with:
-        node-version: '18'
+        node-version-file: '.nvmrc'
         registry-url: 'https://registry.npmjs.org'
 
     - name: Install dependencies
-      uses: pnpm/action-setup@v2
+      uses: pnpm/action-setup@v4
       with:
-        version: 8
         run_install: true
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@floating-ui/monorepo",
   "private": true,
+  "packageManager": "pnpm@8.13.1",
   "version": "0.0.0",
   "description": "Floating UI monorepo",
   "engines": {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     name: Linting and type checking
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
       - uses: ./.github/actions/typescript
 
@@ -18,7 +18,7 @@ jobs:
     name: Unit tests
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
       - uses: ./.github/actions/unit
 
@@ -26,6 +26,6 @@ jobs:
     name: Functional tests
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
       - uses: ./.github/actions/functional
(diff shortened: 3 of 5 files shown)


.github/actions/setup/action.yml
```diff
@@ -4,13 +4,12 @@ runs:
   using: composite
   steps:
     - name: Checkout
-      uses: actions/setup-node@v3
+      uses: actions/setup-node@v4
       with:
-        node-version: '18'
```
sukvvon, line 9: we have '.nvmrc'.
so we can use 'node-version-file' instead of 'node-version'.

.github/actions/setup/action.yml
```diff
@@ -4,13 +4,12 @@ runs:
...
+        node-version-file: '.nvmrc'
         registry-url: 'https://registry.npmjs.org'
 
     - name: Install dependencies
-      uses: pnpm/action-setup@v2
+      uses: pnpm/action-setup@v4
       with:
-        version: 8
```
sukvvon, line 15: adding 'packageManager' in 'package.json', we dont need to use 'version' in 'pnpm/action-setup'.
[its optional when packageManager is not exist.](https://github.com/pnpm/action-setup?tab=readme-ov-file#version)

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@floating-ui/monorepo",
   "private": true,
+  "packageManager": "pnpm@8.13.1",
```
atomiks, line 4: Do you by any chance know how to migrate to pnpm v9 without the lockfile being blown away? Tried earlier but it causes transitive dependencies to update, and they have breaking changes despite being patches/minors which causes the tests to start failing.

Also, while it can be helpful to be exact, I think I'd prefer being permissive versioning here to the major or minor than exact patch.

sukvvon, line 4: pnpm version in the packageManager field of engines has been specified to match the packageManager version.”

sukvvon: > Do you by any chance know how to migrate to pnpm v9 without the lockfile being blown away?

if pnpm is updated to v9, the lock file will inevitably change. 

> Tried earlier but it causes transitive dependencies to update, and they have breaking changes despite being patches/minors which causes the tests to start failing.

can you explain this more detail?

sukvvon: the purpose of this pr is primarily to remove warnings occurring in the action.

.github/actions/setup/action.yml
```diff
@@ -4,13 +4,12 @@ runs:
   using: composite
   steps:
     - name: Checkout
-      uses: actions/setup-node@v3
+      uses: actions/setup-node@v4
       with:
-        node-version: '18'
+        node-version-file: '.nvmrc'
```
sukvvon, line 9: we have '.nvmrc'.
so we can use 'node-version-file' instead of 'node-version'.

package.json
atomiks: > to standardize the environment

Never mind, I see the difference between the  `engines` and `packageManager` field behavior

> if pnpm is updated to v9, the lock file will inevitably change.

What I wanted is for the lockfile adapt to v9's format, but without changing the versions of transitive dependencies used to install. Instead it completely updates all the versions and defeats the purpose of the lockfile, and tests start breaking which seemed nontrivial to fix because the packages don't respect semver. The lockfile can't be updated to v9's format without changing versions simultaneously?

> the purpose of this pr is primarily to remove warnings occurring in the action.

Yeah, this PR looks good on that front. Was just wondering if you had an idea how to solve that issue so we can use pnpm the latest v9 instead of v8 while we're here

sukvvon: > What I wanted is for the lockfile adapt to v9's format, but without changing the versions of transitive dependencies used to install. Instead it completely updates all the versions and defeats the purpose of the lockfile, and tests start breaking which seemed nontrivial to fix because the packages don't respect semver. The lockfile can't be updated to v9's format without changing versions simultaneously?

If this PR is merged, I will look for a solution and submit a new PR.

atomiks:
Thank you!

PR #3117, opened 2024-11-13
fix(useListNavigation): incorrect behaviour with RTL grid

### LINKED ISSUE #3115: RTL `useListNavigation` doesn't work with "both" orientation grid
**Describe the bug**

`useListNavigation` doesn't handle correctly RTL navigation when using a grid with orientation set to `'both'`.

**To Reproduce**

1. Go to https://codesandbox.io/p/sandbox/floating-ui-react-dom-template-forked-rz5mxv?workspaceId=2e61ccf6-8b29-4b40-92d2-72d59e94b20e
2. Click on the "Reference" button.
3. Press down the `ArrowLeft` key.
4. The focus moves to the right (on item 0) while it should move to the left instead (on item 2).

**Expected behavior**

When using an RTL layout, `ArrowRight` should focus the next item in the list, and `ArrowLeft` the previous item.

**Context:**

- OS: Linux
- Browser: Chrome
- Version: @floating-ui/react 0.26.27

madoci:
Fixes #3115

### DIFF
--- a/.changeset/wet-bottles-sing.md
+++ b/.changeset/wet-bottles-sing.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(useListNavigation): incorrect behaviour with RTL grid
--- a/packages/react/src/components/Composite.tsx
+++ b/packages/react/src/components/Composite.tsx
@@ -186,6 +186,7 @@ export const Composite = React.forwardRef<
               event,
               orientation,
               loop,
+              rtl: false,
               cols,
               // treat undefined (empty grid spaces) as disabled indices so we
               // don't end up in them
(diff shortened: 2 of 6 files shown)


atomiks:
Hi, thanks. Just looks like `rtl` needs to be added to `CompositeProps`

madoci:
> Hi, thanks. Just looks like `rtl` needs to be added to `CompositeProps`

I got mixed up and forgot to add the `rtl` parameter to the `getGridNavigatedIndex` function of `Composite`. I am also currently working on #3116, but I was planning to do an other PR for that. I can merge it with this one if you prefer.

atomiks:
I suppose another PR works

PR #3119, opened 2024-11-14
feat(Composite): add RTL support

madoci:
Closes #3116

### DIFF
--- a/packages/react/src/components/Composite.tsx
+++ b/packages/react/src/components/Composite.tsx
@@ -66,6 +66,10 @@ export interface CompositeProps {
    */
   loop?: boolean;
   /**
+   * Whether the direction of the composite’s navigation is in RTL layout.
+   */
+  rtl?: boolean;
+  /**
    * Determines the number of columns there are in the composite
    * (i.e. it’s a grid).
    */
@@ -118,6 +122,7 @@ export const Composite = React.forwardRef<
     render,
     orientation = 'both',
     loop = true,
+    rtl = false,
     cols = 1,
     disabledIndices,
     activeIndex: externalActiveIndex,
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Composite.tsx
```diff
@@ -65,6 +65,10 @@ export interface CompositeProps {
    * or last item.
    */
   loop?: boolean;
+  /**
+   * Whether the direction of the composite’s navigation is in RTL layout.
+   */
+  rtl?: boolean;
```
atomiks, line 71: Given it's a component, does `dir="rtl"` make more sense?

madoci: Good question ! Given that most props of `Composite` are the same as the `useListNavigation` props, I used the same `rtl` prop as well. But `dir="rtl"` might make more sense indeed. However, from what I understand (I never really used RTL layout), the HTML `dir` attribute is propagated to all children until it is overriden. So if we go that road, I think we should also get the current `dir` attribute (which could be set by some higher parent) and add a `dir` attribute if the `dir` prop is set to ensure it is propagated to children. What do you think ?

atomiks: I think let's leave the boolean prop. An auto check for `getComputedStyle(ref).direction === 'rtl'` could be added internally, which would allow it to adapt automatically, but be overridden by the prop if they specify it. This splits behavior and layout. The auto check would need to be added in `useListNavigation` as well, probably though. For now we can leave it as a manual prop?

madoci: I'm fine with that. I realized that the `Composite` component does not render any container element, so we can't propagate down the `dir` attribute anyway.

PR #3131, opened 2024-11-28
feat(FloatingFocusManager): Add the `inert` attribute on outside nodes when `modal=true`

michaldudak:
If available, uses the `inert` attribute to nodes outside of a modal floating elements with `guards=true`. This makes pointers modal as well, not just keyboard/focus movement.

I'm wondering if the opposite, `aria-hidden` when `guards = false`, makes any sense. I'm leaning towards not allowing it (and potentially simplifying the API with a boolean prop).

Fixes the issue outlined in https://github.com/floating-ui/floating-ui/discussions/3098

### DIFF
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -21,7 +21,7 @@ import {createAttribute} from '../utils/createAttribute';
 import {enqueueFocus} from '../utils/enqueueFocus';
 import {getAncestors} from '../utils/getAncestors';
 import {getChildren} from '../utils/getChildren';
-import {markOthers, supportsInert} from '../utils/markOthers';
+import {markOthers} from '../utils/markOthers';
 import {
   getNextTabbable,
   getPreviousTabbable,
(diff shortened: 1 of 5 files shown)


atomiks:
> I'm wondering if the opposite, aria-hidden when guards = false, makes any sense. I'm leaning towards not allowing it (and potentially simplifying the API with a boolean prop).

`guards=false` relies on `inert` to work, so yeah that'd be an invalid state

packages/react/src/components/FloatingFocusManager.tsx
```diff
@@ -430,14 +459,26 @@ export function FloatingFocusManager(
         ...portalNodes,
         startDismissButtonRef.current,
         endDismissButtonRef.current,
+        ...(outsideElementsInert
+          ? [
+              beforeGuardRef.current,
+              afterGuardRef.current,
+              portalContext?.beforeOutsideRef?.current,
+              portalContext?.afterOutsideRef?.current,
```
atomiks, line 467:
```suggestion
              portalContext?.beforeOutsideRef.current,
              portalContext?.afterOutsideRef.current,
```

packages/react/src/components/FloatingFocusManager.tsx
```diff
@@ -40,18 +41,13 @@ function addPreviouslyFocusedElement(element: Element | null) {
...
     (el) => el.isConnected,
   );
-  let tabbableEl = element;
-  if (!tabbableEl || getNodeName(tabbableEl) === 'body') return;
-  if (!isTabbable(tabbableEl, getTabbableOptions())) {
-    const tabbableChild = tabbable(tabbableEl, getTabbableOptions())[0];
-    if (tabbableChild) {
-      tabbableEl = tabbableChild;
+
+  if (element && getNodeName(element) !== 'body') {
```
michaldudak, line 45: This restores this function pre #2826. The logic introduced in this PR is moved just before the focus is about to be applied. This was done because at this point, `tabbable` couldn't find a focusable descendant because it has the `inert` attribute.

I believe this was a bug when `guards = false`.

PR #3146, opened 2024-12-05
fix(useId): add `| undefined` return type for React 17

atomiks:


mihkeleidast:
Hmm, @atomiks what was the reasoning behing this? I get that it's now "technically accurate" type-wise, but this may cause quite a bit of issues where id is required but now it may be undefined because the hook sometimes initially returns `undefined`. Making the hook sort of useless, if we still need to have another fallback.

I sorta fixed it with `const id = useId() ?? '';`, as I don't see any real world issues with it returning undefined/empty id on React 17 initially. Wondering if maybe it would be better to return empty string here instead of undefined so it wouldn't need additional fallbacks type-wise?

FezVrasta:
I think `undefined` is more correct than `''`, if you are okay with `''` you can handle it on user-land.

atomiks:
@mihkeleidast a couple of issues actually appeared internally due to this, as you need to wait for the string to exist. The hook is only useful if you're using React 17 anyway (though, it is for internal id's)

mihkeleidast:
Got it, I think my workaround is fine for me as it's mostly used in places where the id just needs to be statically available in component rendering (so if it's empty/undefined for a millisecond, it doesn't actually make a difference). Using the hook as we still need to support React 17 (can't wait to drop that thing).

PR #3168, opened 2024-12-17
fix(autoUpdate): work around bug where intersection ratio is 1 but element actually moved

### LINKED ISSUE #3166: Under performance constraints, floating element can get stuck away from reference element on move
**Describe the bug**

When CPU is under strain and the reference element moves, the floating element can get stuck and no longer auto-update to stay anchored to the reference element. Once this happens, the floating element only begins tracking the reference element again if the reference element moves back to the exact position it was in at the moment the floating element got stuck.

I believe this is happening because sometimes, at https://github.com/floating-ui/floating-ui/blob/6a9f75821398f71bd3d3c644af41d40d06367544/packages/dom/src/autoUpdate.ts#L82, `ratio` is 1 (and so we don't call `refresh`) because at the time `handleObserve` was called the element was indeed fully intersecting the root area – but by the time that code runs, the element can have already moved entirely outside of the `IntersectionObserver` root. 

When that happens, it seems the `IntersectionObserver` instance internally knows the target element is no longer intersecting (despite that fact that the last time it ran the callback, the target element _was_ intersecting), since it only runs the intersection change callback again when the target element becomes fully intersecting again.

**To Reproduce**

1. Go to https://codesandbox.io/p/sandbox/trt7fc
2. In devtools, throttle your CPU to a 20x slowdown:
    
    CleanShot 2024-12-16 at 15 40 18
3. Drag the range input left and right a few times until you see that the floating element gets stuck. Sometimes it takes a few tries and a little finesse. Try varying between slow and fast motion back and forth.

**Expected behavior** 

The floating element should never get stuck away from the reference element. It should always stay anchored to the reference as the reference moves.

**Screen recording**

https://github.com/user-attachments/assets/f470f8f7-1bfe-40f4-8d18-8cbd55f641fd

In that recording, before I enable CPU throttling, you can see the floating element sometimes gets stuck for a second but then recovers. (That's OK - not the bug I'm reporting here). Once I enable 20x CPU throttling, you can see the floating element gets stuck permanently, until I move the reference element back to the exact spot at which the problem started.

**Context:**

- OS: macOS 15.2
- Browser Chrome
- Version 132.0.6834.46

mrdavidjcole:
Fixes #3166

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -211,13 +211,7 @@ export function autoUpdate(
   function frameLoop() {
     const nextRefRect = getBoundingClientRect(reference);
 
-    if (
-      prevRefRect &&
-      (nextRefRect.x !== prevRefRect.x ||
-        nextRefRect.y !== prevRefRect.y ||
-        nextRefRect.width !== prevRefRect.width ||
-        nextRefRect.height !== prevRefRect.height)
-    ) {
+    if (prevRefRect && !rectsAreEqual(prevRefRect, nextRefRect)) {
       update();
     }
 
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -4,6 +4,7 @@ import {getDocumentElement, getOverflowAncestors} from '@floating-ui/utils/dom';
 import type {FloatingElement, ReferenceElement} from './types';
 import {getBoundingClientRect} from './utils/getBoundingClientRect';
 import {unwrapElement} from './utils/unwrapElement';
+import {rectsAreEqual} from './utils/rectsAreEqual';
 
 export interface AutoUpdateOptions {
   /**
@@ -53,7 +54,8 @@ function observeMove(element: Element, onMove: () => void) {
   function refresh(skip = false, threshold = 1) {
     cleanup();
 
-    const {left, top, width, height} = element.getBoundingClientRect();
+    const elementRectForRootMargin = element.getBoundingClientRect();
+    const {left, top, width, height} = elementRectForRootMargin;
 
     if (!skip) {
       onMove();
@@ -93,6 +95,23 @@ function observeMove(element: Element, onMove: () => void) {
         } else {
           refresh(false, ratio);
         }
+      }
+
+      if (
+        ratio === 1 &&
+        !rectsAreEqual(
+          elementRectForRootMargin,
+          element.getBoundingClientRect(),
+        )
+      ) {
+        // It's possible that even though the ratio is reported as 1, the
+        // element is not actually fully within the IntersectionObserver's root
+        // area anymore. This can happen under performance constraints. This may
+        // be a bug in the browser's IntersectionObserver implementation. To
+        // work around this, we compare the element's bounding rect now with
+        // what it was at the time we created the IntersectionObserver. If they
+        // are not equal then the element moved, so we refresh.
+        refresh();
       }
 
       isFirstUpdate = false;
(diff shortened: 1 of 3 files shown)


packages/dom/src/autoUpdate.ts
```diff
@@ -95,6 +97,23 @@ function observeMove(element: Element, onMove: () => void) {
         }
       }
 
+      if (
+        ratio === 1 &&
+        !rectsAreEqual(
+          elementRectForRootMargin,
+          element.getBoundingClientRect(),
```
FezVrasta, line 104: This will result in a layout paint operation IIRC, I wonder how it will impact performance.

mrdavidjcole: Since we first check if `ratio === 1` (and bail out if that's `false`), this new `element.getBoundingClientRect()` call happens relatively rarely compared to the one at https://github.com/floating-ui/floating-ui/blob/e61070e24ac03f92d43d3aafe8dea4b8fd968af2/packages/dom/src/autoUpdate.ts#L56 (or on my branch, https://github.com/mrdavidjcole/floating-ui/blob/6e3f721482197c0c0a0e1fed6db23899311a5df7/packages/dom/src/autoUpdate.ts#L57).


https://github.com/user-attachments/assets/3befac19-9bc6-4189-aeb2-89a294fff18a



In other words, we already call `element.getBoundingClientRect()` every time we're setting up a new intersection observer (i.e. on every `refresh()` call), and this is a marginal increase to that, since it only happens when the `ratio` goes back to `1`. As far as I can tell there isn't a perceptible performance hit.

FezVrasta:
@atomiks this LGTM, want to take a look too?

FezVrasta (approved):

packages/dom/src/utils/rectsAreEqual.ts
```diff
@@ -0,0 +1,5 @@
+export function rectsAreEqual(a: DOMRect, b: DOMRect) {
+  return (
+    a.x === b.x && a.y === b.y && a.width === b.width && a.height === b.height
+  );
+}
```
atomiks, line 5: This can be reused in `frameLoop` which does the same check (but inline)

mrdavidjcole: Ah, good call! https://github.com/floating-ui/floating-ui/pull/3168/commits/889fd05323725e88bfb6d9aaf768e25b2fcac8fa

atomiks:
With these changes, it seems slightly better but still fails to remain anchored well. I modified the `AutoUpdate` demo to make the reference small and floating element bigger here:

https://github.com/user-attachments/assets/f6833fe6-b1c7-4949-878f-0ce76c5769fc

mrdavidjcole:
> With these changes, it seems slightly better but still fails to remain anchored well. I modified the `AutoUpdate` demo to make the reference small and floating element bigger here:
> 
@atomiks I believe what you're seeing there is an adjacent issue, but not the exact one addressed by this PR. 

This PR is purely about fixing how the floating element can get _permanently_ stuck (well, permanently until you move the reference element back to exactly where it was at the moment the floating element got stuck).

In your recording, I don't see the floating element permanently stuck - only stuck for 1000ms while moving quickly such that `ratio` becomes `0` pretty often, which I believe is a result of this throttling timeout: https://github.com/floating-ui/floating-ui/blob/b80ccaf9cfd7f80f28546906c60284c8385940f0/packages/dom/src/autoUpdate.ts#L90-L92

It'd be great to fix that too! Just a separate issue AFAICT.

atomiks:
Ah I see this part of your issue now:

> In that recording, before I enable CPU throttling, you can see the floating element sometimes gets stuck for a second but then recovers. (That's OK - not the bug I'm reporting here).

It took me a while but eventually I could reproduce it. It looks quite rare, and seems like the adjacent issue is arguably more important to fix since it occurs far more frequently in practice? Although it does eventually "recover", unlike this one, taking 1 second to do so is still poor. 

This PR is still fine though, thanks!

PR #3173, opened 2024-12-22
Support for shorthand transform CSS properties (translate, scale, rotate)

### LINKED ISSUE #3172: Incorrect handling of the new CSS translate property
Please follow these steps to reproduce the issue:

1. Go to the reproduction ['sandobox'](https://codesandbox.io/p/sandbox/floatingui-2p4nrg)
2. Click "Show old dialog" (which uses the old `transform` property)
3. See how the tooltip is positioned correctly
4. Click "Close dialog"
2. Click "Show new dialog" (which uses the new `translate` property)
3. See how the tooltip is positioned incorrectly

**Expected behavior**
The tooltip should be positioned correctly even if using the newer translate CSS property.

SCR-20241221-ognl
SCR-20241221-oglq

There's a new CSS property that replaces the `transform: translate(-50%, -50%)` notation - `translate: -50% -50%`.
Here is the link to its docs - https://developer.mozilla.org/en-US/docs/Web/CSS/translate

It looks like the code that looks up the containing block is not accounting for it here - https://github.com/floating-ui/floating-ui/blob/b80ccaf9cfd7f80f28546906c60284c8385940f0/packages/utils/src/dom.ts#L96

We should probably add a `translate`, `scale`, and other newly added shorthand to the checklist there.

kilobyte2007:
Fixed #3172


packages/utils/src/dom.ts
```diff
@@ -92,13 +92,17 @@ export function isContainingBlock(
     : elementOrCss;
 
   // https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block
+  // https://drafts.csswg.org/css-transforms-2/#individual-transforms
   return (
     css.transform !== 'none' ||
+    css.translate !== 'none' ||
+    css.scale !== 'none' ||
+    css.rotate !== 'none' ||
     css.perspective !== 'none' ||
```
atomiks, line 101: These each need a ternary like `css.translate ? css.translate !== 'none' : false`. Older browsers will return `undefined` and incorrectly think a containing block is created (though if you're using those properties, they don't work in the first place, but still).

At that point you might as well each make them into a `.some()` with the `none` check since they repeat the same work

kilobyte2007: Yeah, I was thinking of this, but I didn't dig deep enough to understand how often is this invoked, and whether there'll be some perf. penalty for using `some` here.

I can implement this change regardless.

FezVrasta: `some` would create a new scope for each iteration and would increase the GC events.

I would suggest to define a helper function that checks for undefined or none and use it here.

Alternatively a for loop will work to and won't produce GC events

kilobyte2007: @FezVrasta @atomiks, I've created a helper function to do that, can you please take a look now? What do you think?

packages/utils/src/dom.ts
```diff
@@ -92,13 +107,13 @@ export function isContainingBlock(
...
+  // https://drafts.csswg.org/css-transforms-2/#individual-transforms
   return (
-    css.transform !== 'none' ||
-    css.perspective !== 'none' ||
+    hasNonEmptyCSSValue(css, ['transform', 'translate', 'scale', 'rotate', 'perspective']) ||
     (css.containerType ? css.containerType !== 'normal' : false) ||
     (!webkit && (css.backdropFilter ? css.backdropFilter !== 'none' : false)) ||
     (!webkit && (css.filter ? css.filter !== 'none' : false)) ||
-    ['transform', 'perspective', 'filter'].some((value) =>
+    ['transform', 'translate', 'scale', 'rotate', 'perspective', 'filter'].some((value) =>
```
atomiks, line 116: Since `.some` was already used below as you mentioned, it should be fine to use for this as well

```js
['transform', 'translate', 'scale', 'rotate', 'perspective'].some((value) =>
  css[value] ? css[value] !== 'none' : false
)
```

kilobyte2007: Done

FezVrasta (approved):

Issue #3067, opened 2024-10-03
Positioning inside container-type: inline-size is off when using position: fixed on the popup


IvandeWolf:
**Describe the bug**

When using the `computePosition()` function in the following context the popup is placed off
(see also https://codesandbox.io/p/sandbox/funny-glade-jfr7kl):

- The button and popup element are located inside a div somewhere in the DOM tree with the CSS property: `container-type: inline-size`
- Using strategy `fixed`
- The popup element has CSS property `position: fixed`. This is because we want the popup to be visible and overflowing even in a small side container for example.

**To Reproduce**

See code sandbox: https://codesandbox.io/p/sandbox/funny-glade-jfr7kl

**Expected behavior**

The popup should be placed below the button.

**Screenshots**

![image](https://github.com/user-attachments/assets/4be152b0-3f55-41ae-bb6f-08791a15baac)

**Additional context**

I have debugged the code and found the following code to be executed which might be 'a problem' but I also might be doing something weird. (Found in dom/src/utils/getRectRelativeToOffsetParent.ts)

https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/utils/getRectRelativeToOffsetParent.ts#L23
```ts
/**
 * Here we get the coordinator for the element and this actually already returns
 * the correct x and y coordinates for the popup placement.
 */
const rect = getBoundingClientRect(element, true, isFixed, offsetParent);
```

https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/utils/getRectRelativeToOffsetParent.ts#L36
```ts
/**
 * We enter this if-statement, because the offsetParent is the div
 * with the container-type: inline-size and that is indeed an element.
 */
if (isOffsetParentAnElement) {
    /**
     * This returns the rect values of the offsetParent.
     * Which means the x and y coordinator of the div with container-type: inline-size.
     */
    const offsetRect = getBoundingClientRect(
      offsetParent,
      true,
      isFixed,
      offsetParent,
    );
    offsets.x = offsetRect.x + offsetParent.clientLeft;
    offsets.y = offsetRect.y + offsetParent.clientTop;
  } else if (documentElement) {
```

And these offsets are being used here:

https://github.com/floating-ui/floating-ui/blob/master/packages/dom/src/utils/getRectRelativeToOffsetParent.ts#L65
```ts
/**
 * Here we substract the offsets from the x and y coordinates which make the position move back.
 * While the rect.left and rect.top where already the correct placement for the popup.
[...]

PR #3191, opened 2025-01-07
fix(dom)!: remove check for containerType in isContainingBlock

ptu14:
This change removes the condition `css.containerType !== 'normal'`, as browser vendors are moving away from this behavior. While MDN is a helpful resource, it is not the definitive source of truth for browser implementation details or CSS specifications. Therefore, aligning with the current state of browser implementations makes the code more robust and consistent.

The logic still retains the check for the contain property, which explicitly relates to the intended creation of a containing block. This ensures that the behavior remains aligned with the specification and the actual implementation of the contain property, providing the desired functionality without relying on deprecated or inconsistent behavior.

By adapting to this reality, we simplify the code and reduce reliance on outdated assumptions while maintaining the core functionality.

this is BREAKING CHANGE

https://github.com/floating-ui/floating-ui/issues/3067
https://github.com/w3c/csswg-drafts/issues/10102#issuecomment-2377653411

### DIFF
--- a/packages/utils/src/dom.ts
+++ b/packages/utils/src/dom.ts
@@ -100,7 +100,6 @@ export function isContainingBlock(
           ? css[value as keyof CSSStyleDeclaration] !== 'none'
           : false,
     ) ||
-    (css.containerType ? css.containerType !== 'normal' : false) ||
     (!webkit && (css.backdropFilter ? css.backdropFilter !== 'none' : false)) ||
     (!webkit && (css.filter ? css.filter !== 'none' : false)) ||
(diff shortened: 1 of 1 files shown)


atomiks:
Hi @ptu14 I'm not going to change how this works until Firefox and Safari have updated to match Chrome.  Your change works on new Chrome, but not old Chrome or Safari & Firefox.

For now, you should always add `contain: layout` along with `container-type: inline-size`. This makes the containment work identically across browsers (Safari/FF implicitly add it, along with old Chrome). See my reasoning here: https://github.com/floating-ui/floating-ui/issues/3067#issuecomment-2403678795

PR #3194, opened 2025-01-09
Request 3181

stopsopa:
PR implementing suggested exports: https://github.com/floating-ui/floating-ui/discussions/3181


FezVrasta (commented):
LGTM but we should also document on the website that these exports don't follow semver and may break on fix releases too.

stopsopa:
@FezVrasta - To be honest I don't know where in the documentation mention it. Any suggestions?
Initially I've rather intended to just export it and not document it, this way it is out of the box non official api which can be found only by reading the code.
I could actually comment it above those exports in the code. How about that?

stopsopa:
So any suggestions where to document that? @FezVrasta

PR #3199, opened 2025-01-16
fix(useListNavigation): use parent's orientation when opening/closing nested menus

mnajdova:
I was trying to build a grid like navigation on a submenu and I found a bug where the opening/closing logic on submenus was depending on the submenu orientation instead of the paren't orientation. 

I also added support for grid-like submenus, by blocking the closing of the submenu  that uses the 'both' orientation until `Escape` key is pressed.

I can create a new demo if we don't want to pollute this one, as this may be considered a niche feature.

Demo recording:

https://github.com/user-attachments/assets/032d94a3-8832-4914-aca0-5d2472eda6cd

### DIFF
--- a/packages/react-dom/src/types.ts
+++ b/packages/react-dom/src/types.ts
@@ -151,5 +151,9 @@ export type UseFloatingOptions<RT extends ReferenceType = ReferenceType> =
        * @default false
        */
       transform?: boolean;
+      /**
+       * The orientation of the list navigation.
+       */
+      orientation?: 'horizontal' | 'vertical' | 'both';
     }
   >;
--- a/packages/react-dom/src/types.ts
+++ b/packages/react-dom/src/types.ts
@@ -151,9 +151,5 @@ export type UseFloatingOptions<RT extends ReferenceType = ReferenceType> =
        * @default false
        */
       transform?: boolean;
-      /**
-       * The orientation of the list navigation.
-       */
-      orientation?: 'horizontal' | 'vertical' | 'both';
     }
   >;
--- a/packages/react/src/hooks/useListNavigation.ts
+++ b/packages/react/src/hooks/useListNavigation.ts
@@ -986,6 +986,7 @@ export function useListNavigation(
   }, [
     activeId,
     ariaActiveDescendantProp,
+    cols,
     commonOnKeyDown,
     disabledIndicesRef,
     focusItemOnOpen,
--- a/packages/react/src/hooks/useListNavigation.ts
+++ b/packages/react/src/hooks/useListNavigation.ts
@@ -246,7 +246,7 @@ export function useListNavigation(
   context: FloatingRootContext,
   props: UseListNavigationProps,
 ): ElementProps {
-  const {open, onOpenChange, elements, dataRef} = context;
+  const {open, onOpenChange, elements} = context;
   const {
     listRef,
     activeIndex,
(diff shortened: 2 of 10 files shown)


packages/react-dom/src/types.ts
```diff
@@ -151,5 +151,9 @@ export type UseFloatingOptions<RT extends ReferenceType = ReferenceType> =
        * @default false
        */
       transform?: boolean;
+      /**
+       * The orientation of the list navigation.
+       */
+      orientation?: 'horizontal' | 'vertical' | 'both';
```
atomiks, line 157: This shouldn't be an option for the positioning (`react-dom`) package. I also don't think it needs to be an option for the `react` one either? `useListNavigation` should be solely responsible for handling it. The `dataRef` prop can share the orientation data

mnajdova: I was hesitant to include the orientation here, but I wasn't sure where else would make sense. I will look into adding it to the dataRef.

mnajdova: Done.

packages/react/src/hooks/useListNavigation.ts
```diff
@@ -289,6 +294,18 @@ export function useListNavigation(
...
+    (node) => node.id === parentId,
+  );
+
+  if (parentNode) {
+    parentOrientation = parentNode.context?.dataRef?.current.orientation;
+  }
+
+  React.useEffect(() => {
+    context.dataRef.current.orientation = orientation;
+  }, [context, orientation]);
```
atomiks, line 308: `useModernLayoutEffect` is likely safer here

packages/react/src/hooks/useListNavigation.ts
```diff
@@ -289,6 +294,18 @@ export function useListNavigation(
...
   const parentId = useFloatingParentNodeId();
   const tree = useFloatingTree();
+  let parentOrientation: UseListNavigationProps['orientation'] = undefined;
+  const parentNode = tree?.nodesRef.current.find(
+    (node) => node.id === parentId,
+  );
+
+  if (parentNode) {
+    parentOrientation = parentNode.context?.dataRef?.current.orientation;
+  }
```
atomiks, line 304: Refs can't be read in render, but instead the the event handler/effect (to share it, you can use `useEffectEvent` function that returns the value to read it on-demand?)

mnajdova: I thought I was still in the useModerlLayout callback, but they are used in different useModerLayout callouts. Never mind I think it was anyway a premature optimization, it would probably unnecessarily complicate the code.

packages/react/src/hooks/useListNavigation.ts
```diff
@@ -94,6 +96,9 @@ function isCrossOrientationCloseKey(
 ) {
   const vertical = rtl ? key === ARROW_RIGHT : key === ARROW_LEFT;
   const horizontal = key === ARROW_UP;
+  if (orientation === 'both') {
+    return key === ESCAPE;
+  }
```
atomiks, line 101: Should this also work for `horizontal` as well? `horizontal` allows you to press `ARROW_RIGHT` to move to a new row, unlike `both` (e.g. like in emoji pickers). Right now when you press `ARROW_RIGHT` for the submenu trigger, it doesn't move focus into it

mnajdova: Can you clarify what combination of parent + submenu are you talking about? I added more examples to be able to test out more combinations.

atomiks:
Thank you - looks good!

PR #3210, opened 2025-01-30
fix(FloatingList): prevent race condition between registering nodes and setting indices for them

### LINKED ISSUE #3209: FloatingList: elementsRef & map values are sometimes stale with dynamically rendered children
**Describe the bug**

So sorry to file a bug without a reproduction, but came across this issue today and it's incredibly hard to reproduce consistently. Maybe you have an idea of how this can happen, or how it can be avoided.

Here's a screenshot of React DevTools (in Edge/Chrome 132) rendering of FloatingList, which I consider faulty:

Image

- As you can see, 5 children are being rendered
- elementsRef current value includes 9 mixed items, including those rendered children, but also nulls for unrendered children. Additionally, there are some `undefined`s in there, which I have no idea how that can happen with refs (as far as I know, they should always be elements or nulls...).
- The map state includes 10 items, so unrendered children haven't been cleaned up I suppose?

In Firefox 128, the same scenario looks like this:

Image

- elementsRef current value seems to have cleaned itself up
- map state still includes 10 items

When developing locally, I can mostly see that the map state updates according to the elements rendered. elementsRef current value still sometimes includes null values, but they are sorted to the end.

**Expected behavior**

elementsRef.current should only include items that are actually rendered.

**Context:**

- OS: Windows 11
- Browser: Edge 132, Chrome 132, Firefox 128
- React 18.3.1
- floating-ui/react 0.27.3 

**Additional context**

Not sure if this is some kind of a race condition in the component/hook? I don't think I'm doing anything very weird with this, just attempting to collect child element refs and get the sorted, as the component should do.

Does it work when developing locally because React fires effects twice, and that maybe cleans some of this up?

I suppose one workaround for this is to reset FloatingList with a key?

mihkeleidast:
Fixes #3209.

Updating the map state went into a race condition between registering/unregistering nodes (this happens in an effect in useListItem) and setting indices for them (this happens in an effect in FloatingList).

It happens when children rerender, and the two effects are scheduled to run/update state in a non-deterministic way.

Since the indices of the nodes are essentially derived from their refs, the index map can be derived with useMemo instead of updating the same state, removing the possibility of this race condition from happening.

Absolutely not sure about this, it's a pretty weird bug that's hard to reproduce and test. Feel free to reject or suggest alternatives about what may be going on.

### DIFF
--- a/packages/react/src/components/FloatingList.tsx
+++ b/packages/react/src/components/FloatingList.tsx
@@ -63,9 +63,9 @@ export function FloatingList(props: FloatingListProps): React.JSX.Element {
 
   const unregister = React.useCallback((node: Node) => {
     setNodes((prevSet) => {
-      const map = new Set(prevSet);
-      map.delete(node);
-      return map;
+      const set = new Set(prevSet);
+      set.delete(node);
+      return set;
     });
   }, []);
 
--- a/packages/react/src/components/FloatingList.tsx
+++ b/packages/react/src/components/FloatingList.tsx
@@ -58,12 +58,12 @@ export function FloatingList(props: FloatingListProps): React.JSX.Element {
   const [nodes, setNodes] = React.useState(() => new Set<Node>());
 
   const register = React.useCallback((node: Node) => {
-    setNodes((prevMap) => new Set(prevMap).add(node));
+    setNodes((prevSet) => new Set(prevSet).add(node));
   }, []);
 
   const unregister = React.useCallback((node: Node) => {
-    setNodes((prevMap) => {
-      const map = new Set(prevMap);
+    setNodes((prevSet) => {
+      const map = new Set(prevSet);
       map.delete(node);
       return map;
     });
--- a/packages/react/src/components/FloatingList.tsx
+++ b/packages/react/src/components/FloatingList.tsx
@@ -148,7 +148,7 @@ export function useListItem(props: UseListItemProps = {}): {
     [index, elementsRef, labelsRef, label],
   );
 
-  useModernLayoutEffect(() => {
+  React.useEffect(() => {
     const node = componentRef.current;
     if (node) {
       register(node);
--- a/packages/react/src/components/FloatingList.tsx
+++ b/packages/react/src/components/FloatingList.tsx
@@ -129,11 +129,11 @@ export function useListItem(props: UseListItemProps = {}): {
 
   const [index, setIndex] = React.useState<number | null>(null);
 
-  const [componentRef, setComponentRef] = React.useState<Node | null>(null);
+  const componentRef = React.useRef<Node | null>(null);
 
   const ref = React.useCallback(
     (node: HTMLElement | null) => {
-      setComponentRef(node);
+      componentRef.current = node;
 
       if (index !== null) {
         elementsRef.current[index] = node;
@@ -149,21 +149,21 @@ export function useListItem(props: UseListItemProps = {}): {
   );
 
   useModernLayoutEffect(() => {
-    const node = componentRef;
+    const node = componentRef.current;
     if (node) {
       register(node);
       return () => {
         unregister(node);
       };
     }
-  }, [componentRef, register, unregister]);
+  }, [register, unregister]);
 
   useModernLayoutEffect(() => {
-    const index = componentRef ? map.get(componentRef) : null;
+    const index = componentRef.current ? map.get(componentRef.current) : null;
     if (index != null) {
       setIndex(index);
     }
-  }, [map, componentRef]);
+  }, [map]);
 
   return React.useMemo(
     () => ({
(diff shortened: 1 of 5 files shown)


mihkeleidast:
Not sure why the functional tests workflow fails, seems to also fail on master.

packages/react/src/components/FloatingList.tsx
```diff
@@ -149,21 +149,21 @@ export function useListItem(props: UseListItemProps = {}): {
...
       };
     }
-  }, [register, unregister]);
+  }, [componentRef, register, unregister]);
 
   useModernLayoutEffect(() => {
-    const index = componentRef.current ? map.get(componentRef.current) : null;
+    const index = componentRef ? map.get(componentRef) : null;
     if (index != null) {
       setIndex(index);
```
mihkeleidast, line 164: index probably does not need to be a separate state if the ref is state, as then we should be able to derive index from map/ref state values directly?

packages/react/src/components/FloatingList.tsx
```diff
@@ -149,21 +149,21 @@ export function useListItem(props: UseListItemProps = {}): {
   );
 
   useModernLayoutEffect(() => {
-    const node = componentRef.current;
+    const node = componentRef;
     if (node) {
       register(node);
```
mihkeleidast, line 154: An alternative fix that also seemed to work is to filter out elements from the map before sorting that are not in the DOM, i.e. `filter((e) => document.body.contains(e))` or something like that.

Not sure which is a better solution, both work. I like the state ref option better, seems less hacky somehow.

atomiks:
Can you add a test? Important to know when/why the ref solution fails

mihkeleidast:
> Can you add a test? Important to know when/why the ref solution fails

I agree, and I would if I could figure out why this is happening. But as I said, it's a pretty weird behavior that's hard to reproduce outside the app I'm currently in.

Normally, when everything works, events happen in the following order. Let's say I'm rendering 3 items.
1. First item added to map
2. Second item added to map
3. Third item added to map
4. FloatingList renders, with map filled with three items & `null` indices
5. The useLayoutEffect runs that sets the indices
6. FloatingList renders again with map filled with items & indices
7. The useLayoutEffect runs again and bails from updating map

At some point, due to content changing and items being switched out, the following happens:

1. First item added to map
2. Second item added to map
3. FloatingList renders, with map filled with three items & indices
4. The useLayoutEffect that sets the indices runs, but it bails from updating map

So in the faulty scenario, for some reason, last item and indices end up in the map, but I do not see it happening via normal events - I've added a bunch of `console.logs` everywhere and it just does not make any sense at all it skips them.

packages/react/src/components/FloatingList.tsx
```diff
@@ -70,32 +70,30 @@ interface FloatingListProps {
...
+      const map = new Set(prevMap);
       map.delete(node);
       return map;
     });
   }, []);
 
-  useModernLayoutEffect(() => {
-    const newMap = new Map(map);
-    const nodes = Array.from(newMap.keys()).sort(sortByDocumentPosition);
+  const map = React.useMemo(() => {
```
mihkeleidast, line 87: I think this is an even better solution, as it removes the possibility of a race condition happening between registering/unregistering nodes and setting indices for them.

The map with nodes and indices is essentially a derived value from the set of nodes themselves.

Still no test to go with it, I don't know how to test such race conditions.

@atomiks any thoughts on this option?

atomiks: I definitely like the idea of deriving the state instead, especially if it has the same correctness and performance, if you tested that? Also should be `prevSet` not `prevMap` now

mihkeleidast: - It is more correct, since it resolves the race condition issue I ran into.
- Technically it should be more performant, as it does not run the effect in FloatingList twice (previously it bailed out on the second run when if the maps were equal).
 
As far as I can see, other uses of this also behave the same, though I haven't done any major testing of this specifically - how would one go about doing more in-depth testing of this?

Fixed the variable name 👍

mihkeleidast: @atomiks friendly ping, would appreciate if this got sorted today 🙇‍♂️

packages/react/src/components/FloatingList.tsx
```diff
@@ -70,32 +55,30 @@ interface FloatingListProps {
...
+    setNodes((prevSet) => new Set(prevSet).add(node));
   }, []);
 
   const unregister = React.useCallback((node: Node) => {
-    setMap((prevMap) => {
-      const map = new Map(prevMap);
+    setNodes((prevSet) => {
+      const map = new Set(prevSet);
       map.delete(node);
       return map;
```
atomiks, line 68:
```suggestion
      const set = new Set(prevSet);
      set.delete(node);
      return set;
```

PR #3214, opened 2025-02-03
chore: update Playwright

DanielleHuisman:
GitHub recently made Ubuntu 24.04 the default for the `ubuntu-latest` runner. This change causes functional tests to fail with an `apt install` error (e.g. https://github.com/floating-ui/floating-ui/actions/runs/12950634075/job/36123822980).

Updating to the latest Playwright version resolves this issue. Several snapshots had to updated due to slight changes in font rendering.

The Playwright upgrade also exposes an issue with `container-type`, which was reported in #3067. A fix was proposed in #3191. As is, this PR still fails functional tests on this issue, but there are several "solutions":
1. Skip the two failing tests.
2. Update the snapshot of the failing tests.
3. Wait for a fix.

Let me know how to proceed.


atomiks:
Hi @DanielleHuisman sorry for the late response

> Skip the two failing tests

Let's just do this for now. The code will be changed in the future as outlined here: https://github.com/floating-ui/floating-ui/pull/3191#issuecomment-2686981744

DanielleHuisman:
@atomiks I updated the PR to skip those tests.

.changeset/pink-readers-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+"@floating-ui/dom": patch
+---
+
+chore: update Playwright to v1.50.0
```
atomiks, line 5: You can delete this changeset as it's only internal stuff that doesn't affect consumers

DanielleHuisman: Removed.

atomiks:
Thanks for this

PR #3217, opened 2025-02-06
ci: fix release workflow

mihkeleidast:
Since I pinned functional tests runner to `ubuntu-22.04` in #3210, this started failing in master.

Using the ci workflow directly means it should use the same runners that are defined there.

Could also just pin the test job here to 22.04, but it's likely better to use the same setup that's used for running tests on PRs.

Hoping to get #3210 released soon :)


atomiks:
Thank you, makes sense

PR #3222, opened 2025-02-08
Support Ref Cleanup in useMergeRefs

### LINKED ISSUE #3150: [React 19] cleanup function support in `useMergeRefs` hook
React 19 is now stable.

One of the changes in react 19 was that ref callbacks can now return a cleanup function[^1]. This does not work with the [`useMergeRefs` hook](https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/hooks/useMergeRefs.ts) yet.

Here is an example implementation that does support them: https://github.com/vercel/next.js/blob/canary/packages/next/src/client/use-merged-ref.ts

[^1]: see https://react.dev/blog/2024/12/05/react-19#cleanup-functions-for-refs

chirokas:
Closes #3150

### DIFF
--- a/.changeset/tiny-bags-brake.md
+++ b/.changeset/tiny-bags-brake.md
@@ -2,4 +2,4 @@
 "@floating-ui/react": patch
 ---
 
-Support Ref Cleanup in `useMergeRefs`
+fix(useMergeRefs): support React 19 ref cleanup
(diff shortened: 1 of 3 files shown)


atomiks:
Thank you for this - looks pretty convoluted to support the ref cleanups!

PR #3233, opened 2025-02-25
Improve performance by only clearing used timeouts

devbrains-com:
We are using floating-ui in a large react SPA with a lot of tooltips and popovers. On every page transition, we experience a big lag due to clearing timeouts from floating-ui called in the useEffect cleanup method, as you can see in the screenshot.

image

This PR mitigates this performance issue by only calling clearTimeout for timeouts that are actually set before and their ref value is not equal to -1.

We tested it locally with a patch and it works as expected.

Please let me know if you have any suggestions, how to improve or refactor it to meet your requirements.

### DIFF
--- a/packages/react/src/utils.ts
+++ b/packages/react/src/utils.ts
@@ -197,10 +197,3 @@ export function isTypeableCombobox(element: Element | null) {
     element.getAttribute('role') === 'combobox' && isTypeableElement(element)
   );
 }
-
-export function clearTimeoutIfSet(timeoutRef: React.MutableRefObject<number>) {
-  if (timeoutRef.current !== -1) {
-    clearTimeout(timeoutRef.current);
-    timeoutRef.current = -1;
-  }
-}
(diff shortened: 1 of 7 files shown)


packages/react/src/utils.ts
```diff
@@ -197,3 +197,10 @@ export function isTypeableCombobox(element: Element | null) {
     element.getAttribute('role') === 'combobox' && isTypeableElement(element)
   );
 }
+
+export function clearTimeoutIfSet(timeoutRef: React.MutableRefObject<number>) {
+  if (timeoutRef.current !== -1) {
+    clearTimeout(timeoutRef.current);
+    timeoutRef.current = -1;
+  }
+}
```
atomiks, line 206: Can you just make this one a private util? There's a local directory for private ones

devbrains-com: Thank you for the feedback, I moved the function to the local utils folder. 

Local linting is fine. Can you please help me with the pipeline issue? It looks like it tries to lint a generated tsc output file and fails there.

Thank you

teleskop150750:
you can also mark setTimeout as completed at the end

```ts
let id = -1
id = setTimeout(() => {
  foo()
  id = -1
}, 100)
```

PR #3245, opened 2025-03-08
fix(FloatingFocusManager): correct tabbing to different document contexts

### LINKED ISSUE #3238: Tab focus order isn’t correct using Iframe and FloatingPortal.
Tab focus order isn’t correct using Iframe and FloatingPortal.

In our application, we use Iframe to integrate widgets into the host page.
Also, we use FloatingPortal with custom root to avoid render tooltips in the host page and render into our iframe.

But we found problems - when we use `initialFocus={-1}` for FloatingFocusManager, after opening the Tooltip by TAB button, the next focus is not put on the first focusable element in the Tooltip and put on the first focusable on the host page - and as a result, the focusing order is disrupted

If we use `initialFocus` with default value - the focus is put on the first focusable element in the Tooltip, but after focusing on the last element it does not come back to the target element but returns to the first focused element on the host page

**Steps to reproduce the behavior:**

Codesandbox with demo - https://codesandbox.io/p/sandbox/floating-ui-iframe-problem-67s2v5

**If InitialFocus={-1}:**
1. Go to page
2. Start press TAB button
3. Focus the button called “Open popover”
4. Press Enter or Space
5. Press Tab again
6. Focus put on the first focusable element on the host page

https://github.com/user-attachments/assets/6e862990-083c-4b2e-ab68-10e107a8b2e9

**If InitialFocus=default:**
1. Go to page
2. Start press TAB button
3. Focus the button called “Open popover”
4. Press Enter or Space
5. Press Tab again and finish focusing element in the Tooltip
6. After the last element, focus put on the first focusable element on the host page

https://github.com/user-attachments/assets/f9c2ec54-b1ca-4193-aacd-d56d7adc4dc0

**Expected behavior** 
**If InitialFocus={-1}**
1. Next focus put on the first focusable element on the Tooltip
2. After focusing the last focusable element on the Tooltip, focus move to the next focusable element in Iframe(“Link in iframe 2” - in demo)

**If InitialFocus=default**
1. After focusing the last focusable element on the Tooltip, focus move to the next focusable element in Iframe(“Link in iframe 2” - in demo)

**In our case we would like to prefer InitialFocus={-1}**

**Context:**

- OS: macOS 15.2
- Browser: chrome
- Version 133.0.6943.98 

**Additional context** 
I think the problem appears in [this place](https://github.com/floating-ui/floating-ui/blob/master/packages/react/src/utils/tabbable.ts#L35), because we found elements in the root document, but iframe contains its own document context.

atomiks:
Fixes #3238

### DIFF
--- a/.changeset/orange-rings-remember.md
+++ b/.changeset/orange-rings-remember.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(FloatingFocusManager): correct tabbing to different document contexts
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -806,7 +806,7 @@ export function FloatingFocusManager(
             ) {
               preventReturnFocusRef.current = false;
               if (isOutsideEvent(event, portalContext.portalNode)) {
-                const nextTabbable = getNextTabbable(floating, domReference);
+                const nextTabbable = getNextTabbable(domReference);
                 nextTabbable?.focus();
               } else {
                 portalContext.beforeOutsideRef.current?.focus();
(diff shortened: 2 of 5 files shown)


vladislav-manakov-1:
@atomiks hi! when will the version with these changes be released? in the current version 0.27.5 I don't see the changes from this PR
[getPreviousTabbable](https://www.npmjs.com/package/@floating-ui/react/v/0.27.5?activeTab=code) function still has the old implementation

atomiks:
@vladislav-manakov-1 I'm sorry but I haven't done a release with it yet. I'm going to try to do one today

PR #3271, opened 2025-04-03
fix(useListNavigation): Fix list navigation for nested lists with mixed orientation

michaldudak:
Keyboard navigation for nested lists with different orientation (like a horizontal menubar with vertical menus) works somewhat inconsistently. When a submenu item is focused, pressing <kbd>RightArrow</kbd> closes the submenu and focuses the next item from the parent. Pressing the <kbd>LeftArrow</kbd>, however, closes the submenu and focuses its trigger instead of the previous item.

In this PR, I detect whether the close key of a nested list is also a navigation key of the parent list. If so, the keyboard event isn't stopped, and the parent has a chance to focus on another item.

~~I'm not sure what should happen when the open submenu's trigger is the first item within the parent and `loop = false`. This solution makes the submenu stay open, as the parent won't change focus. Alternatively, we could call `onOpenChange(false, event.nativeEvent, 'list-navigation');` to make sure the submenu is closed.~~

EDIT:
After more testing, I decided that in case where an open submenu is the first item within its parent, it's more natural just to close the submenu and leave focus on its trigger than to leave it open.

### DIFF
--- a/packages/react/src/hooks/useListNavigation.ts
+++ b/packages/react/src/hooks/useListNavigation.ts
@@ -583,9 +583,10 @@ export function useListNavigation(
     virtual,
   ]);
 
-  const parentOrientation: UseListNavigationProps['orientation'] =
-    tree?.nodesRef.current.find((node) => node.id === parentId)?.context
-      ?.dataRef?.current.orientation;
+  const getParentOrientation = React.useCallback(() => {
+    return tree?.nodesRef.current.find((node) => node.id === parentId)?.context
+      ?.dataRef?.current.orientation as UseListNavigationProps['orientation'];
+  }, [parentId, tree]);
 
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/useListNavigation.ts
```diff
@@ -583,6 +583,10 @@ export function useListNavigation(
     virtual,
   ]);
 
+  const parentOrientation: UseListNavigationProps['orientation'] =
+    tree?.nodesRef.current.find((node) => node.id === parentId)?.context
+      ?.dataRef?.current.orientation;
```
atomiks, line 588: Refs can't be read in render

michaldudak: Right, fixed.

Issue #3260, opened 2025-03-23
Prepare new major versions


atomiks:
## `@floating-ui/vue` v2

[Drop Vue 2 support](https://github.com/floating-ui/floating-ui/discussions/3223) as it reached EoL awhile ago

## `@floating-ui/react` v1

Properly stabilize this package and refine APIs. I've kept it mostly backwards-compatible for 2 years now but it's showing some limitations.

It can be lighter/more tree-shakeable, and the APIs can be less confusing. Remove deprecations and tech debt.

## `@floating-ui/utils` v1

This could be refined, stabilized and properly documented

## `@floating-ui/core,dom,react-dom,react-native`

With CSS anchoring becoming more widespread in the future, it doesn't make sense to change anything/release a new major for these. The APIs have some minor issues, but these can be worked around externally easily (such as supporting logical sides without the confusing `placement` string)

kiran-bhalerao:
@atomiks when can we expect the @floating-ui/react v1 release? im about to start a new project, should i use the current version of @floating-ui/react or should i wait for v1?

atomiks:
Hey @kiran-bhalerao no timeline atm but before the end of the year, so there's no need to wait. 

I'm trying to keep the changes small (prop/config-related) that will be easy to migrate to. Most of the changes will be to improve default behavior (such as #3138), performance/bundle size, etc, rather than fundamental API changes.

barsukov:
Hello, is it any possibility to know that React 19 upgrade will be included soon as part of major version or you plan to ship the lib with a minor update?

teleskop150750:
you might find my attempt to port `floating-ui` to vue useful.
`chore/core`  - actual branch
[floating-vue](https://github.com/perigee-ui/floating-vue/tree/chore/core)

atomiks:
v2 is no longer planned

teleskop150750:
> > Base UI uses @floating-ui/react-dom directly, but it forked @floating-ui/react and they've diverged significantly
> 
> [@atomiks](https://github.com/atomiks) Aren't you part of base ui team? Could you add more context and details on base ui fork intentions and outcomes?

There is not much source code, you can quickly compare both versions with your eyes.

PR #3274, opened 2025-04-06
feat: `NextFloatingDelayGroup`

atomiks:
This exports a new `FloatingDelayGroup` API which doesn't cause all the consumers of `useDelayGroup` to re-render if they don't have to (using refs instead of state). `useHover#delay` and `useHover#restMs` both now can be a function so you can return the `ref.current` values, while still peforming custom calculations before passing the value in.

Migration strategy:
- Next minor: `FloatingDelayGroup` soft-deprecated (`@deprecated` jsdoc only, no warning) -> move to `NextFloatingDelayGroup`. Documentation will be added at this point.
- v1 release: Rename `NextFloatingDelayGroup` back to `FloatingDelayGroup` (with the new implementation)

Part of #3260 

### Usage

```jsx
const delayGroup = useNextDelayGroup(context);

const hover = useHover(context, {
  delay: delayGroup.hasProvider // explicitly tells you whether a <FloatingDelayGroup> exists
    ? () => delayGroup.delayRef.current 
    : delay // from e.g. a prop, fallback
});

// `delayGroup.isInstantPhase` is also now smarter
useTransitionStyles(context, {
  duration: delayGroup.isInstantPhase ? 0 : 200
});
```

### DIFF
--- a/.changeset/shiny-days-vanish.md
+++ b/.changeset/shiny-days-vanish.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+feat: `FloatingDelayGroupOptimized` (and `useDelayGroupOptimized`). Unlike `FloatingDelayGroup`, this component doesn't cause a re-render of unrelated consumers of the context when the delay changes, improving performance. This will become the new default in v1 while the previous implementation will be deprecated.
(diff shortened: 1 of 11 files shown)


sins921 (approved):

atomiks:
@mihkeleidast yeah I remember `FloatingGroup` - a more general component could be useful but it has a code size tradeoff where splitting it into several new components/providers is probably better anyway, e.g. `FloatingDelayGroup`, `FloatingLayoutGroup`, `FloatingGroup` itself could just be for managing events, etc.

PR #3277, opened 2025-04-07
fix(useHover): clean up blockPointerEvents when opened with click after hover

mihkeleidast:
Hope the test kind of explains itself, but putting it here as well: if a floating element has both hover and click hooks, pointer events are blocked on reference hover. If the user then clicks the reference element, it switches off the mouseleave handlers, instead waiting for a outside click or some other dismiss event. But, as currently the pointer events remain blocked, this breaks when such floating components are nested - clicking on the parent floating element when pointer events are blocked also closes the parent, as it considers it an outside press then.

Fairly easy to reproduce in the playground Popover component:
1. Open the Popover playground
2. Turn off modal focus management
3. Enable hover on the second level Popover in source
4. Open the first level Popover
5. Open the second level Popover via hover, then click the reference element (this "switches" useHover to a "click" state)
6. Click first level Popover - see that both popovers close, even though it should only close the second level popover.

The lint/typecheck failure does not seem to be my fault?

### DIFF
--- a/packages/react/test/unit/useHover.test.tsx
+++ b/packages/react/test/unit/useHover.test.tsx
@@ -4,6 +4,9 @@ import {vi, test} from 'vitest';
 
 import {useFloating, useHover, useInteractions} from '../../src';
 import type {UseHoverProps} from '../../src/hooks/useHover';
+import {Popover} from '../visual/components/Popover';
+import {Button} from '../visual/lib/Button';
+import userEvent from '@testing-library/user-event';
 
 vi.useFakeTimers();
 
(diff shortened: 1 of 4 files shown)


packages/react/test/unit/useHover.test.tsx
```diff
@@ -304,3 +307,65 @@ test('reason string', async () => {
...
+  await user.click(screen.getByText('Open parent'));
+  expect(screen.getByText('Parent title')).toBeInTheDocument();
+  await user.click(screen.getByText('Open child'));
+  expect(screen.getByText('Child title')).toBeInTheDocument();
+  await user.click(screen.getByText('Child title'));
+  // clean up blockPointerEvents
+  // userEvent.unhover does not work because of the pointer-events
+  fireEvent.mouseLeave(screen.getByRole('dialog', {name: 'Child title'}));
+  expect(screen.getByText('Child title')).toBeInTheDocument();
+  await user.click(screen.getByText('Parent title'));
```
mihkeleidast, line 366: Not super happy with this test, as it's hard to click on this element to showcase closing when it should not - instead without the useHover change, it fails with "... unable to perform interaction due to blocked pointer events" or something like that. If I disable pointer events checks, it just succeeds in test env, even though it fails in browser then.

Suggestions welcome.

atomiks: Good enough :)

PR #3286, opened 2025-04-09
Fix a bug `useListNavigation` no longer correctly scroll to the selected index

chuganzy:
As titled, this PR fixes a bug `useListNavigation` does not scroll to the selected index on open.

It is because `forceScrollIntoViewRef.current` is set to false right after `focusItem` is called while scroll [can run asynchronously](https://github.com/floating-ui/floating-ui/blob/46f6464cc38b30aa63209ea4eb9f5f413fe6f777/packages/react/src/hooks/useListNavigation.ts#L460).

Related: https://github.com/floating-ui/floating-ui/pull/3141

### DIFF
--- a/.changeset/beige-scissors-work.md
+++ b/.changeset/beige-scissors-work.md
@@ -1,0 +1,5 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(useFloating): correctly scroll to the selected index on open
--- a/.changeset/beige-scissors-work.md
+++ b/.changeset/beige-scissors-work.md
@@ -2,4 +2,4 @@
 "@floating-ui/react": patch
 ---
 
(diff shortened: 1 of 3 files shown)


atomiks:
I've never encountered it not scrolling into view, but I can see why the asynchronicity could cause an issue; how'd you encounter it?

chuganzy:
@atomiks hi, thank you for taking a look. it seems to happen when the floating list was triggered by non-keyboard (ex: click).

atomiks:
Ah I see, it happens if you don't have `FloatingFocusManager` (`initialFocus`) in use which is able to focus the item in spite of `scrollIntoView` not actually taking effect on initial open

chuganzy:
ah, that makes sense! thank you for reviewing and merging this quickly!

PR #3289, opened 2025-04-10
fix(useRole): improve `"combobox"` role

atomiks:


packages/react/src/hooks/useListNavigation.ts
```diff
@@ -330,6 +330,9 @@ export function useListNavigation(
   const focusItem = useEffectEvent(() => {
     function runFocus(item: HTMLElement) {
       if (virtual) {
+        if (item.id?.endsWith('-fui-option')) {
+          item.id = `${floatingId}-${Math.random().toString(16).slice(2, 10)}`;
+        }
```
chuganzy, line 335: This change assumes that the active ID only changes when navigating, however in reality, the list itself can update independently, resulting in a possible mismatch between the active ID and the actual ID..

atomiks: Sorry for the bug. The previous method didn't work in Safari, only Chrome. For a fix, the `aria-activedescendant` attr needs to change on every item highlight change.

atomiks: Tried a few things, but I don't think this bug is fixable in the current API with the root having `getItemProps`, since only one `useId` can be called in this scenario. 

I had the idea of using `activeIndex` in the item prop getter to add the `id`, but due to perf reasons, `activeIndex` can't be in the `item` dependencies.

Each item needs its own `id` which is per-component. This means you need to supply an `id` in your Item component manually to workaround this.

Also fwiw, I'm not sure the mismatch is necessarily a real problem - I think if the list changes, the active index should be set to `null`, because screen readers won't announce that the focused item has changed when the list has changed otherwise.

PR #3296, opened 2025-04-15
(feat: useListNavigation): add the `parentOrientation` parameter

michaldudak:


atomiks:
Let's just add the prop I guess. The API for `useListNavigation` can be refined for v1 #3260

PR #3297, opened 2025-04-16
feat(react): export most utils

atomiks:


ApenasGabs (approved):

packages/react/src/utils.ts
```diff
@@ -1,199 +1,9 @@
...
-}
-
-export function isTypeableCombobox(element: Element | null) {
-  if (!element) return false;
-  return (
-    element.getAttribute('role') === 'combobox' && isTypeableElement(element)
-  );
-}
+import * as composite from './utils/composite';
+import * as tabbable from './utils/tabbable';
```
atomiks, line 2: Neither of these are tree-shakeable due to this convenient namespacing (without changes to the dist files, which seems to be complex to change).

The module structure needs to be preserved instead of Rollup bundling them into the main file as an object, preventing bundlers from removing unused code.

`Composite`, `useListNavigation`, and `FloatingPortal` access *all* of the methods/properties on them anyway, so treeshaking doesn't matter internally. For external consumers that aren't using those, it doesn't work properly, but neither of these modules are too large, so I think it's okay for now

PR #3327, opened 2025-05-25
[useListNavigation] fix: avoid item rerenders

romgrk:
The dependency on `open` in event handlers forces all items to re-render. Those event handlers should probably be stable callbacks, but I don't want to change too much here.


atomiks:
Usually the items all have to re-render upon the initial open anyway - is this change for when it's kept mounted in the DOM after the first render (as in Base UI master currently?), or do they re-render after mounting as well?

romgrk:
They do re-render after mounting, the first time `Select` is opened. There's many causes for those re-renders, one of them being the change in `floatingRootContext` which originates from this point.

PR #3344, opened 2025-06-08
feat!(next): create async and sync versions

atomiks:
Runtime alternative to #3343

### DIFF
--- a/.changeset/purple-insects-wait.md
+++ b/.changeset/purple-insects-wait.md
@@ -1,0 +1,6 @@
+---
+'@floating-ui/core': major
+'@floating-ui/dom': major
+---
+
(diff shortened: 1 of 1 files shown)


packages/core/src/computePosition.runners.ts
```diff
@@ -0,0 +1,39 @@
...
+  const gen = computePositionGen(reference, floating, config);
+  let step = gen.next();
+
+  while (!step.done) {
+    const result = await step.value;
+    step = gen.next(result);
+  }
+
+  return step.value;
+}
```
atomiks, line 39: A rare but useful pattern: you can support async (e.g. React Native) and sync (e.g. DOM) versions with only one codebase (without duplicating) by using generators. So internally `await` becomes `yield` but the runners can turn each yield into an `await` step for the async version, while the sync version just executes instantly.

PR #3347, opened 2025-06-15
refactor(next)!: split `placement` into `side` and `align`

atomiks:
The current `placement` API has a few problems:

- It doesn't support logical sides without being confusing (e.g. `inline-end-start`) because the alignment is part of the string in kebab form
- Specifying a side without an alignment like `top` implicitly adds a `center` alignment which is good as a shortform but can be confusing that it's not explicitly added (`type Alignment = 'start' | 'end' | null`)

Splitting `placement` into `side` and `align` is the same API that Radix and Base UI are using and it works very well.

`Placement` is now an object in the form `{ side: Side; align: Align }`, which removes the need to perform string parsing and is simpler; this is not a fully public API aside from being present in middleware return values though.

_Most likely_ there will be a backwards-compat config object to retain the old `placement: string` API so migration from v1 is easier, which just transforms 

```js
createLegacyConfig({
  placement: 'top-start',
});
```

Into:

```js
{
  side: 'top',
  align: 'start',
}
```


FezVrasta:
align may be a bit ambiguous when taken alone.

atomiks:
@FezVrasta updated the PR description

`side` alone was already ambiguous since it implicitly attached a `center` alignment
`align` is straightforward when passed alone if you know the default, same as `side` alone having a  `center` align by default

This API works well with logical sides, and omitting either feels cleaner:
```js
<Tooltip.Positioner side="block-start" align="end" />
<Tooltip.Positioner side="left" /> {/* center align by default */ }
<Tooltip.Positioner align="start" /> {/* top side by default */ }
```

sins921 (approved):

PR #3399, opened 2025-09-21
feat(autoUpdate): allow not passing a floating element

maesse:
This PR updates `autoUpdate` to allow for `floating` to be `null`

dom/src/autoUpdate.ts:
```diff
function autoUpdate(
  reference: ReferenceElement,
- floating: FloatingElement,
+ floating: FloatingElement | null,
  update: () => void,
  options: AutoUpdateOptions = {},
)
```

**Motivation:**
This enables use cases where you only want to track changes to the reference element without necessarily having a floating element.

**Impact:**
No breaking changes.
Expands the function’s flexibility for broader scenarios.

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -181,7 +181,7 @@ export function autoUpdate(
   let reobserveFrame = -1;
   let resizeObserver: ResizeObserver | null = null;
 
-  if (elementResize && (referenceEl || floating)) {
+  if (elementResize) {
     resizeObserver = new ResizeObserver(([firstEntry]) => {
       if (
         firstEntry &&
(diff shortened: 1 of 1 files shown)


maesse:
Hi atomiks,

I'm basically using only the positional tracking, because my use case is rendering a bunch of lines in an SVG overlay that connects DOM elements. 

Do you think it could be useful to expose that functionality? My thinking was, it's already there, the tracking works great, with a few tweaks to the interface it would allow for cases where it's necessary to do custom rendering (or maybe just tracking positions alone for whatever reason). I was looking for libraries to solve this issue and this was the closest/best I found.

I do realize though that it might be a bit outside the core purpose of the library, and my PR might not be the cleanest solution. I'll admit that it's mostly for my own convenience that I submitted this PR, since it would save me the hassle of having a fork :)

PR #3423, opened 2026-01-05
feat: detectOverflow configurable via platform

Juniper-Eldred:
This MR is intended to allow custom implementations of the `detectOverflow` function used by Floating UI's builtin middleware

Closes https://github.com/floating-ui/floating-ui/discussions/3422

PR #3434, opened 2026-02-18
refactor: update DOM Platform typing to include detectOverflow

Juniper-Eldred:
This MR updates `@floating-ui/dom`'s `Platform` interface to ensure that `detectOverflow` (added by [this PR](https://github.com/floating-ui/floating-ui/pull/3423)) is available

### DIFF
--- a/.changeset/beige-trainers-act.md
+++ b/.changeset/beige-trainers-act.md
@@ -2,4 +2,4 @@
 '@floating-ui/dom': patch
 ---
 
-Change dom Platform type to be a passthrough to the core Platform type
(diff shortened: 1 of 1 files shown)


packages/dom/src/types.ts
```diff
@@ -58,7 +59,7 @@ export type OffsetValue =
 // backwards-compatibility, re-define it here to use the DOM Derivable type.
 export type OffsetOptions = OffsetValue | Derivable<OffsetValue>;
 
-export interface Platform {
+export type Platform = {
```
atomiks: What's the reason to change to `type`? Theoretically it's a breaking change depending how the type was used

PR #3444, opened 2026-02-25
fix(react): avoid duplicate "options" key in middleware object literals

kirkouimet:
## Summary

The reactive middleware wrappers in `@floating-ui/react-dom` spread the base middleware object — which contains an `options` property — and then immediately overwrite `options` with a React-specific tracking tuple:

```typescript
export const offset = (options?, deps?) => ({
  ...baseOffset(options),     // spreads { name, options, fn }
  options: [options, deps],   // duplicate "options" key
});
```

This produces object literals with duplicate `options` keys. While functionally correct (the last key wins per spec), it triggers esbuild's `duplicate-object-key` warning — once for every middleware instance in the bundle.

For anyone deploying through esbuild-based tooling (Cloudflare Wrangler, Vite, etc.), this means 6+ warnings on every single build, for code they can't control. Over time, warning fatigue sets in and real warnings get missed.

## Fix

Destructure out the `options` property before spreading, so it's only assigned once:

```typescript
export const offset = (options?, deps?) => {
  const {options: _, ...rest} = baseOffset(options);
  return {...rest, options: [options, deps]};
};
```

Same mechanical change applied to all 9 middleware wrappers: `offset`, `shift`, `limitShift`, `flip`, `size`, `autoPlacement`, `hide`, `inline`, `arrow`.

## Verification

- All 15 existing tests pass
- TypeScript type checking passes
- Build succeeds with Rollup
- Prettier formatting unchanged
- Built output confirms `options` appears only once per middleware object

Thank you for Floating UI — it's a beautifully designed library that quietly powers positioning for a huge slice of the web. This is a small fix, but it'll make a lot of CI logs a little cleaner.

### DIFF
--- a/packages/react-dom/src/reactiveMiddleware.ts
+++ b/packages/react-dom/src/reactiveMiddleware.ts
@@ -34,8 +34,12 @@ export const offset = (
   options?: OffsetOptions,
   deps?: React.DependencyList,
 ): Middleware => {
-  const {options: _, ...rest} = baseOffset(options);
-  return {...rest, options: [options, deps]};
+  const result = baseOffset(options);
+  return {
+    name: result.name,
+    fn: result.fn,
+    options: [options, deps],
+  };
 };
 
 /**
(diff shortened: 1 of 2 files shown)


kirkouimet:
@atomiks Done! Updated to use explicit property access and applied the same fix to react-native. Thanks for the quick review.

PR #3446, opened 2026-02-26
fix(react): fix shadow root host lookup when one of the parent elements is an anchor

matoilic:
When traversing the parents of a floating element the `host` property was read to traverse from a shadow root to its host element. But it was not properly checked that an element is a shadow root before reading the host property. 
That lead to the issue that if one of the parent elements is an anchor, it would read the `host` property of the a tag which contains the host of the URL the anchor links to. Consequently, it would perform Element#contains calls with a string instead of a `HTMLElement`.

### DIFF
--- a/packages/react/src/utils/markOthers.ts
+++ b/packages/react/src/utils/markOthers.ts
@@ -24,19 +24,13 @@ let lockCount = 0;
 export const supportsInert = (): boolean =>
   typeof HTMLElement !== 'undefined' && 'inert' in HTMLElement.prototype;
 
-const unwrapHost = (
-  node: Element | ShadowRoot | null,
-): Element | null => {
-  if (node == null) {
+function unwrapHost(node: Node | null): Element | null {
+  if (!node) {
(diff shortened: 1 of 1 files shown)


packages/react/test/unit/markOthers.test.ts
```diff
@@ -164,3 +164,21 @@ test('mixed controlAttribute usage (aria-hidden/inert/none)', () => {
...
+
+  const shadowRoot = host.attachShadow({mode: 'open'});
+  const anchor = document.createElement('a');
+  anchor.href = 'https://floating-ui.com';
+
+  const target = document.createElement('button');
+  anchor.appendChild(target);
+  shadowRoot.appendChild(anchor);
+
+  expect(() => markOthers([target], true)).toThrow();
```
atomiks: This test shows the solution has an infinite recursion bug

atomiks: No problem!

PR #3447, opened 2026-02-26
fix(react): ignore ghost mouseenter events from portaled overlay unmounts in useHover

fominv:
## Problem                                                               

When a portaled overlay (drawer, modal, dialog) unmounts and an element underneath happens to overlap the last touch position, that element's tooltip opens unexpectedly — even though the touch originally targeted a different element (e.g. a close button in the overlay).

## Cause

When a portal unmounts, the browser re-evaluates hit-testing and fires a synthetic mouseover → mouseenter on elements now exposed at the last touch position. This mouseenter is not preceded by a pointerenter (no active pointer after touch ends), but useHover treats it as a real hover.

## Fix

useHover already tracks pointerTypeRef via onPointerEnter. For real interactions it's always set before mouseenter fires. For ghost events it stays undefined. The fix adds a guard in onReferenceMouseEnter:

`pointerTypeRef.current == null ||`

Tests updated to fire pointerEnter before mouseEnter, matching real browser behavior. Added a test verifying bare mouseenter (ghost event) doesn't open the floating element.

## Browser events on portal unmount (confirmed on Firefox mobile simulator)

mouseover & mouseenter

No pointer events (pointerover, pointerenter) fire.

### DIFF
--- a/packages/react/src/hooks/useHover.ts
+++ b/packages/react/src/hooks/useHover.ts
@@ -249,6 +249,13 @@ export function useHover(
       blockMouseMoveRef.current = false;
 
       if (
+        // When a portal (e.g. a drawer or modal overlay) unmounts, the browser
+        // re-evaluates hit-testing and fires a synthetic mouseenter on elements
+        // now exposed at the last pointer position. These "ghost" mouseenter
+        // events are NOT preceded by a pointerenter (because there is no active
+        // pointer after the touch ends), so pointerTypeRef is still undefined.
+        // Ignore these to prevent tooltips from opening unexpectedly.
+        pointerTypeRef.current == null ||
         (mouseOnly && !isMouseLikePointerType(pointerTypeRef.current)) ||
         (getRestMs(restMsRef.current) > 0 &&
           !getDelay(delayRef.current, 'open'))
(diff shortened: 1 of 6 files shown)


fominv:
Thanks for the review.

> ## Codex Findings
> 
> P1: `floating-ui/packages/react/src/hooks/useHover.ts:258` makes hover opening contingent on prior pointer events, but pointer type is only captured in `floating-ui/packages/react/src/hooks/useHover.ts:499`. In browsers/contexts without pointer events, `useHover` will no longer open on real mouse hover.

> makes hover opening contingent on prior pointer events

Yes, that is the point. Maybe I am misunderstanding something, but are there any valid contexts where a prior pointer event is not emitted? In any recent browser the pointer event is always emitted beforehand.

> P2: Test updates to `floating-ui/packages/react/test/unit/utils.ts:3` and pointer-event polyfill in `floating-ui/packages/react/test/unit/setupTests.ts:65` remove coverage of mouseenter-only behavior, so this regression is currently untested.

Again, the point is to remove mouseenter-only behavior. 

> so this regression is currently untested.

I added a test to make sure that it is not triggered when only mouseenter-only fires. So it is tested, but the test makes sure that the floating element does not open / render.

This is actually our main goal. We use this library to render tooltips and we have another overlay in front. It so happens that that overlay renders another clickable element. When we click it and the overlay closes. The tooltip exactly underneath is also triggered, which is an undesired, but currently unavoidable side effect.

Maybe our intent here is still unclear. Should I add a stack blitz or other form of reproduction?

PR #3457, opened 2026-04-01
fix(FloatingFocusManager): suppress :focus-visible on pointer-initiated close

### LINKED ISSUE #3394: FloatingFocusManager applies :focus-visible pseudo class to reference element when the floating element is closed
**Describe the bug** A clear and concise description of what the bug is.

When a Dropdown Menu is closed (for example, after an option is selected via mouse pointer event), the `FloatingFocusManager` returns the focus back to the reference element. This is expected, however, we are finding that in some cases, the `:focus-visible` pseduo-class is being applied to the reference element when the focus is returned. Ideally, the `:focus-visible` pseduo-class should only be applied when the dropdown menu has been closed using a keyboard!

**To Reproduce**

1. Go to [floating ui docs home page](https://floating-ui.com/)
2. Click on the trigger element in the `Dropdown Menus` example.
3. Select an option using a mouse click.
4. See that the outline is applied around the reference element when the menu is closed. Upon inspecting the element, you will find that a `:focus-visible` pseudo-class is applied to the reference element.

**Expected behavior** A clear and concise description of what you expected to
happen.

The expected behavior should be that the `:focus-visible` pseudo-class should only be applied when the menu is closed using keyboard navigation and a `:focus` pseudo class for mouse events.

Image

**Context:**

- OS: MacOs
- Browser: Chrome
- Version: 140.0.7339.133

**Additional context** Add any other context about the problem here.

Gdhanush-13:
### What

Fixes #3394

### Problem

When a floating element (e.g. a dropdown menu) is closed via a **pointer interaction** (clicking a menu item, outside press), `FloatingFocusManager` returns focus to the reference element. The browser incorrectly applies the `:focus-visible` pseudo-class to the reference element, causing an unwanted focus ring.

Ideally, `:focus-visible` should only appear when the floating element was closed via **keyboard** (e.g. pressing Escape).

### Root Cause

The `tabbableReturnElement.focus({preventScroll: true})` call in the cleanup function does not specify `focusVisible`, so the browser's heuristic decides whether to apply `:focus-visible`. Due to the microtask timing of the focus call, some browsers apply `:focus-visible` even for pointer-initiated closes.

### Fix

1. Added `isKeyboardCloseRef` to track whether the close was keyboard-initiated (`reason === 'escape-key'` or `event instanceof KeyboardEvent`).
2. When returning focus after a **pointer-initiated** close, pass `focusVisible: false` to the `.focus()` call.
3. When returning focus after a **keyboard-initiated** close (Escape), omit `focusVisible` to let the browser apply its default `:focus-visible` behavior.

```diff
 tabbableReturnElement.focus({
   preventScroll: true,
+  // Suppress :focus-visible for pointer-initiated closes. GH #3394
+  ...(isKeyboardCloseRef.current
+    ? undefined
+    : {focusVisible: false}),
 });
```

### DIFF
--- a/.changeset/fix-focus-visible-on-return.md
+++ b/.changeset/fix-focus-visible-on-return.md
@@ -1,0 +1,7 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(FloatingFocusManager): suppress :focus-visible on pointer-initiated close
+
+When a floating element is closed via a pointer interaction (e.g. clicking a menu item), the returned focus on the reference element no longer incorrectly shows :focus-visible styling. Keyboard-initiated closes (e.g. Escape) still correctly apply :focus-visible.
--- a/packages/react/src/components/FloatingFocusManager.tsx
+++ b/packages/react/src/components/FloatingFocusManager.tsx
@@ -261,6 +261,7 @@ export function FloatingFocusManager(
   const startDismissButtonRef = React.useRef<HTMLButtonElement>(null);
   const endDismissButtonRef = React.useRef<HTMLButtonElement>(null);
   const preventReturnFocusRef = React.useRef(false);
+  const isKeyboardCloseRef = React.useRef(false);
   const isPointerDownRef = React.useRef(false);
   const tabbableIndexRef = React.useRef(-1);
   const blurTimeoutRef = React.useRef(-1);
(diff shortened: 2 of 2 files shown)


ganor38:
Any updates on this? 🙏🏻

Gdhanush-13:
Still open and ready for review! The fix suppresses the :focus-visible ring on FloatingFocusManager when the close was pointer-initiated (click/touch), so keyboard users still get visible focus indicators while mouse users don't see the flash. Happy to rebase or adjust anything if needed.

PR #3468, opened 2026-06-10
fix(core): prefer scrollable overflow direction in flip `bestFit` fallback

### LINKED ISSUE #3014: flip with `boundary` gets clipped when it can't fit within `rootBoundary`
**Describe the bug**

A floating element with `flip({ boundary: document.body, roootBoundary: 'viewport' })` is ending up positioned so it is clipped by the document (or any other custom clipping boundary) when it cannot fit within the viewport.

**To Reproduce**
[Codesandbox link](https://codesandbox.io/p/sandbox/floating-ui-dom-template-forked-g958sd)

The codesandbox shows two examples -- one where the clipping ancestor is `document.body`, and one within a custom scroll region.

For the one within a scroll region, it is positioned above the trigger and clipped by the scroll boundary when scrolled all the way to the top.

For the one where the clipping boundary is `document.body`, to repro do the following:
- Set the viewport height to 225px or less
- scroll all the way to the top
- the floating element is clipped by the document:
Screenshot of a button with a block of white text against a black background above it. The text is cut off at the top.

**Expected behavior** 

Ideally I'd like the floating element to position itself to the best fit based on the viewport, unless that position causes it to be clipped by the document. From the docs, I thought `boundary` would do that, but please let me know if I'm misunderstanding what that prop is intended to do.

**Context:**

- OS: macOS 14.6
- Browser Chrome, Edge, Firefox, Safari (same behavior on all 4)

chatman-media:
# fix(core): prefer scrollable overflow direction in flip `bestFit` fallback

## Problem

Closes #3014

When the floating element cannot fit on **any** candidate placement, `flip`'s `bestFit` fallback picks the placement with the least total positive overflow. That metric is blind to *which direction* the overflow goes, and the two directions are not equivalent:

- Overflow toward `bottom` (or `right` in LTR, `left` in RTL) extends the boundary's scrollable area — the user can scroll the clipped content into view.
- Overflow toward `top` (or the inline scroll-origin side: `left` in LTR, `right` in RTL) displaces the floating element *behind* the scroll origin, where it is clipped and can **never** be scrolled into view.

So when a tall floating element doesn't fit on either side of the reference, "least overflow" can pick the placement whose excess content is permanently unreachable, even though a slightly-more-overflowing placement would have kept all content reachable by scrolling. This is the top/left clipping bias described in #3014.

## Fix

In the `bestFit` fallback, reachability is now the **primary** sort key, with least total overflow as the secondary key:

```js
.sort((a, b) => Number(a[1]) - Number(b[1]) || a[2] - b[2])[0]?.[0];
```

where slot `[1]` is "the placement's main-axis overflow goes toward the clipped scroll-origin side (`top`, or `left`/`right` per direction)" and slot `[2]` is the total positive overflow. A placement that overflows toward the scroll origin loses to one that overflows toward the scrollable direction **even when its total overflow is smaller** — by construction, content clipped behind the scroll origin is worth less than content reachable by scrolling, regardless of the pixel amounts. Among placements on the same side of that divide, least total overflow still decides.

The bias only considers the main axis and is skipped when `mainAxis: false`, since cross-axis overflow is what `shift()` handles. In RTL the inline scroll origin is the `right` side, so the preference flips accordingly.

> **Note:** because the bias is a primary key rather than a tie-breaker, `bestFit` results can change for existing code whenever no placement fits and the candidates overflow in opposite main-axis directions — not just on exact ties. That is the intended behavior change for #3014.

## Interaction with `size()`

The "unreachable overflow" premise only holds while the floating element keeps its natural size. When `size()` actually shrinks the element to fit, the overflow is recoverable by resizing (the element becomes scrollable itself), so the best-fitting placement should win with **no** directional bias — otherwise the bias would needlessly move, say, a combobox popup below the reference when it could fit fine above at a reduced height.

Merely having `size()` in the middleware chain proves nothing, though: `middlewareData.size` only shows the middleware *ran*, not that the element was constrained. Common patterns make this distinction load-bearing — e.g. Base UI-style usage where `apply()` just exposes `--available-height`/`--available-width` CSS variables that the consumer's styles may never consume, or constrains only the cross axis while the main-axis overflow stays unrecoverable.

So `flip` now persists `middlewareData.flip.fallbackDimension` — the floating element's main-axis dimension at the moment the `bestFit` fallback placement was chosen — and re-resolves the placement neutrally (no reachability bias) only when a later pass observes that `size()` has actually shrunk that dimension. If `size()` runs but the overflowing axis keeps its size, the scrollable-direction bias is kept.

This gate also covers non-scrollable clipping boundaries (`overflow: hidden`/`clip`), where the reachability premise wouldn't hold: a floating element inside such a container that fits nowhere needs `size()` regardless (the overflow was silently cut off before this change too), and once `size()` constrains it, the gate neutralizes the bias and the least-overflowing placement wins — now fully fitting instead of being clipped.

`fallbackDimension` is a new optional field on `MiddlewareData['flip']` (types-only addition, patch-level).

## Tests

**Unit** (`packages/core/test/middleware/flip.test.ts`) — drives the real `flip` middleware through `computePosition` with a synthetic `detectOverflow`:

- prefers scrollable `bottom` over clipped `top` on the y-axis, and `right` over `left` on the x-axis (`left` over `right` in RTL)
- stays on the scrollable placement when it is already preferred
- still picks the least-overflowing placement among scrollable candidates
- no directional bias when `mainAxis: false`
- with `size()` shrinking the overflowing axis: the bias is dropped and the least-overflowing placement wins
- with `size()` constraining only the cross axis: the bias is kept (regression test for the `fallbackDimension` gate)

**Functional/visual** (`packages/dom/test/functional/flip-best-fit.test.ts` + new `FlipBestFit` visual spec): real-DOM coverage of the y-axis, x-axis, and RTL cases plus the `size()` neutralization case, with 5 new snapshots. No pre-existing snapshots changed.

The full `@floating-ui/core` suite passes and `tsc`/`eslint` are clean. A changeset (patch) and a `fallbackStrategy` docs update are included.


atomiks:
Thanks for the PR, but the original commit didn't (fully) fix the issue, beyond a small part of it. 

The issue wants `bottom`/`right` (LTR) to be preferred even if `top`/`left` fit better geometrically by a substantial amount, based on if it gets clipped or not, so the natural scroll direction allows seeing more. And this should only be applied if `size()` isn't being actively used to constrain the height/width, in which case it should prefer the geometric best fit like before.

Anyways, I got Fable to do exactly that, which it's added as a commit. I'll verify it later

chatman-media:
Thanks for taking this the rest of the way and wiring up the Fable rework — I pulled it down and went through it. The reachability-as-primary-sort-key approach is exactly right, and the `fallbackDimension` gate is a much better answer than what I'd have reached for: keying off `middlewareData.size` alone would have wrongly dropped the bias whenever `size()` is merely present (e.g. when it constrains only the cross axis). Detecting an *actual* shrink on the overflowing axis handles those cleanly.

Verified locally: full `@floating-ui/core` suite green (32/32, flip 8/8); the flip tests fail against pre-PR `core` (5/8 — both axes' bias, RTL, and both `size()` cases) and pass with the change. `tsc`, eslint, prettier clean on all touched files.

One small thing worth a sanity check when you verify: in the neutral re-resolution path the candidate is picked from the *saved* (pre-shrink) `overflows`, then flip's normal flow re-runs on the new placement with current rects — so it self-corrects via the follow-up cycle. If you'd prefer it re-detect before choosing rather than rely on that, happy to adjust. Otherwise LGTM from my side.

chatman-media:
Verified your scrollable-direction rework end to end against the behavior described above — it does exactly what the issue asks for, so nothing further was needed on the branch.

What I checked:

- **Preference is by reachability, not geometric fit.** With `top` overflowing 10px and `bottom` overflowing 150px, `bestFit` now picks `bottom` — the substantially-worse geometric fit wins because its overflow is scrollable while `top`'s is clipped behind the scroll origin. Same on the x-axis (`right` over `left`), and RTL flips the clipped side to `right` so `left` is preferred. Among candidates on the scrollable side, least total overflow still decides.
- **`size()` gate keys off an actual shrink, not mere presence.** When `size()` constrains the overflowing main axis, the bias drops and the geometric best fit wins as before; when `size()` constrains only the cross axis (overflowing axis keeps its dimension via the persisted `fallbackDimension`), the bias is correctly kept. Both paths covered.
- **mainAxis: false** skips the directional bias, since the main-axis overflow direction is unknown there.

Tests are load-bearing: the new `flip.test.ts` fails 5/8 against pre-PR `core` (both axes' bias, RTL, and the `size()` cross-axis-only case) and passes 8/8 with the change; full `@floating-ui/core` suite 32/32. `tsc -b` (core + dom), `eslint`, and `prettier --check` are clean on all touched files, and the functional snapshots cover the y/x/RTL/size cases in a real DOM.

On the open question about the neutral re-resolution path: I'm satisfied leaving it as-is. The `sizeShrankFallback` branch only uses the saved pre-shrink overflows to choose *which* placement to jump to, then `reset` re-runs the whole lifecycle so the chosen placement is re-detected against the current (post-shrink) rects on the follow-up pass — re-detecting before the jump would just duplicate that work. No change recommended there.

PR #3470, opened 2026-06-15
Remove unnecessary passive flag from scroll events

edg2s:
Scroll events are uncancellable and so always passive. See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#using_passive_listeners

### DIFF
--- a/.changeset/calm-walls-listen.md
+++ b/.changeset/calm-walls-listen.md
@@ -1,0 +1,6 @@
+---
+"@floating-ui/dom": patch
+"@floating-ui/react": patch
+---
+
+fix: remove redundant passive options from scroll listeners
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -170,8 +170,7 @@ export function autoUpdate(
       : [];
 
   ancestors.forEach((ancestor) => {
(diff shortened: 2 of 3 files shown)


atomiks:
Interesting how this stayed/was added in the first place 😅 


> Git history says Popper added it in **2016**:
> 
> `37aaf489 use passive event listeners for perf boost`
>
>That commit put `{passive: true}` on both `resize` and `scroll` listeners. At the time, passive listeners were new, Chrome was loudly promoting them for scroll performance, and “scroll-related listener = passive” was an easy mental shortcut. Slightly too broad, but understandable.
>
>Then Floating UI copied/recreated the pattern later:
>
>- `autoUpdate` got it when introduced in **2022**: `45dad87c feat(dom): autoUpdate (#1541)`
>- `useDismiss` got it in the React interactions package in **2022**: `c203160d9 feat: React DOM Interactions package (#1585)`
>- later refactors renamed/moved the code, but the option stayed
>
>Why it lasted: it didn’t break anything, `removeEventListener` matching ignores `passive` anyway, and nobody pays much attention to a “performance-looking” option on scroll listeners unless they’re specifically auditing it. So it was basically inert code.

atomiks (approved):

PR #3471, opened 2026-06-17
fix(react): dismiss outside-press for inline floating element inside …

andyticker:
…a modal portal

When a floating element is rendered inline (without its own FloatingPortal) inside a modal that uses FloatingPortal + FloatingFocusManager(modal), pressing a sibling element outside the floating element was misidentified as a third-party element injected after the floating element rendered.

The modal places its floating element in a portal container under <body> and marks every other body-level subtree with `data-floating-ui-inert`. The inline Popover's floating element lives in the same portal container, so when a sibling is pressed, the target's root ancestor (the portal container) contains no markers, triggering the third-party element skip.

Add a check so that when the target root ancestor contains the floating element, the press is treated as a normal outside press instead of being ignored. This doesn't affect the original third-party element detection, which only applies when the target lives in a different root tree from the floating element.

### DIFF
--- a/.changeset/modern-popover-dismiss.md
+++ b/.changeset/modern-popover-dismiss.md
@@ -1,0 +1,7 @@
+---
+"@floating-ui/react": patch
+---
+
+fix(useDismiss): avoid skipping outside-press dismissal when the target shares the same root ancestor as the floating element
+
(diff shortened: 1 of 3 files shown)


tychenjiajun:
For me, this issue exist since 0.27.*

ryanticker:
@atomiks Can you review this PR?

PR #3472, opened 2026-06-22
fix(inline): use max group height as line-detection threshold in getRectsByLine

JSap0914:
## Problem

`getRectsByLine` groups inline rects by line using the heuristic:

```ts
if (!prevRect || rect.y - prevRect.y > prevRect.height / 2) {
  // new line group
}
```

When a **short rect** (e.g. a superscript, subscript, or small inline element like `<sup>`) follows a taller rect within the same line group, `prevRect` becomes that short rect. Its `height / 2` threshold becomes too small, so the **next rect**—even if it's still within the tall rect's vertical extent—is incorrectly placed in a new group.

**Concrete example** (height=20 tall rect → height=3 short → height=5 rect that should stay on same line):

| Before fix | After fix |
|---|---|
| 3 groups: `[tall+short]`, `[medium]`, `[next-line]` | 2 groups: `[tall+short+medium]`, `[next-line]` |

This produces incorrect floating element positioning when the `inline` middleware operates on a selection that spans text with mixed font sizes (e.g. text containing a superscript or subscript).

## Fix

Track `maxGroupHeight` — the maximum height seen so far in the current group — and use it as the comparison threshold. This prevents a small intermediate rect from inadvertently lowering the threshold for subsequent rects.

```ts
if (!prevRect || rect.y - prevRect.y > maxGroupHeight / 2) {
  groups.push([rect]);
  maxGroupHeight = rect.height;
} else {
  groups[groups.length - 1].push(rect);
  maxGroupHeight = Math.max(maxGroupHeight, rect.height);
}
```

All existing unit tests continue to pass; a new test case is added to cover the regression.

> AI-assisted: this fix was identified and implemented with AI assistance.

### DIFF
--- a/.changeset/fix-inline-getrectsbyl-maxgroupheight.md
+++ b/.changeset/fix-inline-getrectsbyl-maxgroupheight.md
@@ -1,0 +1,14 @@
+---
+"@floating-ui/core": patch
+---
+
+fix(inline): use max group height as line-detection threshold in `getRectsByLine`
+
+When a short rect (e.g. a superscript or small inline element) appears after a
+taller rect within the same line group, `prevRect.height` becomes very small.
+Subsequent rects that are still within the taller rect's vertical extent were
+incorrectly placed in a new group because `prevRect.height / 2` was too small
+a threshold.
+
+Track the maximum rect height seen so far in the current group
+(`maxGroupHeight`) and use that as the threshold instead of `prevRect.height`.
--- a/packages/core/src/middleware/inline.ts
+++ b/packages/core/src/middleware/inline.ts
@@ -28,12 +28,15 @@ export function getRectsByLine(rects: Array<ClientRectObject>) {
   const sortedRects = rects.slice().sort((a, b) => a.y - b.y);
   const groups = [];
   let prevRect: ClientRectObject | null = null;
+  let maxGroupHeight = 0;
   for (let i = 0; i < sortedRects.length; i++) {
     const rect = sortedRects[i];
-    if (!prevRect || rect.y - prevRect.y > prevRect.height / 2) {
+    if (!prevRect || rect.y - prevRect.y > maxGroupHeight / 2) {
       groups.push([rect]);
+      maxGroupHeight = rect.height;
     } else {
       groups[groups.length - 1].push(rect);
+      maxGroupHeight = Math.max(maxGroupHeight, rect.height);
     }
     prevRect = rect;
   }
--- a/packages/core/test/middleware/inline.test.ts
+++ b/packages/core/test/middleware/inline.test.ts
@@ -49,4 +49,21 @@ describe('getRectsByLine', () => {
       rectToClientRect({x: 20, y: 20, width: 10, height: 5}),
     ]);
   });
+
+  test('short rect in group should not truncate line-height threshold', () => {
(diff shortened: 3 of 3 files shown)


atomiks:
# PR review

The PR fixes a real grouping bug but the replacement heuristic regresses adjacent cases: `maxGroupHeight` never decays within a group while the comparison anchor (`prevRect.y`) keeps advancing, so once a tall rect enters a group, later rects on genuinely different lines chain into it. Confirmed by simulation against master; merge-blocking.

## Bugs (1)

### 1. 🔴 Non-decaying `maxGroupHeight` with advancing `prevRect.y` anchor merges subsequent lines into one group

**Location:** `packages/core/src/middleware/inline.ts:36`

```ts
if (!prevRect || rect.y - prevRect.y > maxGroupHeight / 2) {
```

The threshold grows to the tallest rect seen in the group and never shrinks, but the delta is still measured from the *previous* rect, whose `y` advances with every merged rect. The two combined make merging sticky: any rect whose `y` is within `maxGroupHeight / 2` of the previous rect joins the group, raises the anchor, and lets the next line join too. On master, the short-rect threshold (the very behavior this PR removes) broke these chains.

**Failure scenario:** a line containing a tall inline rect (height 20) and a superscript at `y: 7`, followed by a normal 10px line at `y: 14`: master groups it as `[line1], [line2]`, this PR merges line 2 into line 1 (`14 − 7 = 7 < 20/2`). With a slightly taller rect (height 30, line pitch 13) every following line chains in and the whole multi-line range collapses into a single bounding rect — `inline()` then positions off a box spanning all lines, exactly what the middleware exists to avoid. Both cases verified by simulating master vs. PR code side by side.

**Fix:** keep `maxGroupHeight`, but anchor the delta to the group's first (topmost) rect instead of `prevRect`:

```ts
let groupStartY = 0;
let maxGroupHeight = 0;
for (let i = 0; i < sortedRects.length; i++) {
  const rect = sortedRects[i];
  if (!prevRect || rect.y - groupStartY > maxGroupHeight / 2) {
    groups.push([rect]);
    groupStartY = rect.y;
    maxGroupHeight = rect.height;
  } else {
    groups[groups.length - 1].push(rect);
    maxGroupHeight = Math.max(maxGroupHeight, rect.height);
  }
  prevRect = rect;
}
```

Verified this passes all four existing tests, the PR's new test, and both regression scenarios above. (`prevRect` then only serves the first-iteration check and could become a plain boolean.)

## Tests (1)

### 1. 🟠 No test covers lines following a tall rect at normal line pitch

**Location:** `packages/core/test/middleware/inline.test.ts:53`

The new test's second line sits at `y: 30`, far below the tall rect's extent, so it can't catch the chaining failure above. **Failure scenario:** the untested case — tall rect (height 20+) plus short rect on line 1, next line starting within `tallHeight / 2` of the short rect (e.g. `y: 14`) — passes on master and fails on this PR, and nothing in the suite would flag it. **Fix:** add a case like `[{y: 0, h: 30}, {y: 4, h: 8}, {y: 17, h: 11}, {y: 30, h: 11}]` asserting three groups.

## Simplifications (0)

No findings.

## Docs (0)

No findings.

## Verdict

**Request changes** — the new threshold regresses multi-line grouping whenever a group's tallest rect exceeds twice the following line pitch.

---

🤖 Review generated with Claude Code

PR #3475, opened 2026-07-01
fix(dom): observe container-type ancestors in autoUpdate

### LINKED ISSUE #3206: `autoUpdate` not called consistently if an ancestor has `container-type: size/inline-size` with `contain: layout` workaround
**Describe the bug** 

We've run into a scenario involving https://github.com/floating-ui/floating-ui/issues/3067 where `autoUpdate` is not called consistently after the layout is updated.

**To Reproduce**

1. Go to https://codepen.io/jcfranco/pen/MYgZXmG?editors=1000 or https://codesandbox.io/p/sandbox/vigorous-payne-4ysyyw?workspaceId=ws_HRxDc3feKm7cBasXfR7mgK
1. Before loading, ensure the preview area is set to a side-by-side layout with sufficient space to increase the width.
1. After loading, widen the panel so the final width is larger than the initial panel width.
1. Click the "My Button" reference button.
1. Observe that the tooltip is not positioned correctly because `autoUpdate` is not being triggered.

Thanks to @driskull for identifying the repro steps! 🏆

**Expected behavior** 

`autoUpdate` to be called consistently.

Image

**Context:**

- OS: MacOS 15.2
- Browser: Chrome
- Version: 131.0.6778.265

**Additional context** 

N/A

tsushanth:
## What

`autoUpdate` now observes ancestors with `container-type: size/inline-size` via the existing `ResizeObserver`.

## Why

Closes #3206.

When an ancestor has `container-type: size/inline-size`, CSS containment makes that ancestor a sizing context. Resizing it causes the reference element's viewport position to change — but the reference element itself does not resize, so neither the `elementResize` `ResizeObserver` nor the `ancestorResize` `resize` event listeners fire (those cover overflow ancestors). The floating element ends up mis-positioned until the next unrelated trigger.

## How

Walk up the ancestor tree from the reference element and collect elements whose `containerType` computed style is `"size"` or `"inline-size"`. Observe them in the same `ResizeObserver` alongside the reference and floating elements. The existing `resizeObserver.disconnect()` in the cleanup path handles unobserving them — no extra teardown needed.

The walk uses `isLastTraversableNode` / `isHTMLElement` guards already available in `@floating-ui/utils/dom` to stay consistent with the rest of the codebase and handle shadow roots correctly.

### DIFF
--- a/packages/dom/src/autoUpdate.ts
+++ b/packages/dom/src/autoUpdate.ts
@@ -1,14 +1,32 @@
 import {floor, max, min} from '@floating-ui/utils';
 import {
+  getComputedStyle,
   getDocumentElement,
   getOverflowAncestors,
   getWindow,
+  isHTMLElement,
+  isLastTraversableNode,
 } from '@floating-ui/utils/dom';
 
 import type {FloatingElement, ReferenceElement} from './types';
 import {getBoundingClientRect} from './utils/getBoundingClientRect';
 import {unwrapElement} from './utils/unwrapElement';
 import {rectsAreEqual} from './utils/rectsAreEqual';
+
+function getContainerTypeAncestors(element: Element): Element[] {
+  const result: Element[] = [];
+  let node: Node | null = element.parentNode;
+  while (node && !isLastTraversableNode(node)) {
+    if (isHTMLElement(node)) {
+      const {containerType} = getComputedStyle(node);
+      if (containerType === 'size' || containerType === 'inline-size') {
+        result.push(node);
+      }
+    }
+    node = node.parentNode;
+  }
+  return result;
+}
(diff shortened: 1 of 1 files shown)


tsushanth:
Rebased onto current master to resolve the merge conflict and fixed the prettier formatting issue in `autoUpdate.ts` (the `containerAncestors` ternary is now inlined as prettier expects). CI should be green now.

atomiks:
@tsushanth thanks but I already fixed this with a more general solution (https://github.com/floating-ui/floating-ui/pull/3485)

tsushanth:
Thanks @atomiks! Closing in favour of your more general fix in #3485.

PR #3476, opened 2026-07-01
fix(react): defer scroll to selectedIndex by one extra frame on first open

### LINKED ISSUE #3378: Initial scroll to `selectedIndex` doesn't work when use `useListNavigation`
**Describe the bug** Initial scroll to selectedIndex doesn't work when use `useListNavigation`

**To Reproduce**

Precondition:
activeIndex and selectedIndex have initial value

1. Go to 'https://codesandbox.io/p/devbox/strange-butterfly-jmf74z?workspaceId=ws_A5ZEcXgqNumqhheUh5LgRr'
2. Click to the button 
3. Popover doesn't scroll to the selectedIndex
4. Click outside popover for close it 
5. Click to the button again 
6. Popover scrolled to the selectedIndex

**Expected behavior** 
Popover should scroll to the selectedIndex on first open

**Screenshots** 
https://github.com/user-attachments/assets/7227811f-3b7b-4fd4-920e-1623869bf545

**Context:**

- OS: MacOS
- Browser : Chrome
- Version: 138

**Additional context** Add any other context about the problem here.

tsushanth:
## Summary

Fixes #3378 — `useListNavigation` does not scroll to `selectedIndex` on the first open of a floating element, but works correctly on subsequent opens.

**Root cause:** When a floating element opens with a pre-selected index, `forceScrollIntoViewRef` is set to `true` and `scrollIntoView` is scheduled inside a single `requestAnimationFrame`. On the first open, floating-ui's position middleware runs asynchronously, so the floating element hasn't been positioned by the time that rAF fires. `scrollIntoView` runs on an element at its initial (unpositioned) location and has no visible effect. On subsequent opens the element already has its last-known position, so the scroll works.

**Fix:** When `forceScrollIntoView` is set (the initial-open-with-selectedIndex path) and focus is not synchronously scheduled, the `scrollIntoView` call is deferred into a second `requestAnimationFrame`. This gives floating-ui's positioning commit time to complete before the scroll runs.

## Changes

- `packages/react/src/hooks/useListNavigation.ts`: capture `isSyncScheduler` before the scheduler decision; wrap `scrollIntoView` in a nested `requestAnimationFrame` when `forceScrollIntoView && !isSyncScheduler`
- `packages/react/test/unit/useListNavigation.test.tsx`: update the `scrollIntoView on open` test to flush two rAF passes (focus pass → scroll pass)

### DIFF
--- a/packages/react/src/hooks/useListNavigation.ts
+++ b/packages/react/src/hooks/useListNavigation.ts
@@ -356,12 +356,13 @@ export function useListNavigation(
 
     const initialItem = listRef.current[indexRef.current];
     const forceScrollIntoView = forceScrollIntoViewRef.current;
+    const isSyncScheduler = forceSyncFocusRef.current;
 
     if (initialItem) {
       runFocus(initialItem);
     }
 
-    const scheduler = forceSyncFocusRef.current
+    const scheduler = isSyncScheduler
       ? (v: () => void) => v()
       : requestAnimationFrame;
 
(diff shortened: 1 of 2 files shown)


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #3483, opened 2026-07-03
ci: add CVE Lite dependency audit workflow

sonukapoor:
This PR adds a CVE Lite dependency audit workflow to floating-ui. CVE Lite CLI is an OWASP Lab Project that scans pnpm, npm, Yarn, and Bun lockfiles locally for known vulnerabilities without requiring a package install step.

A scan of the current lockfile found 72 findings across 2003 packages: 7 critical, 35 high, 25 medium, and 5 low. Of those, 8 are direct dependencies with copy-ready upgrade commands, including 2 critical-severity devDependencies (`vitest` and `@vitest/browser`). A companion PR with those direct upgrades is coming separately.

The workflow runs on every push and pull request to master, and on a weekly Monday schedule. It produces a SARIF file that is uploaded to GitHub Code Scanning, so findings appear inline on the Security tab. The `fail-on: high` setting means the job will fail when high or critical findings are present, which acts as a merge gate for vulnerable dependency upgrades.

All Actions are pinned to immutable commit SHAs rather than floating version tags. The scan reads the lockfile directly so no `pnpm install` step is needed.

More about CVE Lite CLI and the OWASP Lab Project at https://owasp.org/cve-lite-cli

### DIFF
--- a/.github/workflows/cve-lite.yml
+++ b/.github/workflows/cve-lite.yml
@@ -29,8 +29,9 @@ jobs:
           sarif: 'true'
 
       - name: Upload SARIF to Code Scanning
-        if: always() && hashFiles('*.sarif') != ''
+        if: always()
+        continue-on-error: true
         uses: github/codeql-action/upload-sarif@411bbbe57033eedfc1a82d68c01345aa96c737d7 # v4
         with:
-          sarif_file: cve-lite-*.sarif
+          sarif_file: .
           category: cve-lite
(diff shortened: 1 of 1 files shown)


sonukapoor:
Hey @atomiks - thanks for merging the fix PR!

One quick thing: CVE Lite CLI v1.26.0 shipped a scheduled fix mode that can automate what we just did manually. Adding `fix: true` and `create-pr: true` to this workflow would have it open a single batched PR with the next round of upgrades automatically. Here's an example of what that PR looks like: https://github.com/sonukapoor/cve-lite-test/pull/2

Want me to update this PR to include that? It does require enabling "Allow GitHub Actions to create and approve pull requests" in your repo settings first.

